PR #226, opened 2022-08-23
MCP-1640 Not Found Response

anthonyvelazquez:
# MCP-1640 Not Found Response 

### What was the problem?

Invalid `claimSubmissionID`s that were passed to the PDF Generator still returned a valid response

### How does this fix it?

Display a proper status for an ID that has never been used before. Specifically `NOT_FOUND` rather than `ERROR` or `IN_PROGRESS`

### Jira Tickets

- [MCP-1640](https://amida.atlassian.net/browse/MCP-1640)

## How to test this PR

- Make a request to fetch a PDF for a random `claimSubmissionID` that hasn't been used previously and the response should say `NOT_FOUND`


controller/src/main/java/gov/va/vro/controller/VroController.java
```diff
@@ -95,6 +95,9 @@ public ResponseEntity<Object> fetchPdf(String claimSubmissionId)
 
       log.info("RESPONSE from fetchPdf: {}", pdfResponse.toString());
       if (pdfResponse.hasContent()) {
+        if (pdfResponse.getStatus() == "NOT_FOUND") {
+          return new ResponseEntity<>(pdfResponse, HttpStatus.NOT_FOUND);
+        }
```
au2, line 100: We are now handling NOT_FOUND but not ERROR. Error should be handled too.

api/src/main/java/gov/va/vro/api/resources/VroResource.java
```diff
@@ -68,7 +68,6 @@ ResponseEntity<GeneratePdfResponse> generatePdf(
       description =
           "Downloads the evidence pdf previously generated by the posted patient health data and diagnostic code")
   @GetMapping("/evidence-pdf/{claimSubmissionId}")
-  @ResponseStatus(HttpStatus.OK)
```
au2, line 71: This is only for Swagger documentation. I think we might want to find out how to list multiple responses instead of removing the only one we have.

service-python/pdfgenerator/src/lib/queues.py
```diff
@@ -49,9 +49,10 @@ def on_fetch_callback(channel, method, properties, body):
 						logging.info(f"Fetched PDF")
 						response = {"claimSubmissionId": claim_id, "status": "COMPLETE", "pdfData": str(pdf.decode("ascii"))}
 				else:
-						logging.info(f"PDF still generating")
-						response = {"claimSubmissionId": claim_id, "status": "IN_PROGRESS", "pdfData": ""}
-		except:
+						logging.info(f"Claim ID not found")
+						response = {"claimSubmissionId": claim_id, "status": "NOT_FOUND", "pdfData": ""}
+		except Exception as e:
+				logging.warn(f"Exception: {e}")
```
au2, line 55: I would say this is a error not warn.

au2 (changes requested):

controller/src/main/java/gov/va/vro/controller/VroController.java
dimitri-amida: The catch(Exception) block should handle all other errors. They are redirected to GlobalExceptionHandler

au2: Right but it is not an exception. We return status ERROR from the Python code (exception is handled in the python code).

api/src/main/java/gov/va/vro/api/resources/VroResource.java
au2: You are right. ResponseStatus is not what I thought it was. It is better to remove it I agree.

controller/src/main/java/gov/va/vro/controller/VroController.java
```diff
@@ -101,6 +101,11 @@ public ResponseEntity<Object> fetchPdf(String claimSubmissionId)
       FetchPdfResponse pdfResponse = objectMapper.readValue(response, FetchPdfResponse.class);
       log.info("RESPONSE from fetchPdf returned status: {}", pdfResponse.getStatus());
       if (pdfResponse.hasContent()) {
```
au2, line 103: NOT_FOUND and ERROR still returns 200 because pdfResponse.hasContent is false and your logic never comes into play.

api/src/main/java/gov/va/vro/api/resources/VroResource.java
```diff
@@ -50,7 +52,11 @@ ResponseEntity<HealthDataAssessmentResponse> postHealthAssessment(
       description =
           "Posts health data for a specific patient and diagnostic code to launch evidence pdf generation")
   @PostMapping("/evidence-pdf")
-  @ResponseStatus(HttpStatus.CREATED)
+  @ApiResponses(
+      value = {
+        @ApiResponse(responseCode = "201", description = "Successful Request"),
+        @ApiResponse(responseCode = "500", description = "PDF Generator Server Error")
```
au2, line 58: Can we also add 401 here for unauthorized? That will happen when there is no api key in the header.

api/src/main/java/gov/va/vro/api/resources/VroResource.java
```diff
@@ -68,7 +74,14 @@ ResponseEntity<GeneratePdfResponse> generatePdf(
...
   @GetMapping("/evidence-pdf/{claimSubmissionId}")
-  @ResponseStatus(HttpStatus.OK)
+  @ApiResponses(
+      value = {
+        @ApiResponse(responseCode = "200", description = "Successful Request"),
+        @ApiResponse(
+            responseCode = "404",
+            description = "PDF generation for specified claimSubmissionID not requested"),
+        @ApiResponse(responseCode = "500", description = "PDF Generator Server Error")
+      })
```
au2, line 84: 401 here as well.

au2 (changes requested):

au2:
@anthonyvelazquez  Can we add  @JsonInclude(JsonInclude.Include.NON_EMPTY)  to pdfData in FetchPDFResponse? That way null or empty pdfDatais not added to the response.

au2:
@anthonyvelazquez please also merge develop for the latest.

au2 (approved):

jbregstone (approved):

PR #230, opened 2022-08-25
Version 1 implementation of Asthma and Hypertension processors

jbregstone:
# Version 1 implementation 

### What was the problem?
Asthma and Hypertension processors contained more logic than is necessary for version 1

### How does this fix it?
Implements the exact logic from `vets-api` for filtering medication and blood pressure observations, but in a micro-service architecture

### Jira Tickets

- [MCP-1792](https://amida.atlassian.net/browse/MCP-1792)

## How to test this PR

- Step 1
- Try the full-health-assessment endpoint sample request and ICN 38 with vasrd diagnostic code 6602
- Step 2 
- For 7101 it will sort medication by date, it will not return bp data in the sample test case because it was not taken within the last year
- For 6602 medications are primarily sorted by relevance to the Asthma claim (medications that are used to treat Asthma show up first), secondarily sorted by date.

## Reminders


service-data-access/src/main/java/gov/va/vro/abd_data_access/service/FhirClient.java
```diff
@@ -71,8 +71,7 @@ public String getUrl() {
               }),
           new AbstractMap.SimpleEntry<String, AbdDomain[]>(
               "6602", new AbdDomain[] {
-                      AbdDomain.MEDICATION,
-                      AbdDomain.CONDITION
+                  AbdDomain.MEDICATION
```
wlin2019, line 74: I thought we just added this CONDITION. You don't want it now?

wlin2019 (approved):
LGTM

service-data-access/src/main/java/gov/va/vro/abd_data_access/service/FhirClient.java
jbregstone: For version 1 I realized it is not needed

service-python/assessclaimdc7101/src/lib/continuous_medication.py
```diff
@@ -25,9 +13,9 @@ def continuous_medication_required(request_body):
 
   veterans_medication = request_body["evidence"]["medications"]
   for medication in veterans_medication:
-    medication_display = medication["description"]
-    for keyword in [x.lower() for x in hypertension_medications]:
```
au2, line 29: It is a bit odd that we are lower casing a constant list. Can't the list be just lower cases?

service-python/assessclaimdc7101/src/lib/continuous_medication.py
```diff
@@ -25,9 +13,9 @@ def continuous_medication_required(request_body):
...
-      if (keyword in medication_display.lower()):
-        relevant_medications.append(medication)
+    if medication["status"].lower() == "active":
+      relevant_medications.append(medication)
```
au2, line 18: I would break here. Otherwise you might end up adding a medication twice as in 200 ACTUAT Albuterol 0.09 MG/ACTUAT Metered Dose Inhaler

au2 (changes requested):

au2:
@jbregstone "200 ACTUAT Albuterol 0.09 MG/ACTUAT Metered Dose Inhaler" gets entered twice to the list. I suggested a change so that a medication can only enter the list once. You are leaving multiple empty lines in places. I am pretty sure that that is not recommended in python either. Do we have a linter we are using here to check?

service-python/assessclaimdc7101/src/lib/continuous_medication.py
jbregstone: This line was removed, but yes I agree

service-python/assessclaimdc7101/src/lib/continuous_medication.py
jbregstone: I am iterating through the medications, If it meets the condition it is added to the list. I do not see how it will be added twice, the keyword list and conditional logic were removed for now

au2: Because two keywords can be in the medication description like in the example I gave that comes from the sandbox ("200 ACTUAT Albuterol 0.09 MG/ACTUAT Metered Dose Inhaler" ). IN this case both Albuterol and Inhal gets hit. SInce you do not break after you find the first one the medication is added twice.

au2: Why did you remove the logic? I thought we need the logic to match va.gov. You just need to fix it.

jbregstone: The medication list is from the hypertension prototype. My understanding is that in va.gov there is no medication list for hypertension.

jbregstone:
> @jbregstone "200 ACTUAT Albuterol 0.09 MG/ACTUAT Metered Dose Inhaler" gets entered twice to the list. I suggested a change so that a medication can only enter the list once. Also you are leaving multiple empty lines in places. I am pretty sure that that is not recommended in python either. Do we have a linter we are using here to check?

I will apply the Black python code linter. A common convention is to have two empty lines before a function definition. The other places with extra lines should be taken care of by the code linter.

au2 (approved):

PR #234, opened 2022-08-26
Only publish images for branches `main` and `develop`

yoomlam:
# Feature Name 
Split `main.yml` GH Action into `build.yml` and `publish.yml` since we only need to publish container images for branches `main` and `develop`.

### What was the problem?

Unnecessarily publishing container images for all branches. This pollutes the GitHub Container Registry with images that we don't use. The images are only needed for LHDI deploys.

### How does this fix it?

Only publishes container images for git pushes to branches `main` and `develop`.

## Reminders

I'll update https://github.com/department-of-veterans-affairs/abd-vro/wiki/Github-Actions

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,79 +1,16 @@
-name: Main
+name: Publish images
 
 on:
   push:
     branches:
-      - "*"
+      - main
+      - develop
 
 env:
   GITHUB_ACCESS_TOKEN: ${{ secrets.ACCESS_TOKEN }}
   COMMIT_SHA: ${{ github.sha }}
 
 jobs:
-  cis-benchmark-tests:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/publish.yml
```diff
@@ -1,79 +1,16 @@
...
-
-      - name: "DC7101: PIP Install"
-        run: pip install -r service-python/assessclaimdc7101/src/requirements.txt
-
-      - name: "PDF Generator: PIP Install"
-        run: pip install -r service-python/pdfgenerator/src/requirements.txt
-
-      - name: "Pytest: Code Coverage"
-        run: cd service-python && python -m pytest --cov-report term-missing --cov
-
```
yoomlam, line 76: These jobs are moved to the `build.yml` action.

advcoder (approved):

PR #262, opened 2022-09-07
SD Elements: Docker Non Root User

anthonyvelazquez:
# SD Elements: Docker Non Root User 

### What was the problem?

`root` is used by default for all the Dockerfile commands

### How does this fix it?

Create a new non root user to comply with SD Elements request


au2 (changes requested):
1) You need to fix all the Dockerfiles (including the java ones). Should not be much different. If you want to do those in another pr that is fine too.
2) Please fix the checks. build-docker-images is failing. Something to do with python version.

jbregstone (approved):
I will approve but I did not try running as a rootless user

service-data-access/src/docker/Dockerfile
```diff
@@ -6,4 +6,5 @@ WORKDIR /project
 COPY ${ENTRYPOINT_FILE} entrypoint.sh
 RUN chmod +x entrypoint.sh
 COPY ${JAR_FILE} service-data-access.jar
+RUN adduser -m docker
 CMD ["sh", "entrypoint.sh"]
```
au2, line 10: We are only adding the user? I thought we needed to switch to the user as well?

app/src/docker/Dockerfile
```diff
@@ -8,4 +8,5 @@ COPY ${JAR_FILE} starter-java.jar
 EXPOSE 8081
 HEALTHCHECK CMD curl --fail http://localhost:8081/health || exit 1
 LABEL org.opencontainers.image.source=https://github.com/department-of-veterans-affairs/abd-vro
+RUN adduser -m docker
```
au2, line 11: Only adding user? Shouldn't we switch as well?

au2 (changes requested):

au2 (approved):
@anthonyvelazquez  I think we should return an error when there is no content but I do not think it will ever go to that line so will let it go. For future I would identify if that line is ever reached always return error when there is no content.

au2 (changes requested):
I meant to approve the other pr which I did. This still needs the inline comments addressed.

au2 (approved):
End points are working. Moving this.

PR #265, opened 2022-09-08
PDF Template Updates

anthonyvelazquez:
# PDF Template Updates 

### What was the problem?

- Some headings needed to be a larger font size
- Some timestamps were missing time zones
- Some text needed to be on its own line


service-python/pdfgenerator/src/requirements.txt
```diff
@@ -12,6 +12,7 @@ pdfkit==1.0.0
 pytest==7.1.2
 pytest-cov==3.0.0
 python-dateutil==2.8.2
+pytz
```
jbregstone, line 15: You should specify a version

jbregstone (changes requested):
The sample data on the swagger page returns an error. I think the medication data may not be parse-able.

anthonyvelazquez:
> The sample data on the swagger page returns an error. I think the medication data may not be parse-able.
> 
> edit: the medication authoredOn, in the sample request data, is not matching the expected format.

Swagger Example is fixed

service-python/pdfgenerator/src/lib/templates/hypertension/rating_schedule.html
```diff
@@ -1,4 +1,4 @@
-<h3>Hypertension Rating Schedule</h3>
+<h2>Hypertension Rating Schedule</h2>
```
jbregstone, line 1: This change does not seem to have an effect on the text size in the PDF. I tried putting in the <h1> or <h6> and it looks the same.

anthonyvelazquez: Hmm not sure what the issue might be. I reran it after making the same updates and got the changes
image
image
image

jbregstone: Interesting, I tried rebuilding a couple times but I will try again

jbregstone: Screen Shot 2022-09-12 at 1 46 54 PM
For me it still looks like this, not sure why

jbregstone (approved):

PR #266, opened 2022-09-08
Remove unnecessary build-docker-app job from the build GH Action

yoomlam:
# Feature Name 

### What was the problem?

The `build-docker-app` job runs as part of the `build.yml` GitHub Action for every PR.
This job doesn't seem relevant since VRO is runs entirely from images.
We don't use the generated "fat jar" file at all, so we don't need to run this job.

I also question the need for the `./Dockerfile`.

### How does this fix it?

Remove unnecessary `build-docker-app` job


yoomlam:
To make sure this PR doesn't break anything (by removing `./Dockerfile` and replacing with the vro-app image in `secrel/config.yml`), I manually started SecRel pipeline: https://github.com/department-of-veterans-affairs/abd-vro-internal/actions/runs/3018651396

PR #267, opened 2022-09-08
Remove ./Dockerfile

yoomlam:
# Feature Name 

## Description

Remove `./Dockerfile` and replace with vro-app image in `secrel/config.yml`

### What was the problem?

Questioning the value/purpose of `./Dockerfile`


yoomlam:
To make sure this PR doesn't break anything (by removing `./Dockerfile` and replacing with the vro-app image in `secrel/config.yml`), I manually started SecRel pipeline: https://github.com/department-of-veterans-affairs/abd-vro-internal/actions/runs/3018651396
which now runs faster since the "fat jar" isn't being analyzed. Hmmm, this means that dependency jars are not being scanned -- but do they need to be if the scanner is able to interpret Gradle dependencies from the `build.gradle` files? I'll start a conversation with the SecRel team.

advcoder (approved):

PR #263, opened 2022-09-07
Update to flyway version 9.2.3 to use updated H2 and Postgres drivers

yoomlam:
# Feature Name 

### What was the problem?

Aqua alerts about vulnerabilities with older H2 and Postgres drivers

### How does this fix it?

Flyway version 9.2.3 uses h2-2.1.214.jar and postgresql-42.4.1.jar (located in the `/flyway/drivers` folder of the image)

### How to test

* Manually mirror this repo to the internal repo - https://github.com/department-of-veterans-affairs/abd-vro/wiki/Github-Actions#mirror
* In the internal repo, manually run the SecRel workflow action on this branch - https://github.com/department-of-veterans-affairs/abd-vro-internal/actions/workflows/secrel.yml
* Check the results in Aqua - https://aqua.lighthouse.va.gov/#/images/CiCdScans

PR #276, opened 2022-09-09
Update base image of app's Dockerfile

yoomlam:
# Feature Name 

### What was the problem?

Aqua alerts about vulnerabilities with openssl and glibc for the vro-app image.

Similar to PR #263 but for a different image.

### How does this fix it?

Use an updated jre image

## How to test this PR

https://github.com/department-of-veterans-affairs/abd-vro-internal/wiki/Secure-Release-process#to-test-prs-in-the-secrel-pipeline

### DIFF
--- a/app/src/docker/Dockerfile
+++ b/app/src/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM openjdk:17
+FROM eclipse-temurin:17-jre
 ARG JAR_FILE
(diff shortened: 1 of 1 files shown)


yoomlam:
Confirmed that this PR reduces the `critical` and `high` alerts to 0. There are 12 `med` and 32 `low` alerts remaining.

PR #305, opened 2022-09-16
Remove init-container image from SecRel

yoomlam:
# Feature Name 

### What was the problem?

init-container image has `critical` SecRel alerts.

The image is [not needed](https://dsva.slack.com/archives/C01Q7979Z7D/p1662589199987369).

### How does this fix it?

Removes init-container image from SecRel processing


advcoder (approved):

yoomlam:
Remove `container-init` folder and references to the Gradle subproject.

PR #325, opened 2022-09-20
MCP-1884 PDF Generator Status Codes

anthonyvelazquez:
# MCP-1884 PDF Generator Status Codes 

## Description

Updating Swagger to match error codes that get generated


ianwolson (approved):

jbregstone:
#326 Will change the HTTP code to 401 instead of 403 which may make these changes unnecessary

ianwolson (approved):

anthonyvelazquez:
> PR #326 Will change the HTTP code to 401 instead of 403 which may interfere with these changes

Ah missed that. Will close

PR #331, opened 2022-09-21
MCP-1870 Python Lint + Code Format

anthonyvelazquez:
# MCP-1870 Python Lint + Code Format 

## Description

Added linting, code formatting and a check to Github

### Jira Tickets

- [MCP-1870](https://amida.atlassian.net/browse/MCP-1870)


jbregstone (commented):
If the format and lint command is run in in Github actions, will the code changes be reflected in the PR? If not, we could add the command to a pre-push config.

au2 (changes requested):
Can we add this to ./gradle build as well ?

au2:
@anthonyvelazquez consider https://github.com/linkedin/pygradle. did not dig in much but fro what i understand it does have some of the task that are very useful. flake8 for example. it does appear that it will require some setup.

ianwolson (approved):

au2:
I think I will move these changes as add a new ticket for gradle integration

au2 (approved):

PR #332, opened 2022-09-21
MCP-1873 PDF Generator Error Reason Field

anthonyvelazquez:
#MCP-1873 PDF Generator Error Reason Field 

## Description

- PDF Generator was still returning a 200 even though we got a `status: ERROR`
- PDF Generator now adds a `reason` field in addition to the `status: ERROR` to give a better description of what went wrong

### Jira Tickets

- [MCP-1873](https://amida.atlassian.net/browse/MCP-1873)

## How to test this PR

- Turn off Redis container and generate a PDF. Status should be `500` and you should see a `reason`


au2 (changes requested):
A test calling generate pdf is failing.

au2 (changes requested):
@anthonyvelazquez I put a fix for the test in branch MCP-1873-Fix, please take a look. Essentially the test is erroneous, results in 500 but expects 200. Since we are returning 200 before for errors, it was passing. Now it is failing. It actually turned out to be a good before and after  test for this pr. In MCP-1873-Fix I put proper input and also return proper output from the mock route.

au2 (approved):

anthonyvelazquez:
image

ianwolson (approved):

PR #333, opened 2022-09-21
Config/add qa deploy

advcoder:
fix for dev auto-deploy to use the right image

add qa auto deployment


.github/workflows/deploy-qa.yml
```diff
@@ -0,0 +1,93 @@
+name: Publish images
```
yoomlam, line 1: Rename this to differentiate from the "Publish images" action.  @advcoder 
Also it's failing: https://github.com/department-of-veterans-affairs/abd-vro/actions/workflows/deploy-qa.yml

PR #374, opened 2022-09-26
Add Aqua Trivy GH action

yoomlam:
# Feature Name

### DIFF
--- a/.github/workflows/trivy.yml
+++ b/.github/workflows/trivy.yml
@@ -1,0 +1,40 @@
+name: Trivy scan
+
+on:
+  pull_request:
+    branches: [main, develop]
+    # Limit to certain PR event types since this action doesn't need to run for every commit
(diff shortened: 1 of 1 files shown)


advcoder (approved):

.github/workflows/trivy.yml
```diff
@@ -0,0 +1,70 @@
...
+          scan-type: 'image'
+          image-ref: 'va/abd_vro-app:latest'
+          format: 'sarif'
+          output: 'trivy-image-results.sarif'
+          severity: 'CRITICAL,HIGH'
+      - name: Upload image results to GitHub Security tab
+        uses: github/codeql-action/upload-sarif@v2
+        with:
+          sarif_file: 'trivy-image-results.sarif'
+          category: trivy-image-app
```
yoomlam, line 70: Note this is only scanning 1 of the many images we have. If this is useful, we can add the other images.

PR #395, opened 2022-09-27
Service Ruby Delete

anthonyvelazquez:
# Service Ruby Delete 

## Description

Delete `service-ruby` related files


ianwolson (approved):

jbregstone:
Line 45 in `cicd-deploy.sh` references service ruby, 
`--set-string images.serviceRuby.tag="${IMAGE_TAG}"\`

I think it should be removed

au2 (changes requested):
There is rby related thing in cicd-deploy. NOt sure how it is being used but I would remove it.

au2 (approved):
We need to remove ruby from whatever running codeql but I do not think that is this pr's job. Although we need to wait  to make sure that error goes away to merge.

PR #409, opened 2022-09-28
Revert "MCP-1911: Removed unused dependencies"

dimitri-amida:
Reverts department-of-veterans-affairs/abd-vro#400

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -1,7 +1,4 @@
-import java.text.SimpleDateFormat
-
 plugins {
-    id("io.gatling.gradle") version "${gatling_plugin_version}" apply false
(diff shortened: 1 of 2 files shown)


dimitri-amida:
@au2 @wlin2019 you both mentioned an unspecified warning after these changes, so I am reverting the PR. Admittedly, I don't know the impact of removing these libraries.

au2 (changes requested):
The changes you have does not fix the problem. You need to add gatling related tasks back to app build.gradle  to the list in tasks.named('forkedSpringBootRun').configure {

I would keep non gatling related changes here. I do not think they cause any problems.

dimitri-amida:
closing this

PR #411, opened 2022-09-28
Python in Gradle

anthonyvelazquez:
# Python in Gradle 

## Description

Added 3 commands to Gradle Build

- `./gradlew pyisort`
- `./gradlew pyflake8`
- `./gradlew pytest`

They each depend off each other so if you want to run the full code formatting + testing then just run `./gradlew pytest`


jbregstone:
The `./gradlew pytest` command is working for me but I was unable to verify that the python tests run on the `./gradlew build check docker` build command. Is it supposed to run on every build?

build.gradle
```diff
@@ -123,3 +157,22 @@ tasks.register('resetloop') {
...
+    command = "."
+    dependsOn pyisort
+}
+
+tasks.register("pytest", PythonTask) {
+    workDir = "service-python"
+    module = 'pytest'
+    command = "--cov-report term-missing --cov=. --cov-fail-under=80 --cov-config=../.coveragerc"
+    dependsOn pyflake8
+}
```
yoomlam, line 178: Nice! Since all these are running in the `service-python` directory, I'm wondering if these would be in a `build.gradle` file in the `service-python` directory. Then it can be run with `./gradlew service-python:pytest`, or can be triggered to run when `./gradlew test` is run by adding a gradle test dependency `test.dependsOn(pytest)` (or using something like https://github.com/datastax/junitpytest).

build.gradle
```diff
@@ -39,6 +41,38 @@ plugins {
...
+    pip 'pytz:2022.1'
+    pip 'redis:4.3.4'
+    pip 'requests:2.28.1'
+    pip 'Werkzeug:2.1.1'
+    pip 'zipp:3.8.0'
+
+    pip 'cerberus:1.3.4'
+    pip 'python-util:1.2.1'
+    pip 'python-dotenv:0.20.0'
+    pip 'six:1.16.0'
```
yoomlam, line 74: Can these dependencies be put in a `requirements.txt` file so vulnerability scanners and Dependabot can help us keep these updated? 
Since there are common libraries across python subprojects, perhaps we can put commonalities in a `service-python/requirements.txt` file, and run `pip install -r requirements.txt ../requirements.txt` (https://stackoverflow.com/questions/11704287/split-requirements-files-in-pip, https://dev.to/fronkan/requirements-txt-in-your-requirements-txt-4loc) when needed?

build.gradle
anthonyvelazquez: Added

anthonyvelazquez:
Pytest now runs on gradle build as part of the process and pip requirements have been fixed so they aren't hard coded

yoomlam (approved):
Code changes look good! I did not test.

jbregstone:
When I tried running the `./gradlew buid check docker` command it failed because of files from a previous build. I think the linting needs to be restricted from looking at a previous build because I can't fix that by building again(since that was the command I was trying to run)

au2 (approved):
I made an intentional python error and gradle build caught it so all good.

PR #416, opened 2022-09-29
Dependabot/test pr

yoomlam:
Partially testing #415
Have to wait for Dependabot to create a PR to really see if workaround actually works


yoomlam:
Seems to work

PR #444, opened 2022-10-03
MCP-1827: V2 asthma

jbregstone:
# V2 Update for Asthma

### What was the problem?

More information is available and should be added to the claim to be used as evidence. 

### How does this fix it?

Proposal to rename `asthmaRelevant` to `conditionRelated` so it can be reused for all conditions.

### Classification of Prescribed Medications
`suggestedCategory` is included in the medication object to categorize medications by class. This information is required for RVSR's to enter into the calculator. We can provide an accurate mapping of Asthma related medications to their medication classes. 

This PR also adds the Condition FHIR resource made available by the Lighthouse API. Active conditions are filtered for relevance to asthma. 

note: SNOMEDCT and ICD-10 codes are included in the same sets because they will not intersect. 

### Jira Tickets

- [MCP-1827](https://amida.atlassian.net/browse/MCP-1827)

## How to test this PR

- Step 1 
`./gradlew -PenablePrototype=true build check docker`:
-`export COMPOSE_PROFILES=prototype`
- `./gradlew :app:dockerComposeDown :app:dcPrune :app:dockerComposeUp`

- Step 2 
- go to Swagger page
- enter "6602v2" for `diagnosticCode`

## Reminders


service-data-access/src/test/java/gov/va/vro/abddataaccess/service/FhirClientTest.java
```diff
@@ -89,24 +89,24 @@ private void mockGetBundle(Bundle bundle, AbdDomain domain) throws AbdException
...
-      AbdEvidence evidence = client.getMedicalEvidence(testClaim);
-      assertNotNull(evidence);
-      assertTrue(evidence.getMedications().size() > 0);
-      assertEquals(evidence.getMedications().size(), medicationBundle.getEntry().size());
-    } catch (Exception e) {
-      log.error("testGetMedicalEvidence error: {}", e.getMessage(), e);
-      fail("testGetMedicalEvidence");
-    }
-  }
+  //
```
jbregstone, line 92: @wlin2019 How could we fix this test, now that I am adding back in Condition to the 6602 claim data

wlin2019: I wonder whether this branch has been merged with the develop branch correctly or not. The test directories look different. I'll try to fix it.

jbregstone: Thanks, I will update the branch with develop.

wlin2019: I updated V2-asthma branch to make service-data-access/test directory the same as the one in develop branch. All service-data-access tests are passed.

wlin2019 (commented):
LGTM for Java code.

anthonyvelazquez:
Is this a proof of concept or will this replace V1? Just so I know if I should make a PR for the PDF Generator to look for the new field

jbregstone:
> Good question, this is more of a proof of concept and won't replace V1 until it gets full approval, you should not have to make a PR for the PDF generator for the time being.

anthonyvelazquez (approved):

au2 (changes requested):
You need to fix the Synk alert.

app/src/docker/docker-compose.yml
```diff
@@ -153,6 +154,24 @@ services:
       - internal
     restart: on-failure:5
 
+  assess-claim-dc6602v2:
```
au2, line 157: Let's not add this here for a while. There is a docker-compose extension files. I think that would be better at this stage.

au2 (changes requested):
inline

app/src/docker/docker-compose.yml
jbregstone: Like this? https://docs.docker.com/compose/extends/

app/src/docker/docker-compose.v2.yml
```diff
@@ -0,0 +1,26 @@
...
+      RABBITMQ_PLACEHOLDERS_USERNAME: ${RABBITMQ_PLACEHOLDERS_USERNAME}
+      RABBITMQ_PLACEHOLDERS_USERPASSWORD: ${RABBITMQ_PLACEHOLDERS_USERPASSWORD}
+    security_opt:
+      - no-new-privileges:true
+    depends_on:
+      rabbitmq-service:
+        condition: service_healthy
+    networks:
+      - internal
+    restart: on-failure:5
```
au2, line 25: I removed this from others. Docker compose is only for development so we do not really need it.

app/src/docker/docker-compose.v2.yml
```diff
@@ -0,0 +1,26 @@
...
+  assess-claim-dc6602v2:
+    image: va/abd_vro-assessclaimdc6602v2:latest
+    build: ../../../service-python/assessclaimdc6602v2/src/.
+    environment:
+      RABBITMQ_PORT: 5672
+      RABBITMQ_RETRY_LIMIT: 3
+      RABBITMQ_TIMEOUT: 10800
+      RABBITMQ_PLACEHOLDERS_HOST: rabbitmq-service
+      RABBITMQ_PLACEHOLDERS_USERNAME: ${RABBITMQ_PLACEHOLDERS_USERNAME}
+      RABBITMQ_PLACEHOLDERS_USERPASSWORD: ${RABBITMQ_PLACEHOLDERS_USERPASSWORD}
```
au2, line 17: See if we can use the blocks in docker-compose.yml here.

service-python/assessclaimdc6602v2/build.gradle
```diff
@@ -0,0 +1,23 @@
...
+    copySpec.into(".").from("../"){
+        include("*.py")
+    }
+    // Use container image cache
+    noCache false
+}
+
+dockerRun {
+    image "${dockerRegistry}/${rootProject.name}-service-assessclaimdc6602"
+    env 'RABBITMQ_PLACEHOLDERS_USERNAME': getEnvOrDefault('DB_SERVICE_PASSWORD', 'also-not-the-service-password'),
```
au2, line 20: I do not think DB_SERVICE_PASSWORD is the right env variable. We should match whatever docker-compose is using including the defaults.

service-python/assessclaimdc6602v2/src/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+RUN adduser --disabled-password docker
+USER docker
+
+WORKDIR /home/docker
+
+COPY --chown=docker:docker . /home/docker/
+
+RUN pip install --no-cache-dir --user -r requirements.txt
+
```
au2, line 19: Make sure this matches the existing. Me thinks we should be able to share this file somehow but not this pr's problems.

service-python/assessclaimdc6602v2/src/lib/condition.py
```diff
@@ -0,0 +1,23 @@
...
+    :rtype: dict
+    """
+    relevant_conditions = []
+
+    veterans_conditions = request_body["evidence"]["conditions"]
+    for condition in veterans_conditions:
+        if condition["status"].lower() in ["active", "relapse", "recurrence"]:
+            condition_code = condition["code"]
+            if condition_code in condition_codesets.asthma_conditions:
+                relevant_conditions.append(condition)
```
au2, line 19: Some of these lists are quite big. I would consider using Set instead of list.

service-python/assessclaimdc6602v2/src/lib/main.py
```diff
@@ -0,0 +1,37 @@
...
+        response_body.update(
+            {
+                "evidence": {
+                    "medications": active_medications["medications"],
+                    "conditions": active_conditions,
+                },
+                "evidenceSummary": {
+                    "relevantMedCount": active_medications["relevantMedCount"],
+                    "totalMedCount": active_medications["totalMedCount"],
+                },
```
au2, line 31: add similar count  for conditions here.

service-python/assessclaimdc6602v2/src/lib/medication.py
```diff
@@ -0,0 +1,68 @@
...
+        for medication in medication_dict[category_id]:
+            if medication in medication_display.lower():
+                medication_category.append(category_id)
+                break
+    return medication_category
+
+
+def medication_required(request_body):
+    """
+    Determine if there is the veteran requires continuous medication for hypertension
```
au2, line 28: Why does this talk about hypertension?

service-python/assessclaimdc6602v2/src/lib/medication.py
```diff
@@ -0,0 +1,68 @@
...
+    veterans_medication = request_body["evidence"]["medications"]
+    for medication in veterans_medication:
+        if medication["status"].lower() in ["active", "on-hold", "completed", "stopped", "unknown"]:
+            medication["conditionRelated"] = "false"
+            medication_display = medication["description"]
+            category = categorize_med(medication_display)
+            medication["suggestedCategory"] = category
+            if category:
+                medication["conditionRelated"] = "true"
+                relevant_medications.append(medication)
```
au2, line 47: I thught only active medications mattered?

service-python/assessclaimdc6602v2/src/main_consumer.py
```diff
@@ -0,0 +1,80 @@
+import logging
+import os
```
au2, line 2: I thought Python services shared this.

svc-lighthouse-api/src/main/java/gov/va/vro/abddataaccess/model/AbdMedication.java
```diff
@@ -16,6 +16,8 @@ public class AbdMedication implements Comparable<AbdMedication> {
   private String description;
   private int refills;
   private Boolean asthmaRelevant;
```
au2, line 18: SO what are we doing with asthmaRelevant?

au2 (changes requested):
Sone initial comments.

au2:
@jbregstone what patient(s) are we testing this with?

service-python/assessclaimdc6602v2/src/lib/medication.py
jbregstone: Based on the finalized asthma report "completed" and "unknown" medications should also be included. Refer to page 20-24. 

> 

> Q: Why does the algorithm also consider the medications labeled as “completed” or “unknown”?
> A: A “completed” medication suggests that all actions that are implied by the prescription have occurred. The algorithm considers “completed” medications because adjudicators can use this information (as well as the completed medication date) to evaluate the recency and relevancy of Veterans’ asthma medications.
> An “unknown” medication is one where the source system does not know which of the status values currently applies for this observation. The algorithm includes “unknown” medications in case the medication is “active.” Adjudicators might need to review additional treatment records to make the “active” determination.
>

service-python/assessclaimdc6602v2/src/main_consumer.py
jbregstone: This is included as a stop-gap measure to ensure the image can be built. It will be removed when the image is added to the gradle build.

svc-lighthouse-api/src/main/java/gov/va/vro/abddataaccess/model/AbdMedication.java
jbregstone: It should be removed when it is safe to do so.

service-python/assessclaimdc6602v2/src/main_consumer.py
jbregstone: I recognize this is not ideal

jbregstone:
> 

ICN 38 has conditions of Asthma

1012666073V986297 and 1012661611V839382 have medications but no conditions

app/src/docker/docker-compose.v2.yml
jbregstone: it is not possible

jbregstone: If they are used multiple times, if sinusitis and other conditions are added to this file, I will define the blocks again here so they can be used.

au2 (changes requested):
@jbregstone  I looked at this some more; the structure is not ended up what I expected. I think the way we are doing it is not really maintainable. These prototype will be conditional for a while so maybe we need a better solution. I was thinking we can put these into the main docker-compose conditionally using profiles (https://docs.docker.com/compose/profiles/). And we can add this to the build conditionally using https://www.baeldung.com/gradle-conditional-dependencies. Please take a look and we can discuss tomorrow.

service-python/assessclaimdc6602v2/build.gradle
jbregstone: This file has been changed

settings.gradle
```diff
@@ -89,5 +88,6 @@ include ':service-python'
 include ':service-python:assessclaimdc7101'
 include ':service-python:assessclaimdc7101v2'
 include ':service-python:assessclaimdc6602'
+include ':service-python:assessclaimdc6602v2'
```
jbregstone, line 91: it seems that if the subproject is included here, it gets included in the build no matter if the prototype property is set

jbregstone: I tried using this a as a conditional in settings.gradle but it did not work, `startParameter.projectProperties.hasProperty("enablePrototype")`, The only way I could figure out how to set a property in this file was through gradle.properties

au2 (approved):

PR #459, opened 2022-10-03
MCP-1924 Python Consumer Updates

anthonyvelazquez:
# MCP-1924 Python Consumer Updates 

## Description

Added more details on how the consumer works as well as a check to see if the connection was established before trying to close it which was causing errors in the QA env


service-python/main_consumer.py
```diff
@@ -26,7 +26,8 @@ def __init__(self, config):
         self.setup_queues()
```
au2, line 26: Still calling this if self.connection is None. If I see self.connection to be None in garbage collection it can be None here too.

service-python/main_consumer.py
```diff
@@ -44,17 +45,25 @@ def setup_queues(self):
...
+# since the retry loop will recreate it
 if __name__ == "__main__":
 
     start_timer = None
     current_timer = None
 
     while(True):
+        consumer = None
         try:
             consumer = RabbitMQConsumer(CONSUMER_CONFIG)
```
au2, line 62: This I do not understand. Suppose _create_connection failed in all tries (so self.connection is None)? What then? the consumer will not be None and we will happily call start_consuming?

au2 (changes requested):
Inline comments

anthonyvelazquez:
Script now checks if the connection is not `None` in the class creation as well as in the runtime loop, it checks if **channel** is not `None` as well

jbregstone (approved):

service-python/main_consumer.py
```diff
@@ -37,24 +39,33 @@ def _create_connection(self):
             except Exception:
                 logging.warning(f"RabbitMQ Connection Failed. Retrying in 30s ({i + 1}/{self.config['retry_limit']})")
```
au2, line 40: One more thing. Can we add information about the exception here? Looks like logging.warning has a exc_info parameter where you can specify the exception instance.

service-python/main_consumer.py
```diff
@@ -37,24 +39,33 @@ def _create_connection(self):
...
     while(True):
+        consumer = None
         try:
             consumer = RabbitMQConsumer(CONSUMER_CONFIG)
-
-            consumer.channel.start_consuming()
+            if consumer.channel:
+                consumer.channel.start_consuming()
         except Exception:
+            del consumer
```
au2, line 68: Please put a logging warning here for the exception.  Looks like logging.warning has a exc_info parameter where you can specify the exception instance.

au2 (changes requested):
@anthonyvelazquez last two changes.

au2 (approved):
Some of the exceptions still not what I expect but we have messages now and it appears that the service recovers from rabbit errors so approving. We can do another round later.

PR #509, opened 2022-10-11
Saving evidence count summary as json object

ianwolson:
# MCP-1887 Assessment Result Population

### What was the problem?

We need to be able to save the Evidence Summary information from full-health-data-assessment into the Assessment Result table in the DB.

### How does this fix it?

Adjusted SaveToDbService to save evidence summary as JSON in assessment result table

### Jira Tickets

- [MCP-1887](https://amida.atlassian.net/browse/MCP-1887)

## How to test this PR

- Execute full-health-data-assessment in swagger
- Verify the evidence summary is in the assessment_result table in the Postgres DB

## Reminders


service/db/src/main/java/gov/va/vro/service/db/SaveToDbServiceImpl.java
```diff
@@ -30,6 +39,59 @@ public Claim insertClaim(Claim claim) {
...
+      for (Object value : summary.values()) {
+        count += (Integer) value;
+      }
+    }
+    AssessmentResultEntity assessmentResultEntity = new AssessmentResultEntity();
+    assessmentResultEntity.setEvidenceCount(count);
+    assessmentResultEntity.setEvidenceCountSummary(summary);
+    for (ContentionEntity contention : claimEntity.getContentions()) {
+      if ((contention.getDiagnosticCode().equals(diagnosticCode))
+          && (contention.getClaim().getId() == id)) {
```
dimitri-amida, line 74: The id check is not necessary because we already obtained the contentions from the claim

service/db/src/main/java/gov/va/vro/service/db/SaveToDbServiceImpl.java
```diff
@@ -30,6 +39,59 @@ public Claim insertClaim(Claim claim) {
...
+        for (AssessmentResultEntity ar : contention.getAssessmentResults())
```
dimitri-amida, line 75: These nested structure make my head hurt. Can we simplify a bit by introducing methods? Maybe something like this:

```
var contention = findContention(claim)
var assessmentResult = findAssessmentResult(contention)
...
 
```

service/provider/src/main/java/gov/va/vro/service/provider/camel/PrimaryRoutes.java
```diff
@@ -64,9 +68,29 @@ private void configureRouteClaimSubmitForFull() {
...
         // https://examples.javacodegeeks.com/apache-camel-headers-vs-properties-example/
+        .setProperty("id", simple("${body.recordId}"))
+        .setProperty("diagnosticCode", simple("${body.diagnosticCode}"))
+        .setProperty("veteranIcn", simple("${body.veteranIcn}"))
+        .setProperty("claimSubmissionId", simple("${body.claimSubmissionId}"))
+        .setProperty("idType", simple("${body.idType}"))
         .setProperty("diagnosticCode", simple("${body.diagnosticCode}"))
         .routingSlip(method(SlipClaimSubmitRouter.class, "routeClaimSubmit"))
-        .routingSlip(method(SlipClaimSubmitRouter.class, "routeClaimSubmitFull"));
+        .routingSlip(method(SlipClaimSubmitRouter.class, "routeClaimSubmitFull"))
```
dimitri-amida, line 78: Should we call only claimSubmitFull instead of both claimSubmit and claimSubmitFull? @au2 ?

service/provider/src/main/java/gov/va/vro/service/provider/camel/PrimaryRoutes.java
```diff
@@ -64,9 +68,29 @@ private void configureRouteClaimSubmitForFull() {
...
+        .convertBodyTo(String.class)
+        .process(
+            new Processor() {
+              @Override
+              public void process(Exchange exchange) throws Exception {
+                String response = (String) exchange.getIn().getBody();
```
dimitri-amida, line 84: This confuses me immensely. If all the information we need is available before we call "Claim Submit", why do we need to persist this data *after* claim Submit? We can just persist it as part of the initial "insertClaim". What am I missing?

app/src/test/java/gov/va/vro/ClaimRepositoryTest.java
```diff
@@ -20,11 +23,20 @@ void insertQuery() {
...
-    contention2.addAssessmentResult(1);
+    Map<String, String> evidence2 = new HashMap<>();
+    AssessmentResultEntity ar2 = new AssessmentResultEntity();
+    evidence2.put("medicationsCount", "10");
+    ar2.setEvidenceCountSummary(evidence2);
+    ar2.setEvidenceCount(1);
+    contention2.addAssessmentResult(ar2);
     contention2.addAssessmentResult(2);
     var claim = TestDataSupplier.createClaim("123", "type", veteran);
     claim.addContention(contention1);
```
dimitri-amida, line 42: We need code to retrieve the assessment result and verify that it contains the correct JSON

build.gradle
```diff
@@ -92,6 +93,7 @@ subprojects {
             annotationBom platform("gov.va.starter:starter-bom:${starter_boot_version}")
             // https://github.com/department-of-veterans-affairs/lighthouse-di-starter-boot/blob/main/checkstyle-bom/build.gradle
             checkstyleRules platform("gov.va.starter:checkstyle-bom:${starter_boot_version}")
+            implementation 'com.vladmihalcea:hibernate-types-55:2.20.0'
```
dimitri-amida, line 96: same comment as before: Let's move these to the common directory

dimitri-amida (changes requested):
Good job on persisting the JSON object! I have several comments/questions in the comments

service/provider/src/main/java/gov/va/vro/service/provider/camel/PrimaryRoutes.java
```diff
@@ -64,9 +68,29 @@ private void configureRouteClaimSubmitForFull() {
         .process(FunctionProcessor.fromFunction(saveToDbService::insertClaim))
         // Use Properties not Headers
         // https://examples.javacodegeeks.com/apache-camel-headers-vs-properties-example/
+        .setProperty("id", simple("${body.recordId}"))
+        .setProperty("diagnosticCode", simple("${body.diagnosticCode}"))
```
au2, line 72: Diagnostic code is already set a couple of lines below.

service/provider/src/main/java/gov/va/vro/service/provider/camel/PrimaryRoutes.java
```diff
@@ -64,9 +68,29 @@ private void configureRouteClaimSubmitForFull() {
...
+        .setProperty("veteranIcn", simple("${body.veteranIcn}"))
+        .setProperty("claimSubmissionId", simple("${body.claimSubmissionId}"))
+        .setProperty("idType", simple("${body.idType}"))
```
au2, line 75: Who is exactly using claimSubmissionId and idType that we are storing them here?

service/provider/src/main/java/gov/va/vro/service/provider/camel/PrimaryRoutes.java
```diff
@@ -64,9 +68,29 @@ private void configureRouteClaimSubmitForFull() {
         .process(FunctionProcessor.fromFunction(saveToDbService::insertClaim))
         // Use Properties not Headers
         // https://examples.javacodegeeks.com/apache-camel-headers-vs-properties-example/
+        .setProperty("id", simple("${body.recordId}"))
```
au2, line 71: This surprises me. I thought these audit saves were not supposed to change the messages. Looks like they do. I do not really know the implication. I would have preferred these did not changed the message and this property is set within the processor.  The downstream microservice certainly does not expect these additional fields that are artifacts of the database record.

service/provider/src/main/java/gov/va/vro/service/provider/camel/PrimaryRoutes.java
```diff
@@ -64,9 +68,29 @@ private void configureRouteClaimSubmitForFull() {
...
+        .process(
+            new Processor() {
+              @Override
+              public void process(Exchange exchange) throws Exception {
+                String response = (String) exchange.getIn().getBody();
+                String icn = (String) exchange.getProperty("veteranIcn");
+                UUID id = (UUID) exchange.getProperty("id");
+                String claimSubmissionId = (String) exchange.getProperty("claimSubmissionId");
+                String idType = (String) exchange.getProperty("idType");
+                String diagnosticCode = (String) exchange.getProperty("diagnosticCode");
```
au2, line 89: claimSubsmissionId and idType is not really used in insertAssessmentResult. veteranIcn and diagnostic code is available from response. I do not think we need these parameters. id is not descriptive enough. maybe claimRecordId?  I really wish claimSubmissionId would be the thing to search for the claim record but that will probably need some indexing changes in the database.

service/spi/src/main/java/gov/va/vro/service/spi/model/AssessmentResult.java
```diff
@@ -0,0 +1,37 @@
...
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+
+import java.util.Map;
+
+@Getter
+@Setter
+@NoArgsConstructor
+@AllArgsConstructor
+public class AssessmentResult {
```
au2, line 16: This class confuses me. I gather we return this after assessment. But the controller expects the object that healthassessment microservice uses. If this is indeed changing the message it should not really to my understanding. Save to dbs were supposed to be audit logs that do not change anything in the messages. I actually do not understand if this really becomes the return value and how it is used. I need to dig in some more.

service/db/src/main/java/gov/va/vro/service/db/SaveToDbServiceImpl.java
```diff
@@ -30,6 +40,74 @@ public Claim insertClaim(Claim claim) {
...
+      String claimSubmissionId,
+      String idType,
+      String diagnosticCode)
+      throws NoSuchFieldException {
+    ClaimEntity claimEntity =
+        claimRepository
+            .findById(id)
+            .orElseThrow(
+                () ->
+                    new NoSuchFieldException(
```
au2, line 57: Who exactly will handle this exception in the flow? Since it is inside Camel I am assuming it cannot be the controller. Did we test this?

service/db/src/main/java/gov/va/vro/service/db/SaveToDbServiceImpl.java
```diff
@@ -30,6 +40,74 @@ public Claim insertClaim(Claim claim) {
...
+                        "Could not match claim ID {" + id.toString() + "} in DB."));
+    Map<String, Object> evidence = new HashMap<>();
+    try {
+      evidence = objMapper.readValue(assessmentResult, Map.class);
+      if (evidence.isEmpty()) {
+        throw new NoSuchFieldException("Could not map assesmentResult, evidence summary is null.");
+      }
+    } catch (Exception e) {
+      log.error("Could not map assesmentResult, evidence summary is null.");
```
au2, line 66: So we will catch what we threw a line back here and then happily continue the rest of the code?

service/db/src/main/java/gov/va/vro/service/db/SaveToDbServiceImpl.java
```diff
@@ -30,6 +40,74 @@ public Claim insertClaim(Claim claim) {
...
+    }
+    Map<String, Object> defaultEvidence = new HashMap<>();
+    defaultEvidence.put("medicationsCount", 0);
```
au2, line 69: why do we have a default here. there might be possibly conditions that do not use medications.

service/db/src/main/java/gov/va/vro/service/db/SaveToDbServiceImpl.java
```diff
@@ -30,6 +40,74 @@ public Claim insertClaim(Claim claim) {
...
+    }
+    AssessmentResultEntity assessmentResultEntity = new AssessmentResultEntity();
+    assessmentResultEntity.setEvidenceCount(count);
+    assessmentResultEntity.setEvidenceCountSummary(summary);
+    ContentionEntity contention = findContention(claimEntity, diagnosticCode);
+    for (AssessmentResultEntity ar : contention.getAssessmentResults()) {
+      if (ar.getContention().getId() == contention.getId()) {
+        log.info("Assessment result already populated, continuing.");
+        break;
+      }
```
au2, line 85: I do not understand this for loop. break just gets out of the for loop and continues. Is this just for the message? Otherwise it does not change the flow at all.

au2 (changes requested):
Please see the inline comments. I think I am especially surprised that we seem to be changing message in these "supposed to be audit" processors so I will need to dig a bit deeper to understand. Also I have not looked at the test yet. But we do seem to be storing what I thought we should store.

persistence/model/build.gradle
```diff
@@ -11,6 +11,10 @@ dependencies {
     // include jdbc tracing here since we are assuming for this version we are using JDBC SQL
     // implementation "io.opentracing.contrib:opentracing-spring-cloud-jdbc-starter"
     implementation 'org.springframework.boot:spring-boot-starter-data-jpa'
+// https://mvnrepository.com/artifact/com.vladmihalcea/hibernate-types-55
+    implementation 'com.vladmihalcea:hibernate-types-55:2.20.0'
```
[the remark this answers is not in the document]
au2: no i do not remember. i actually read that is discouraged unless it is a convention plugin. that is in another comment.

persistence/model/src/main/java/gov/va/vro/persistence/model/ContentionEntity.java
```diff
@@ -27,14 +27,14 @@ public class ContentionEntity extends BaseEntity {
...
-      fetch = FetchType.LAZY,
+      fetch = FetchType.EAGER,
       cascade = CascadeType.ALL,
       orphanRemoval = true)
   private List<AssessmentResultEntity> assessmentResults = new ArrayList<>();
 
   @OneToMany(
       mappedBy = "contention",
-      fetch = FetchType.LAZY,
+      // fetch = FetchType.EAGER,
```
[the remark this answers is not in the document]
ianwolson: Not sure if we need, but need 4 lines to pass spotless apply and keep consistent styling

persistence/model/src/main/java/gov/va/vro/persistence/model/ContentionEntity.java
```diff
@@ -65,6 +65,15 @@ public AssessmentResultEntity addAssessmentResult(int evidenceCount) {
     return assessmentResult;
   }
 
+  public AssessmentResultEntity addAssessmentResult(AssessmentResultEntity ar) {
```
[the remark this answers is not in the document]
ianwolson: Up to you, I can remove it.

service/db/src/main/java/gov/va/vro/service/db/SaveToDbServiceImpl.java
ianwolson: I had set it to null but @dimitri-amida suggested this was a bad idea.  Should I make it ("default", 0)?

ianwolson:
Will make a new branch with a more descriptive name, my apologies.

PR #526, opened 2022-10-13
Adjust project dependencies

yoomlam:
# Feature Name 

### What was the problem?

Various clean up:
- The `app` container shouldn't be dependent on microservice containers
- Don't rebuild db-init docker image every time `dockerComposeUp` is run
- Remove old reference to `gov.va.starter.example.persistence.model`

Fix:
- Spring's RabbitHealthIndicator needs a connection to RabbitMQ -- [Slack](https://dsva.slack.com/archives/C01Q7979Z7D/p1665675931459679)

### How does this fix it?

Fixes the items above.


app/build.gradle
```diff
@@ -71,6 +71,10 @@ tasks.dockerPrepare.configure {
     dependsOn(tasks.bootJar)
 }
 
+docker {
+    noCache false
+}
+
```
yoomlam, line 77: Anyone see a problem with using this setting (allowing docker to use image caches and not forcing it to recreate the same image layer) for all projects with `Dockerfile`s? This change only applies to the `app` image.

app/src/main/resources/conf-camel-rabbitmq.yml
```diff
@@ -6,3 +6,11 @@ message-queue:
...
 mas-processing-initial-delay: "${MAS_PROCESSING_INITIAL_DELAY:2000}"
 mas-processing-subsequent-delay: "${MAS_PROCESSING_SUBSEQUENT_DELAY:1000}"
+
+# Spring's RabbitHealthIndicator needs a connection to RabbitMQ:
+spring:
+  rabbitmq:
+    host: "${RABBITMQ_PLACEHOLDERS_HOST:localhost}"
+    port: 5672
+    user: "${RABBITMQ_PLACEHOLDERS_USERNAME:guest}"
+    password: "${RABBITMQ_PLACEHOLDERS_USERPASSWORD:guest}"
```
yoomlam, line 16: Fixes `Rabbit health check failed` error in the `app` container:
```
o.s.a.r.c.CachingConnectionFactory       : Attempting to connect to: [localhost:5672]
o.s.b.a.amqp.RabbitHealthIndicator       : Rabbit health check failed
org.springframework.amqp.AmqpConnectException: java.net.ConnectException: Connection refused
```

dimitri-amida (approved):
lgtm. Fixes the healthcheck problem

app/build.gradle
dimitri-amida: will it recreate the image if the Dockerfile changes?

yoomlam: It should. You can double-check by running `./gradlew :app:docker` and it should say `CACHED` if nothing has changed:
```
#6 [3/7] WORKDIR /project
#6 sha256:60ed93b080ff8a664f6b2614f423ad674ce0b98e8e61f8b33c9fa9b3d1127f68
#6 CACHED

#8 [4/7] COPY entrypoint.sh entrypoint.sh
#8 sha256:0b74e2307f05f49125a38ef5ff143ebd80d5a38471494b144ac2332f3d70931e
#8 CACHED
```

au2 (approved):
lgtm

app/build.gradle
dimitri-amida: then I don't see a problem with that

PR #551, opened 2022-10-18
Move domain folder to new shared folder

yoomlam:
## What was the problem?

A new `shared` folder was created in #550.
The `domain` folder is a shared library and should be under the `shared` folder.

## How does this fix it?

Move domain folder to new shared folder

## How to test this PR
Manual testing


dimitri-amida:
For my part, I don't see the need for these nested sub-projects. They overcomplicate the structure without adding much value. We can very easily just separate internal modules by packages and use archunit to enforce dependency rules, if necessary.

yoomlam:
> We can very easily just separate internal modules by packages and use archunit to enforce dependency rules, if necessary.

Could you describe more or demonstrate this?

To be clear, this PR intends to organize all the shared libraries under a single folder so that we can more easily see the top-level projects. I anticipate having more shared libraries. If you have a better approach, I’m all ears.

au2:
@yoomlam please resolve conflicts

au2 (approved):
app still works. i do not have an issue with grouping things.

yoomlam:
Added documentation: https://github.com/department-of-veterans-affairs/abd-vro/wiki/Code-structure#gradle-projects

PR #561, opened 2022-10-21
Mcp 1887 populate assessment result

ianwolson:
## What was the problem?

- [MCP-1887](https://amida.atlassian.net/browse/MCP-1887)

## How does this fix it?[^1]
After claim is processed in full-health-assessment the assessment result is populated in the db

## How to test this PR
- try out full-health-assessment
- check database to see that assessment result is populated and evidence count summary holds proper JSON


service/spi/src/main/java/gov/va/vro/service/spi/model/AssessmentResult.java
```diff
@@ -0,0 +1,33 @@
...
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+
+import java.util.Map;
+
+@Getter
+@Setter
+@NoArgsConstructor
+@AllArgsConstructor
+public class AssessmentResult {
```
au2, line 16: This is essentially copy of FullHealthDataAssessmentResponse except it is missing ABDEvidence. That makes sense since that is the result of Health Assessment. Can't we just use that one here? Maybe by including api project?

service/db/src/main/java/gov/va/vro/service/db/SaveToDbServiceImpl.java
```diff
@@ -30,6 +40,57 @@ public Claim insertClaim(Claim claim) {
...
+    AssessmentResultEntity assessmentResultEntity = new AssessmentResultEntity();
+    assessmentResultEntity.setEvidenceCountSummary(summary);
+    ContentionEntity contention = findContention(claimEntity, diagnosticCode);
+    contention.addAssessmentResult(assessmentResultEntity);
+    claimRepository.save(claimEntity);
+    AssessmentResult resp = new AssessmentResult();
+    resp.setEvidenceSummary(assessmentResultEntity.getEvidenceCountSummary());
+    resp.setVeteranIcn(veteranIcn);
+    resp.setDiagnosticCode(diagnosticCode);
+    return resp;
```
au2, line 77: This return value is not being used. Which is the right thing to do I believe. But then why are we returning this? Also we can probably have this object in the signature of this function instead of string..

service/db/src/main/java/gov/va/vro/service/db/SaveToDbServiceImpl.java
```diff
@@ -30,6 +40,57 @@ public Claim insertClaim(Claim claim) {
     return claim;
   }
 
+  @Override
+  public AssessmentResult insertAssessmentResult(
+      UUID id, String assessmentResult, String veteranIcn, String diagnosticCode)
```
au2, line 45: assesmentResult has both veteranIcn and diagnosticCode fields so I am not sure why we need them in the signature. See the comment a few lines below.

service/db/src/main/java/gov/va/vro/service/db/SaveToDbServiceImpl.java
```diff
@@ -30,6 +40,57 @@ public Claim insertClaim(Claim claim) {
...
+    ClaimEntity claimEntity =
+        claimRepository
+            .findById(id)
+            .orElseThrow(
+                () ->
+                    new NoSuchFieldException(
+                        "Could not match claim ID {" + id.toString() + "} in DB."));
+    Map<String, Object> evidence = new HashMap<>();
+    try {
+      evidence = objMapper.readValue(assessmentResult, Map.class);
```
au2, line 56: Can't you just map to AssesmentResult here? Since you have already defined it. It can simplify the following code quite a bit.  If you do that ICN and diagnostic code params become unnecessary.

service/db/src/main/java/gov/va/vro/service/db/SaveToDbServiceImpl.java
```diff
@@ -30,6 +40,57 @@ public Claim insertClaim(Claim claim) {
...
+      if (evidence.isEmpty()) {
+        throw new NoSuchFieldException("Could not map assesmentResult, evidence summary is null.");
+      }
+    } catch (Exception e) {
+      log.error("Could not map assesmentResult, evidence summary is null.");
+    }
```
au2, line 62: I still do not understand why we are continuing here if there has been an exception. evidence is not valid in that case so continuation does not make sense to me.

service/db/src/main/java/gov/va/vro/service/db/SaveToDbServiceImpl.java
```diff
@@ -30,6 +40,57 @@ public Claim insertClaim(Claim claim) {
...
+            .findById(id)
+            .orElseThrow(
+                () ->
+                    new NoSuchFieldException(
+                        "Could not match claim ID {" + id.toString() + "} in DB."));
+    Map<String, Object> evidence = new HashMap<>();
+    try {
+      evidence = objMapper.readValue(assessmentResult, Map.class);
+      if (evidence.isEmpty()) {
+        throw new NoSuchFieldException("Could not map assesmentResult, evidence summary is null.");
```
au2, line 58: Isn't this exception going to be caught two lines below? You will also lose this "Could not map assesmentResult, evidence summary is null." message.

service/provider/src/main/java/gov/va/vro/service/provider/camel/PrimaryRoutes.java
```diff
@@ -64,9 +68,23 @@ private void configureRouteClaimSubmitForFull() {
...
+        .setProperty("veteranIcn", simple("${body.veteranIcn}"))
         .routingSlip(method(SlipClaimSubmitRouter.class, "routeClaimSubmit"))
-        .routingSlip(method(SlipClaimSubmitRouter.class, "routeClaimSubmitFull"));
+        .routingSlip(method(SlipClaimSubmitRouter.class, "routeClaimSubmitFull"))
+        .convertBodyTo(String.class)
+        .process(
+            new Processor() {
+              @Override
+              public void process(Exchange exchange) throws Exception {
+                String response = (String) exchange.getIn().getBody();
```
au2, line 81: I think you can use <T> T getIn(Class<T> type) with AssessmentResult here since you already defined that class (but also see the comment on that class0.  If you do that there is no need for veteranIcn and diagnosticCode below since AssessmentResult will have those.

service/provider/src/main/java/gov/va/vro/service/provider/camel/PrimaryRoutes.java
```diff
@@ -64,9 +68,23 @@ private void configureRouteClaimSubmitForFull() {
         .process(FunctionProcessor.fromFunction(saveToDbService::insertClaim))
         // Use Properties not Headers
         // https://examples.javacodegeeks.com/apache-camel-headers-vs-properties-example/
+        .setProperty("id", simple("${body.recordId}"))
```
au2, line 71: I would name this to be a bit less generic.  claim-db-id or something like that.

au2 (changes requested):
@ianwolson Please see the inline comments. I could not get this working because evidence_count column cannot be null and we removed it from the entity. Even when I modified flyway script  to drop the column I cannot get it to work from swagger thought. I did not investigate why.

service/db/src/main/java/gov/va/vro/service/db/SaveToDbServiceImpl.java
ianwolson: The response from the controller in the end does have the ICN and the diagnostic code, but this assessment result is just the evidence from Primary Routes.  I double checked to see what is in it when it is sent to SaveToDb and this is what we get:

This is the assessment result{"evidence": {"medications": [{"status": "Active", "notes": ["Naproxen sodium 220 MG Oral Tablet"],   ETC 

I will look into seeing if I can get the entire response.

service/db/src/main/java/gov/va/vro/service/db/SaveToDbServiceImpl.java
```diff
@@ -30,6 +39,51 @@ public Claim insertClaim(Claim claim) {
...
+
+  private ContentionEntity findContention(ClaimEntity claim, String diagnosticCode)
+      throws NoSuchFieldException {
+    ContentionEntity resp = new ContentionEntity();
+    for (ContentionEntity contention : claim.getContentions()) {
+      if (contention.getDiagnosticCode().equals(diagnosticCode)) {
+        resp = contention;
+      } else {
+        log.error("Could not find match contention with diagnostic code.");
+        throw new NoSuchFieldException("Could not find match contention with diagnostic code.");
```
dimitri-amida, line 81: NoSuchField is not the right exception for this situation. Also, if we throw an exception here, what are the consequences for the route?

service/db/src/main/java/gov/va/vro/service/db/SaveToDbServiceImpl.java
```diff
@@ -30,6 +39,51 @@ public Claim insertClaim(Claim claim) {
...
+  @Override
+  public void insertAssessmentResult(UUID claimId, String evidence, String diagnosticCode)
+      throws NoSuchFieldException {
+
+    ClaimEntity claimEntity =
+        claimRepository
+            .findById(claimId)
+            .orElseThrow(
+                () ->
+                    new NoSuchFieldException(
```
dimitri-amida, line 51: NoSuchField is not the right exception.

dimitri-amida (changes requested):

service/db/src/main/java/gov/va/vro/service/db/SaveToDbServiceImpl.java
```diff
@@ -30,6 +39,36 @@ public Claim insertClaim(Claim claim) {
     return claim;
   }
 
+  @Override
+  public void insertAssessmentResult(
+      UUID claimId, AbdEvidenceWithSummary evidenceResponse, String diagnosticCode) {
+    ClaimEntity errorClaim = new ClaimEntity();
+    errorClaim.setIncomingStatus("ERROR");
+    ClaimEntity claimEntity = claimRepository.findById(claimId).orElse(errorClaim);
```
au2, line 47: Why all this effort with errorClaim. Can't we just use orElse(null) and check null?

service/provider/src/main/java/gov/va/vro/service/provider/services/AssessmentResultProcessor.java
```diff
@@ -0,0 +1,38 @@
...
+    String responseBody = exchange.getIn().getBody(String.class);
+    if (claimId == null || diagnosticCode == null || responseBody == null) {
+      if (claimId == null) log.warn("Claim Id was empty, exiting");
+      if (diagnosticCode == null) log.warn("Diagnostic Code was empty, exiting.");
+      if (responseBody == null)
+        log.warn("Evidence response from claim-submit-full was empty, exiting");
+      return;
+    }
+    ObjectMapper mapper = new ObjectMapper();
+    AbdEvidenceWithSummary evidence = mapper.readValue(responseBody, AbdEvidenceWithSummary.class);
```
au2, line 35: No. Use AbdEvidenceWithSummary.class with getBody above. Is that not working?

service/db/src/main/java/gov/va/vro/service/db/SaveToDbServiceImpl.java
```diff
@@ -30,6 +39,36 @@ public Claim insertClaim(Claim claim) {
...
+    if (claimEntity.getIncomingStatus().equals("ERROR")) {
+      log.warn("Could not match Claim ID in insertAssessmentResult, exiting.");
+      return;
+    }
+    Map summary = evidenceResponse.getEvidenceSummary();
+    AssessmentResultEntity assessmentResultEntity = new AssessmentResultEntity();
```
au2, line 53: Do you not want to check summary for null or empty?

au2 (changes requested):
SOme inline comments.

service/provider/src/main/java/gov/va/vro/service/provider/services/AssessmentResultProcessor.java
ianwolson: @au2 That is not working, this is the only way I could get it to work.

service/provider/src/main/java/gov/va/vro/service/provider/camel/PrimaryRoutes.java
```diff
@@ -71,8 +73,11 @@ private void configureRouteClaimSubmitForFull() {
         // Use Properties not Headers
         // https://examples.javacodegeeks.com/apache-camel-headers-vs-properties-example/
         .setProperty("diagnosticCode", simple("${body.diagnosticCode}"))
-        .routingSlip(method(claimSubmitRouter, "routeClaimSubmit"))
-        .routingSlip(method(claimSubmitRouter, "routeClaimSubmitFull"));
+        .setProperty("claim-id", simple("${body.recordId}"))
+        .routingSlip(method(SlipClaimSubmitRouter.class, "routeClaimSubmit"))
+        .routingSlip(method(SlipClaimSubmitRouter.class, "routeClaimSubmitFull"))
+        .convertBodyTo(String.class)
```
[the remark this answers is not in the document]
ianwolson: @au2 This is the only way I can get it to map from JSON to an object.  Not sure how to convert it from the byte array.

service/db/src/main/java/gov/va/vro/service/db/SaveToDbServiceImpl.java
ianwolson: @au2 We check for null in AssessmentResultProcessor before sending it over here, but I can do a second check.

service/provider/src/main/java/gov/va/vro/service/provider/services/AssessmentResultProcessor.java
au2: I looked at this. I thought this was already in but please add gov.va.vro.model.* to dto classes as shown below. This is in conf-camel.yml. 
....
    dto-classes:
      - gov.va.vro.service.spi.model.*
      - gov.va.vro.model.mas.*
      - gov.va.vro.model.*
....
Then AbdEvidenceWithSummary evidence = exchange.getIn().getBody(AbdEvidenceWithSummary.class); works and you save two lines and possible exception.

service/db/src/main/java/gov/va/vro/service/db/SaveToDbServiceImpl.java
```diff
@@ -30,6 +39,38 @@ public Claim insertClaim(Claim claim) {
...
+      return;
+    }
+    if (evidenceResponse.getEvidenceSummary() == null) {
+      log.warn("Evidence Summary is empty in insertAssessmentResult, exiting.");
+      return;
+    }
+    Map summary = evidenceResponse.getEvidenceSummary();
+    AssessmentResultEntity assessmentResultEntity = new AssessmentResultEntity();
+    assessmentResultEntity.setEvidenceCountSummary(summary);
+    ContentionEntity contention = findContention(claimEntity, diagnosticCode);
```
au2, line 57: We are not error handling here. If findContention logs an error we should not continue here. I would just return null from findContention and test that and return here.

service/db/src/main/java/gov/va/vro/service/db/SaveToDbServiceImpl.java
```diff
@@ -30,6 +39,38 @@ public Claim insertClaim(Claim claim) {
...
+  }
+
+  private ContentionEntity findContention(ClaimEntity claim, String diagnosticCode) {
+    ContentionEntity resp = new ContentionEntity();
+    for (ContentionEntity contention : claim.getContentions()) {
+      if (contention.getDiagnosticCode().equals(diagnosticCode)) {
+        resp = contention;
+      } else {
+        log.error("Could not find match contention with diagnostic code.");
+      }
```
au2, line 69: I would return null here after error so that caller do not continue.

au2 (changes requested):
@ianwolson One last thing inline.

au2 (approved):

dimitri-amida (approved):

PR #564, opened 2022-10-21
MCP-1985: audit event infrastructure

dimitri-amida:
## What was the problem?

We need a mechanism to capture the lifecycles of claims, automated claims, etc and persiste them as described in (https://amida.atlassian.net/browse/MCP-1983)

- [MCP-1985](https://amida.atlassian.net/browse/MCP-1985)

## How does this fix it?[^1]

This PR addresses the infrastructure part of the ticket. The mechanism is introduced to log events as part of camel routes. The persistence, queues, and notification mechanisms will be addressed in the remaining tasks.

## How to test this PR
- This is not yet testable


rajesh-Krishnasamy (approved):

service/provider/src/main/java/gov/va/vro/service/event/AuditEventProcessor.java
```diff
@@ -0,0 +1,60 @@
...
+
+@Component
+@RequiredArgsConstructor
+public class AuditEventProcessor {
+
+  private final EventService eventService;
+
+  public Object logEvent(Object o, String routeId, String message) {
+    if (!(o instanceof Auditable auditableObject)) {
+      // object cannot be audited
```
au2, line 18: Not sure why we are expecting not Auditable object in the arguments and using an object. I would say that would be a software error. I would at least add a log.error or log.info here based on that what the thinking is.

service/provider/src/main/java/gov/va/vro/service/event/AuditEventProcessor.java
```diff
@@ -0,0 +1,60 @@
...
+      return o;
+    }
+    AuditEvent event =
+        AuditEvent.builder()
+            .eventId(auditableObject.getEventId())
+            .payloadType(auditableObject.getClass())
```
au2, line 24: So we send the class and send the only interface value. Why not just send the object (auditableObject) itself? We will repeat the same lines in logException below.

au2 (approved):
@dimitri-amida I added a couple of inline comments but not major to hold so approving.

service/provider/src/main/java/gov/va/vro/service/event/AuditEventProcessor.java
dimitri-amida: There is a general exception handler that catches any exception, but we don't want to audit all exceptions, only the ones for objects that we are auditing. Therefore the check and conversion is necessary. It is not really an error if the object is not auditable, because we just skip it. I can add a log.info to make this clear in the next ticket.

service/provider/src/main/java/gov/va/vro/service/event/AuditEventProcessor.java
dimitri-amida: We do not log the object itself because it may contain PII. In the next ticket, I will provide a mechanism for selectively extract information from the object.

PR #582, opened 2022-10-25
MCP-1986 audit event persistence model

dimitri-amida:
## What was the problem?

We need a way to track the history of a manual or automated claim. 

- [MCP-1986](https://amida.atlassian.net/browse/MCP-1986)

## How does this fix it?[^1]

This PR introduces the following enhancements to the audit event framework:

- audit events are persisted in the database
- The framework allows extracting information from the Auditable object and storing this information in a "details" column
- Mas endpoints are recording audit events

## How to test this PR
- Call automatedClaim endpoint from Swagger
- Examine the database: (select * from audit_event) to see the history of this claim
- Do the same for examOrderingStatus


service/provider/src/main/java/gov/va/vro/service/provider/camel/PrimaryRoutes.java
```diff
@@ -35,10 +37,14 @@ public class PrimaryRoutes extends RouteBuilder {
 
   private final SaveToDbService saveToDbService;
 
+  private final SlipClaimSubmitRouter claimSubmitRouter;
```
dimitri-amida, line 40: Use the Spring-managed bean rather than a new instance every time

api/src/main/java/gov/va/vro/api/responses/MasResponse.java
```diff
@@ -9,7 +9,7 @@
 @Schema(
     name = "MASClaimDetailsResponse",
     description = "Indicate that the request information has been collected")
-public class MasClaimResponse {
+public class MasResponse {
```
dimitri-amida, line 12: Renaming because it applies to both Mas APIs

rajesh-Krishnasamy (approved):

wlin2019 (approved):

db-init/src/main/resources/database/migrations/V3.0__Create_Event_Table.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE IF NOT EXISTS audit_event (
+    id uuid DEFAULT uuid_generate_v4 () NOT NULL,
+    event_id VARCHAR NOT NULL,
+    route_id VARCHAR,
+    payload_type VARCHAR NOT NULL,
+    message VARCHAR,
+    throwable TEXT,
+    details TEXT,
```
yoomlam, line 8: How can we ensure that no PII/PHI is stored in the DB?

dimitri-amida: eventId, routeId, palyload_type do not contain PII. The message and details is under the control of the developer. This leaves the possibility that some PII makes into the stacktrace somehow, but in this case it would also go to the log files, which is a bigger problem.

yoomlam: I'm referring to the `details` field. What data goes into that and how can we ensure that a developer does not put PII in it? Perhaps just adding a reminder in code comments?

Issue #562, opened 2022-10-21
Implement a VRO Console for monitoring and adding internal VRO messages


yoomlam:
As a follow-on to prototyping in #502, implement a VRO Console for monitoring (for diagnosing) and adding internal VRO messages (for recovery).

yoomlam:
PR being reviewed.

PR #531, opened 2022-10-13
Add VRO Console container

yoomlam:
## What was the problem?

With VRO's deployment to prod, we'll need ways to investigate and recover from errors in the production environment.

When an error occurs, to investigate and diagnose, we'll need to examine the processing state of the claim by looking at Camel routes and the DB contents.

Associated ticket: #502 

## How does this fix it?[^1]

Create a "VRO Console" container that connects to the DB to allow investigation of DB contents.
To keep this relatively PR small, the following will be in separate PRs:
- Connecting with Camel within the Console.
- Adding this console container to deploy scripts and GH Actions
- Tests for the new code
- Add Groovy test code coverage check
- ~Remove `LABEL` command from all `Dockerfile`s~ (SecRel uses the Dockerfile directly, so let's keep the LABEL)

## How to test this PR
- `export COMPOSE_PROFILES=debug` -- the console container only starts if the docker-compose `debug` profile is enabled.
- start VRO. Note the new `vro-console-1` container; check out the logs: `docker logs vro-console-1`.
- Attach to the container: `docker attach vro-console-1`
- Review https://groovy-lang.org/groovysh.html
- On the `groovy:000>` prompt, try the following:
```groovy
?  # display help
:show variables
claimsT.findAll().collect{ it.claimSubmissionId }
c = getAnyClaim()
// Different ways to do the same thing:
printJson c
printJson getAnyClaim()
pj c
// exit   # This will stop the container
// Instead, press Ctrl-p Ctrl-q to detach from the container without stopping it
```

If you need to restart the console container:
```
cd app/src/docker
docker-compose up -d console
docker attach vro-console-1
```

[^1]: [Pull-Requests guidelines](https://github.com/department-of-veterans-affairs/abd-vro/wiki/Pull-Requests). If PR is significant, update [Current Software State](https://github.com/department-of-veterans-affairs/abd-vro/wiki/Current-Software-State) wiki page.
[^secrel]: To check if a PR will succeed in the SecRel workflow, [test PRs in the SecRel pipeline](https://github.com/department-of-veterans-affairs/abd-vro-internal/wiki/Secure-Release-process#to-test-prs-in-the-secrel-pipeline).

PR #597, opened 2022-10-30
VRO Console: monitor Camel messages

yoomlam:
## What was the problem?

As a follow-on to #531, we'd like monitor claims as they are being processed along Camel routes for diagnosing problems particularly in prod.
The RabbitMQ Management UI is not sufficient.

- #562 

## How does this fix it?

Adds [Camel wiretaps](https://camel.apache.org/components/3.18.x/eips/wireTap-eip.html) at critical points in Camel routes to monitor messages and their content.

## How to test this PR
- `export COMPOSE_PROFILES=debug` -- the console container only starts if the docker-compose `debug` profile is enabled.
- Start VRO. Note the `vro-console-1` container; check out the logs: `docker logs vro-console-1`.
- Attach to the container: `docker attach vro-console-1`
- On the `groovy:000>` prompt, try the following:
```groovy
// Check out http://localhost:15672/#/queues for current queues and see how `console-*` queues are added as the following commands are run.

// Initially no routes in the CamelContext of the Console container
groovy:000> camel.routes
===> []

groovy:000> wireTap claim-submitted
// Now, submit a claim using Swagger
// Expect to see a log message

groovy:000> wireTap generate-pdf
// Now, generate a pdf using Swagger
// Expect to see a log message

groovy:000> camel.routes
===> [
  Route[rabbitmq://tap-generate-pdf?exchangeType=topic&queue=console-generate-pdf -> null], 
  Route[rabbitmq://tap-claim-submitted?exchangeType=topic&queue=console-claim-submitted -> null]
]
```
- Stopping the Console container will remove the `console-*` queues.
- Check that the `app` container logs has no errors.


console/src/main/groovy/gov/va/vro/consolegroovy/VroConsoleShell.groovy
```diff
@@ -33,16 +34,17 @@ class VroConsoleShell {
   VeteranRepository veteranRepository
 
   String submitSeda() {
-    return producerTemplate.requestBody("seda:foo", "Hello", String.class)
+    return producerTemplate.requestBody("seda:foo", "Hello", String)
   }
```
yoomlam, line 38: Placeholder for future capability to inject messages into Camel routes.

service/provider/src/main/java/gov/va/vro/service/provider/camel/PrimaryRoutes.java
```diff
@@ -64,9 +64,18 @@ private void configureRouteClaimSubmit() {
         // Use Properties not Headers
         // https://examples.javacodegeeks.com/apache-camel-headers-vs-properties-example/
         .setProperty("diagnosticCode", simple("${body.diagnosticCode}"))
+        .wireTap(wireTapTopicFor("claim-submitted"))
```
yoomlam, line 67: Declare some wireTap points for the Console to subscribe to.

au2 (approved):
Works as described. Did not see anything in the code. Maybe do log.info instead of log.warn for the Execute args. Warning indicates some type of error to me. Also looks like log messages is 1000 characters long and gets clipped for longer. Not sure if there is a setting in the console for it.

yoomlam:
@au2 Hmmm. Maybe I'll also save the JSON to a file (once we are allowed to store PII), using a timestamp and claim id as the filename.

sethdarragile6 (approved):
LGTM (looks groovy to me)

PR #598, opened 2022-10-31
Camel: assign name to pdfRoute's queue

yoomlam:
## What was the problem?

If no queue name is provided, Camel assigns a numeric name to the queue for `pdfRoute`, making debugging difficult.

## How does this fix it?

Set a name for the queue.

## How to test this PR

View the queue names at http://localhost:15672/#/queues before and after the PR.
Ensure pdf generation and fetching still work.


dimitri-amida (approved):
LGTM. The pdf functionality still works, since the exchange is set up so that the queue name and routing key are the same.

anthonyvelazquez (approved):

PR #603, opened 2022-10-31
update docker configs to not install as user pip

advcoder:
publish is not running


au2:
@advcoder if this is important you need to do the same change in assessclaim7701 Dockerfile.

PR #614, opened 2022-11-01
VRO Console: connect to Redis

yoomlam:
## What was the problem?

As a follow-on to #531, we'd like examine contents of Redis in prod.

Associated tickets or Slack threads:

- #562
- #563 

## How does this fix it?

Adds Redis connection to the VRO Console.

## How to test this PR
- `export COMPOSE_PROFILES=debug` -- the console container only starts if the docker-compose `debug` profile is enabled.
- Start VRO. Note the `vro-console-1` container; check out the logs: `docker logs vro-console-1`.
- Generate a pdf using Swagger so that the pdf is added to Redis
- Attach to the container: `docker attach vro-console-1`
- On the `groovy:000>` prompt, try the following:
```groovy
groovy:000> redis.keys "*"
===> [1234_type, 1234]

// "1234" is the claim id
groovy:000> redis.get "1234_type"
===> hypertension

groovy:000> redis.get "1234"
// Shows the base64-encoded PDF contents

// Using RedisTemplate redisT
groovy:000> ops=redisT.opsForValue()
groovy:000> ops.get "1234_type"
===> hypertension
groovy:000> ops.get "1234"
// Shows the base64-encoded PDF contents
```

PR #615, opened 2022-11-01
Update pdf-generator to use Redis hash

tjheffner:
## What was the problem?
pdf generator stored multiple related items from a claim in separate keys

[Python code](https://github.com/department-of-veterans-affairs/abd-vro/blob/9a1ad2222ff5bd6c213d8ad3f34ecf7e5827a499/service-python/pdfgenerator/src/lib/redis_client.py#L29) saves the generated PDF doc as a [Redis String](https://redis.io/docs/data-types/#strings) with the lookup key being the claim id. 
1.  The key should probably be more descriptive like `1234-generated_pdf` , where `1234` is the claim id.
2.  The Python code is also saving another Redis String with key `1234_type` and value `hypertension`. Since these 2 Redis String entries are related, I'm wondering if they should be stored as a Redis Hash instead.

## How does this fix it?[^1]
Updates pdf generator to store related items to a single claim under one hash. Behaves as before otherwise.

## How to test this PR
- locally pulled in #614 to use redis in the console container as well
- generate a pdf using swagger (`test-key-01` to authorize locally). check generated submission in swagger also, this should work the same as before.

- inside the console container (`docker attach vro-console-1`), 
```
groovy:000> redis.keys "*"
===> [claim-1234]
groovy:000> redis.hlen("claim-1234")
===> 2
groovy:000> redis.hget("claim-1234", "type")
===> hypertension
groovy:000> redis.hget("claim-1234", "pdf")
===> JVBERi0xLjQKMSAwIG9iago8PAovVGl0bGUgKP7/KQovQ3JlYX ... (truncated base64 encoding of the generated pdf)
```

### DIFF
--- a/service-python/pdfgenerator/src/lib/queues.py
+++ b/service-python/pdfgenerator/src/lib/queues.py
@@ -31,7 +31,7 @@ def on_generate_callback(channel, method, properties, body):
         # logging.info(f"Variables: {variables}")
         template = pdf_generator.generate_template_file(diagnosis_name, variables)
         pdf = pdf_generator.generate_pdf_from_string(template)
-        redis_client.save_hash_data(f"{claim_id}-pdf", mapping={"contents": base64.b64encode(pdf).decode("ascii"), "type": diagnosis_name})
+        redis_client.save_hash_data(f"{claim_id}-pdf", mapping={"contents": base64.b64encode(pdf).decode("ascii"), "diagnosis": diagnosis_name})
         logging.info("Saved PDF")
         response = {"claimSubmissionId": claim_id, "status": "COMPLETE"}
     except Exception as e:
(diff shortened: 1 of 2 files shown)


service-python/pdfgenerator/src/lib/queues.py
```diff
@@ -31,8 +31,7 @@ def on_generate_callback(channel, method, properties, body):
         # logging.info(f"Variables: {variables}")
         template = pdf_generator.generate_template_file(diagnosis_name, variables)
         pdf = pdf_generator.generate_pdf_from_string(template)
-        redis_client.save_data(claim_id, base64.b64encode(pdf).decode("ascii"))
-        redis_client.save_data(f"{claim_id}_type", diagnosis_name)
+        redis_client.save_hash_data(f"claim-{claim_id}", mapping={"pdf": base64.b64encode(pdf).decode("ascii"), "type": diagnosis_name})
```
yoomlam, line 34: Since the hash contents is only about the pdfgenerator, how about these keys:
```suggestion
        redis_client.save_hash_data(f"{claim_id}-pdf", mapping={"contents": base64.b64encode(pdf).decode("ascii"), "diagnosis": diagnosis_name})
```

yoomlam (changes requested):

anthonyvelazquez (approved):
LGTM. Learned something new as well. DIdn't know about Redis hashes 👍

PR #654, opened 2022-11-07
Enable spotless on *.gradle files

yoomlam:
## What was the problem?

Spotless linter was not being applied to `build.gradle` files.

## How does this fix it?

Add a spotless `groovyGradle` section to all projects

## How to review this PR
Enable "Hide whitespace" when reviewing to see that only formatting has been changed for all files except for file `/build.gradle`:
image

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -54,14 +54,14 @@ allprojects {
     rules = ['unused-dependency']
     // criticalRules will fail the build in the event of a violation
     criticalRules = [
-        'minimum-dependency-version',
-        'overridden-dependency-version',
-        'unused-exclude-by-dep',
-        'unused-exclude-by-conf',
-        'dependency-parentheses',
-        'duplicate-dependency-class',
-        'archaic-wrapper',
-        'all-nebula-renames'
+      'minimum-dependency-version',
+      'overridden-dependency-version',
+      'unused-exclude-by-dep',
+      'unused-exclude-by-conf',
+      'dependency-parentheses',
+      'duplicate-dependency-class',
+      'archaic-wrapper',
+      'all-nebula-renames'
     ]
   }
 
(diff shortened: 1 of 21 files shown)


wlin2019 (approved):
LGTM

yoomlam:
> I do not think this caught everything. db-init/build.gradle for example does not appear to be linted.

I see -- some subprojects don't apply the `spotless` plugin.
```
❯ g db-init:spotlessApply
...
FAILURE: Build failed with an exception.

* What went wrong:
Task 'spotlessApply' not found in project ':db-init'.
```

Let me see a good way to address this.

build.gradle
```diff
@@ -1,206 +1,207 @@
...
-    def tb = taskBuilder('spotlessCheck')
-    codeModules.each { dependsOn tb(it) }
-}
-
-def spotlessApply = tasks.register('spotlessApply') {
-    def tb = taskBuilder('spotlessApply')
-    codeModules.each { dependsOn tb(it) }
+  // https://axion-release-plugin.readthedocs.io/en/latest/configuration/basic_usage
+  project.version = scmVersion.version
 }
```
yoomlam, line 155: Adding the `spotless` plugin adds these 2 tasks.
(Note: we don't (currently) use pre-commit hooks, so removing these lines has no impact)

db-init/build.gradle
```diff
@@ -3,53 +3,55 @@
...
-            'servicepassword': 'also-not-the-service-password',
-            'adminpassword': 'also-not-the-admin-password'
-    ]
-    configurations = [ 'flywayMigration' ]
+  url = 'jdbc:postgresql://vro:not-the-password@localhost:5432/claims'
+  user = 'vro_user'
+  password = 'not-the-password'
+  locations = [
+    "filesystem:${project.buildDir}/docker/database/migrations"
+  ]
```
yoomlam, line 45: Fixed by `spotless`.

au2 (approved):

PR #676, opened 2022-11-10
Polish API spec for external use

yoomlam:
## What was the problem?

The VRO API point of contact should be a VA employee. Premal agreed.

## How does this fix it?[^1]

Set the VRO API point of contact to be Premal.

Also, adjust the pod names in LHDI's EKS from `vro-api` to just `vro`.
Current names:
image


au2 (approved):
Did not really test deployment but LGTM otherwise.

PR #678, opened 2022-11-10
Fix java-install-tools action step

yoomlam:
## What was the problem?

PR #521 incorrectly updated a line, causing the [cache key](https://github.com/department-of-veterans-affairs/abd-vro/actions/caches) to be just `hadolint-`:
image

## How does this fix it?[^1]

Fix the echo command.

## How to test this PR
Check [cache](https://github.com/department-of-veterans-affairs/abd-vro/actions/caches) after the action runs

### DIFF
--- a/.github/actions/install-java-tools/action.yml
+++ b/.github/actions/install-java-tools/action.yml
@@ -7,7 +7,7 @@ runs:
     # http://man7.org/linux/man-pages/man1/date.1.html
     - name: Get Date
       id: get-date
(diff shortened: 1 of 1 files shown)


yoomlam:
https://github.com/department-of-veterans-affairs/abd-vro/actions/runs/3441043193/jobs/5740175333#step:4:73 shows correct cache key:
> Cache not found for input keys: hadolint-20221110

and [caches](https://github.com/department-of-veterans-affairs/abd-vro/actions/caches) shows
image

PR #679, opened 2022-11-11
Use gradle-build-action for better caching

yoomlam:
## What was the problem?

[Our cache](https://github.com/department-of-veterans-affairs/abd-vro/actions/caches) exceeds 10GB and our oldest cache is only 2 days old (except for one 4.5MB 7-day-old cache)
image

This suggest better caching can be done.

## How does this fix it?

Try `gradle-build-action` to see if there are any improvements.
https://github.com/gradle/gradle-build-action#why-use-the-gradle-build-action

### DIFF
--- a/.github/actions/setup-vro/action.yml
+++ b/.github/actions/setup-vro/action.yml
@@ -10,7 +10,15 @@ runs:
       with:
         distribution: temurin
         java-version: '17'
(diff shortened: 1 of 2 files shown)


yoomlam:
Several caches created:
image

PR #680, opened 2022-11-11
Minimize GitHub Action checks and improve speed

yoomlam:
## What was the problem?

GitHub Action checks were running unnecessary actions for PRs

## How does this fix it?[^1]

Minimize GitHub Action checks and improve speed for checks on PRs

### DIFF
--- a/.github/workflows/publish-dev.yml
+++ b/.github/workflows/publish-dev.yml
@@ -2,14 +2,14 @@ name: Publish dev images
 
 on:
   push:
-    branches: [ main, develop ]
+    branches: [ develop ]
 
   # Allow manual triggering
   workflow_dispatch:
 
   workflow_call:
     secrets:
-      DEV_KUBE_CONFIG: 
+      DEV_KUBE_CONFIG:
         required: true
 
 env:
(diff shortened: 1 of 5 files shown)


.github/workflows/end2end-test.yml
```diff
@@ -1,10 +1,15 @@
...
-    branches:
-      - "*"
+    branches: [ main, develop ]
+    # Limit to certain PR event types since this action doesn't need to run for every commit
+    types: [ready_for_review, review_requested]
+
+  # When changes are pushed to special branches
+  push:
+    branches: [main, develop]
+
```
yoomlam, line 12: @au2 Are you good with this change?

.github/workflows/publish-dev.yml
```diff
@@ -2,14 +2,14 @@ name: Publish dev images
 
 on:
   push:
-    branches: [ main, develop ]
+    branches: [ develop ]
```
yoomlam, line 5: Since branch `main` deploys to PROD env, it's confusing to deploy `main` to DEV env. cc: @advcoder

.github/workflows/end2end-test.yml
au2: ok. i think there is a path in this setup to go update develop after changes without end to end test run but hopefully people with request reviews after changes.

yoomlam: Yea, I'm not too concerned since the end-to-end test will run against the `develop` branch after the PR is merged, and we'll get alerted if it fails.

service-python/build.gradle
```diff
@@ -30,3 +36,8 @@ tasks.register("pytest", PythonTask) {
   command = "--cov-report term-missing --cov=. --cov-fail-under=80 --cov-config=../.coveragerc"
   dependsOn pyflake8
 }
+
+check {
+  // TODO: uncomment after fixes are made
+  // dependsOn pytest
```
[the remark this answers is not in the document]
anthonyvelazquez: Fixed up the linting/formatting errors. Just failing some tests @jbregstone Seems like its an issue with your health assess queues. Could you take a look?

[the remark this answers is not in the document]
jbregstone: I did not realize that python tests were not running on push. Anyways, it should be fixed now.

PR #707, opened 2022-11-17
MCP-2043: add tests for MAS routes - possible paths

dimitri-amida:
## What was the problem?

MAS integration follows a complex workflow combining several external services. Calls to services can result in a fork of the logic path or in an exception. We need to ensure that the software correctly covers the happy paths and can also handle exceptional conditions.

- [MCP-2043](https://amida.atlassian.net/browse/MCP-2043)

## How does this fix it?[^1]

This ticket adds positive and negative test cases for Mas integration and adds some more error handling. Whereas the test scenarios do not cover every single case, the framework is established so we can build upon it and make sure future code changes do not introduce regressions.

## How to test this PR
- Execute "automatedClaim" with a collection that has sufficient evidence. 
- Execute "automatedClaim" with a collection that has insufficient evidence. At the moment "orderExam" is not implemented so it will throw an exception


app/src/test/java/gov/va/vro/controller/DevControllerTest.java
```diff
@@ -1,4 +1,4 @@
-package gov.va.vro;
+package gov.va.vro.controller;
```
dimitri-amida, line 1: Moved controller tests to the controller package

app/src/test/java/gov/va/vro/service/MasServiceTest.java
```diff
@@ -1,7 +1,8 @@
-package gov.va.vro;
+package gov.va.vro.service;
```
dimitri-amida, line 1: moved service tests to service package

service/provider/build.gradle
```diff
@@ -25,6 +25,7 @@ dependencies {
   implementation "org.springframework.boot:spring-boot-starter-data-redis"
   implementation "org.apache.camel.springboot:camel-spring-redis-starter:${camel_version}"
   implementation "org.apache.camel.springboot:camel-jsonpath-starter:${camel_version}"
+  implementation "org.apache.camel.springboot:camel-bean-validator-starter:${camel_version}"
```
dimitri-amida, line 28: validate beans as part of the route

service/provider/src/main/java/gov/va/vro/service/provider/camel/MasIntegrationRoutes.java
```diff
@@ -196,6 +165,22 @@ public void process(Exchange exchange) throws Exception {
         .end();
   }
 
+  private static Function<List<Exchange>, HealthDataAssessment> combineExchangesFunction() {
+    return exchanges -> {
+      for (Exchange exchange : exchanges) {
+        if (exchange.isFailed()) {
+          throw new MasException(
+              "Failed to collect evidence", exchange.getException(Throwable.class));
```
dimitri-amida, line 173: This is the only way to tell that one of the paths has failed

service/provider/src/main/java/gov/va/vro/service/provider/mas/service/MasCollectionService.java
```diff
@@ -93,55 +89,39 @@ public static HealthDataAssessment combineEvidence(
...
+    log.info("combineEvidence >> MAS : " + ((masApiEvidence != null) ? "not null" : "null"));
     AbdEvidence compositeEvidence = new AbdEvidence();
     compositeEvidence.setBloodPressures(
         merge(
-                lighthouseEvidence != null ? lighthouseEvidence.getBloodPressures() : null,
-                masApiEvidence != null ? masApiEvidence.getBloodPressures() : null)
-            .stream()
-            .flatMap(s -> Stream.ofNullable(s))
-            .distinct()
-            .collect(Collectors.toList()));
```
dimitri-amida, line 106: implemented remove duplicate logic by putting in a set. See "merge"

service/provider/src/main/java/gov/va/vro/service/provider/mas/service/MasCollectionService.java
```diff
@@ -93,55 +89,39 @@ public static HealthDataAssessment combineEvidence(
...
+            lighthouseEvidence != null ? lighthouseEvidence.getProcedures() : null,
+            masApiEvidence != null ? masApiEvidence.getProcedures() : null));
     lighthouseAssessment.setEvidence(compositeEvidence);
 
     return lighthouseAssessment;
   }
 
   private static <T> List<T> merge(List<T> list1, List<T> list2) {
-    List<T> result = new ArrayList<>();
+    Set<T> result = new LinkedHashSet<>();
```
dimitri-amida, line 117: This will take care of dupes

wlin2019, line 117: It is good to use Set to avoid dup. Will it be better to have Lighthouse and MAS services return SETs and avoid to have this "merge" process? On the other hand, those Abd Evidence objects have a comparable defined. I'm not sure whether this will have an unintended impact.

dimitri-amida: Even if they return sets, we'd have to combine them, so there is a still a merge. Set does not care about comparable, only equals and hashCode.

Ideally, comparable equality should be compatible with equals() equality. I don't think this is the case here, so perhaps instead of making them Comparable we should use a Comparator. That's a different topic.

ianwolson (approved):

service/provider/build.gradle
yoomlam: This has introduced a vulnerability -- issue #735 created.

dimitri-amida: This must be the straw that broke the camel's back :) We don't necessarily need this library. We can perform validation by other means. Let me know if you'd like me to remove it.

yoomlam: Yes, let's replace it. The alert states:
> org.glassfish:javax.el is deprecated, users can move to use org.glassfish:jakarta.el instead where this issue is first fixed in version 3.0.4.

PR #574, opened 2022-10-24
MCP-2034 Evidence Summary PDF

anthonyvelazquez:
## What was the problem?

Adding a new PDF based on iMVP Evidence Summary

Associated tickets or Slack threads:

- [MCP-1958](https://amida.atlassian.net/browse/MCP-1958)
- [MCP-2034](https://amida.atlassian.net/browse/MCP-2034)
- [MCP-2035](https://amida.atlassian.net/browse/MCP-2035)

## How does this fix it?[^1]

Want to merge some of the iMVP changes before this PR gets too large. Future PRs will be broken down into easier to test sections

## How to test this PR
- Generate a PDF using code `0000`. 
- As long as PDF generates, then this PR should be good to go.

PR #712, opened 2022-11-17
Add executable script to lib.pdf_generator for visual QA

nanotone:
## What was the problem?
I figured it could be useful for visual QA purposes to allow PDFGenerator to be called from command line with minimal setup (and essentially no infrastructure). If there's already an easy way to do this, feel free to close this PR -- it was more of a way for me to get familiarized with the pdf generation code.

- [Slack thread](https://dsva.slack.com/archives/C01Q7979Z7D/p1668622315342869?thread_ts=1668617826.587689&cid=C01Q7979Z7D) of design feedback for visual tweaks

## How does this fix it?

Python [executable modules](https://wiki.python.org/moin/ExecutableModules) are a great way to expose library functionality to the command line by piggybacking onto the [python -m](https://docs.python.org/3/using/cmdline.html#cmdoption-m) invocation, so that a user doesn't have to open up a separate REPL. See the below section for an example command-line session.

## How to test this PR
```bash
$ brew install wkhtmltopdf  # on Mac, the only system dependency I needed

# if starting from scratch, set up a virtualenv for python deps
# it can be discarded, or kept around for future reuse
$ python3 -m venv ~/tmp-env
$ source ~/tmp-env/bin/activate
$ cd abd-vro/service-python/pdfgenerator/src
$ pip install -r requirements.txt

# running with no extra args prints a helpful usage guide
$ python -m lib.pdf_generator
usage: pdf_generator.py [-h] [-o OUTPUT] input
pdf_generator.py: error: the following arguments are required: input
$ python -m lib.pdf_generator ../examples/sample_hypertension.json -o output.pdf

$ open output.pdf  # on Mac, this just opens Preview
```

This PR also adds a `sample_hypertension.json` file to the examples directory. Executing the module on the sample produces this [output.pdf](https://github.com/department-of-veterans-affairs/abd-vro/files/10034761/output.pdf) (no surprises to see here).


yoomlam:
All I had to do:
```bash
cd service-python/pdfgenerator/src
python -m lib.pdf_generator ../examples/sample_hypertension.json
open output.pdf
```

service-python/pdfgenerator/examples/sample_hypertension.json
```diff
@@ -0,0 +1,98 @@
+{
```
yoomlam, line 1: Since this file is not referenced/used in the code, it may be accidentally deleted. Either add comments for how to use this file, reference it in a test, or use it as the default `input` value for the `__main__` block.

yoomlam (changes requested):

nanotone:
While addressing @yoomlam's feedback and starting to transfer some of this PR's text to the README, I noticed that #574 added a similar `__main__` script to generate PDFs locally.

@anthonyvelazquez I'm thinking about merging our two approaches into a one local script that has a bunch of configurable options. Do you prefer to keep it in `local_pdf_test.py` or should we use the `python -m` approach here?

anthonyvelazquez:
I think the `python -m` approach is preferable. I just made a quick file but I like your approach better

au2:
@nanotone is this ready to be pr'ed again?

nanotone:
@au2 No, this PR is on the backburner but I do have concrete plans to return to it. It's a bit stale but quite easy to reconcile with latest code; is it okay if we keep it open?

au2:
Sure. Just wanted to make sure you were not waiting for PR.

PR #747, opened 2022-11-28
MCP-2060: Condition processors error handling

jbregstone:
## What was the problem?

Error message was not indicative of the problem. If validation failed the error message was "no data found"

- [MCP-000](https://amida.atlassian.net/browse/MCP-000)

## How does this fix it?[^1]

Adds a status field to condition processors response to indicate errors. Clarifies what went wrong when errors occur.
## How to test this PR
- Step 1
- Step 2[^secrel]


au2 (commented):
Your PRs will need to start with the Ticket number going forward.

au2 (changes requested):
Your PRs will need to start with the Ticket number going forward.

api/src/main/java/gov/va/vro/api/responses/FullHealthDataAssessmentResponse.java
```diff
@@ -66,9 +66,10 @@ public class FullHealthDataAssessmentResponse {
    * @param errorMessage error message
    */
   public FullHealthDataAssessmentResponse(
-      String veteranIcn, String diagnosticCode, String errorMessage) {
+      String veteranIcn, String diagnosticCode, String errorMessage, String status) {
     this.veteranIcn = veteranIcn;
```
wlin2019, line 70: I don't think it is a good practice to design the object this way - mix the error information and valid data.

wlin2019 (approved):
I cannot comment on the python code and didn't test it. If it runs, it looks good to me.

api/src/main/java/gov/va/vro/api/responses/FullHealthDataAssessmentResponse.java
au2: @wlin2019 good point. @jbregstone we are not even using this object when there are errors. See GlobalExceptionHandler.  So there is no need for this new status and indeed the previous errorMessage fields in this rest response object.  errorMessage is never used and with HTTP status code values there is no need for a status in the response object. Use an intermediate object instead. We already have one in EvidenceWithSummary.  That has errorMessage as well it looks like.

controller/src/main/java/gov/va/vro/controller/DevController.java
```diff
@@ -51,7 +51,7 @@ public ResponseEntity<HealthDataAssessment> postHealthAssessment(
           objectMapper.readValue(responseAsString, HealthDataAssessment.class);
       if (response.getEvidence() == null) {
         throw new ClaimProcessingException(
-            claim.getClaimSubmissionId(), HttpStatus.NOT_FOUND, "No evidence found.");
+            claim.getClaimSubmissionId(), HttpStatus.INTERNAL_SERVER_ERROR, "No evidence found.");
```
au2, line 54: This controller will go away so this change is not really important.

controller/src/main/java/gov/va/vro/controller/VroController.java
```diff
@@ -129,6 +129,15 @@ public ResponseEntity<FullHealthDataAssessmentResponse> postFullHealthAssessment
...
+            "Response from condition processor returned error message: {}",
+            response.getErrorMessage());
+        throw new ClaimProcessingException(
+            claim.getClaimSubmissionId(),
+            HttpStatus.INTERNAL_SERVER_ERROR,
+            "Error in condition processor.");
+      }
       if (response.getEvidence() == null) {
         throw new ClaimProcessingException(
             claim.getClaimSubmissionId(), HttpStatus.INTERNAL_SERVER_ERROR, "No evidence found.");
```
au2, line 143: This error still does not make sense. If there is no evidence found I am assuming you are supposed to return empty lists. And I do not even think you would come here after the other changes in this PR. If there is a path here you should change the error messages.

controller/src/main/java/gov/va/vro/controller/VroController.java
```diff
@@ -129,6 +129,15 @@ public ResponseEntity<FullHealthDataAssessmentResponse> postFullHealthAssessment
...
       FullHealthDataAssessmentResponse response =
           objectMapper.readValue(responseAsString, FullHealthDataAssessmentResponse.class);
+      if (response.getStatus().equals("ERROR")) {
+        log.info(
+            "Response from condition processor returned error message: {}",
+            response.getErrorMessage());
+        throw new ClaimProcessingException(
+            claim.getClaimSubmissionId(),
+            HttpStatus.INTERNAL_SERVER_ERROR,
+            "Error in condition processor.");
```
au2, line 139: Condition processor probably will not make much sense to caller. Maybe "Internal error during processing data for the claim.".

service-python/assessclaimdc6602/src/lib/queues.py
```diff
@@ -18,7 +18,7 @@ def on_request_callback(channel, method, properties, body):
         response = main.assess_asthma(message)
     except Exception as e:
         logging.error(e, exc_info=True)
-        response = {"status": "ERROR", "evidence": {}, "evidenceSummary": {}}
+        response = {"status": "ERROR", "evidence": {}, "evidenceSummary": {}, "errorMessage": str(e)}
```
au2, line 21: Why empty objects? Null would be better I think.

au2 (changes requested):
Please see inline comments.

au2 (approved):

PR #749, opened 2022-11-28
MCP-2037 Metrics Page Improvements

ianwolson:
## What was the problem?
We want to include more count totals in the metrics service.  These counts include total assessment results, medications, total blood pressure readings, recent blood pressure readings, and procedures.

- [MCP-2037](https://amida.atlassian.net/browse/MCP-2037)

## How does this fix it?[^1]

Includes assessmentResult, medicationsCount, totalBpReadings, recentBpReadings, and proceduresCount to the claim-metrics endpoint

## How to test this PR
Step 1: populate db with claims and assessment results by executing the full health assessment endpoint
Step 2: execute claim-metrics endpoint
Step 3: Verify the total claims, assessmentResults, and EvidenceSummaryDocuments are the right counts.


au2 (changes requested):
1) Your PRs will need to start with Ticket number.
2) Since not everybody has access to the tickets you will need similar descriptions in the PR.

au2 (changes requested):
This is not really what I have in mind. But this can be useful.
1) I do not think there is any need to  for the number of individual medical counts. That is useful for individual claims (what I had in mind) but not overall. So I would remove those.
2) If assessment results is a measure of what I think it is that is useful. That is the number of times evidence is generated. That should be equal to the number of claims unless there were errors. If it is please leave it. If it is not please put the number of times evidence is generated. I would also change the name to numberOfEvidenceGeneration or something like that.
3) I would also add number of times pdf is generated. That should be equivalent to number of times the evidence was adequate.
4) Please work with Julian to count number of times the evidence was enough. I believe he has a flag for it now on the results. You can use that one. In theory that should be equal to the number in 3. We should display both.

In the future we might group these numbers per diagnostic ocde but for now that is not necessary. I will add another ticket to get details for a particular claim.

au2 (approved):

PR #765, opened 2022-12-01
MCP 2058: Add recordedDate to condition

jbregstone:
## What was the problem?

PDF design requires conditions to be sorted by recordedDate, not onsetDate.

Additional context from Carolyn, 
“““

It’s because the onset date column is usually inaccurate, so the only reliable data is  recorded date.

Background detail: We noticed when looking at production data that the onset date” and recorded date were usually exactly the same. We asked why, and the Lighthouse team explained that when there was no data in onset date, they insert recorded date

- [MCP-2058](https://amida.atlassian.net/browse/MCP-2058)

## How to test this PR
- Step 1 Build and run project. Try health-assessment endpoint for diagnosticCode "0" an verify that recordedDate is populated. 
- Step 2[^secrel]


anthonyvelazquez:
LGTM 
image

anthonyvelazquez (approved):

au2 (approved):
I am moving this although you cannot test with "0" in health-data-assessment because that does not call to the python service. full-health-data-assessment fails as expected with 0. I thought we have a v2 7101 so that probably should have changed so please write a ticket. Ed to end test is passing and this is a step forward so moving.

PR #778, opened 2022-12-05
MCP-2093: Fixed exception occuring during MAS processing

dimitri-amida:
## What was the problem?
MAS processing terminates with an exception suggesting that it did not properly communicate with Lighthouse.
The problem was occuring because the exchange was incorrectly set as one-way communication instead of request/response.

- [MCP-2093](https://amida.atlassian.net/browse/MCP-2093)

## How does this fix it?[^1]

Using "requestBody" instead of "sendBody" implicitly sets the exchange to the request/response

## How to test this PR
- Run "automatedClaim" with collection 350. Note that you will still get an exception in the "order exam" step because this endpoint is not implemented by IBM. But there will be no exception during lighthouse processing.

### DIFF
--- a/service/provider/src/main/java/gov/va/vro/service/provider/CamelEntrance.java
+++ b/service/provider/src/main/java/gov/va/vro/service/provider/CamelEntrance.java
@@ -60,12 +60,13 @@ public class CamelEntrance {
   }
 
   public void processClaim(MasAutomatedClaimPayload masAutomatedClaimPayload) {
-    producerTemplate.sendBody(
+    producerTemplate.requestBody(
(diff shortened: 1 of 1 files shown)


service/provider/src/main/java/gov/va/vro/service/provider/camel/MasIntegrationRoutes.java
```diff
@@ -220,4 +225,16 @@ private void configureOffRampClaim() {
         .routeId("mas-offramp-claim")
         .log("Request to off-ramp claim received");
   }
+
+  private void configureRouteGeneratePdf() {
```
dimitri-amida, line 229: bypass redis which is causing an exception

rajesh-Krishnasamy (approved):

au2 (approved):

PR #779, opened 2022-12-05
MCP-2056: Update server response for malformed JSON from 500 to 400

alan-amida:
MCP-2056: Update server response for malformed JSON from 500 to 400

## What was the problem?
If you enter an invalid JSON in our end points you will get a 500 Internal Error. We should instead get a 400 and error should indicate that JSON is wrong.

- [MCP-2056](https://amida.atlassian.net/browse/MCP-2056)

## How does this fix it?[^1]
As it says, when you attempt to submit malformed JSON against any endpoint, Spring should capture it first (before any other layer) and trigger a `HttpMessageNotReadableException` which then filters the capture to a `JsonParseException` and specifically triggers a `400 Bad Request` instead of the catch-all `500 Server Error` from before

## How to test this PR
- Step 1: Use curl or Postman and attempt to post sample malformed JSON to an endpoint:
POST to `http://localhost:8080/v1/full-health-data-assessment`
```JSON
{
  "veteranIcn": "9000682",
  "diagnosticCode": "7101",
  "claimSubmissionId": "1234",
  "disabilityActionType": "NEW",
  "extra-field": "hkjsdfjkhsfjfd",
}
```
You should get back a `400 Bad Request`

### DIFF
--- a/app/src/test/java/gov/va/vro/controller/BaseControllerTest.java
+++ b/app/src/test/java/gov/va/vro/controller/BaseControllerTest.java
@@ -18,6 +18,11 @@ public abstract class BaseControllerTest extends BaseIntegrationTest {
     return exchange(url, request, HttpMethod.POST, responseType);
   }
 
+  protected <I, O> ResponseEntity<O> post(
+      String url, I request, Map<String, String> headers, Class<O> responseType) {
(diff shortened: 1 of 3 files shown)


au2 (approved):
Tested  and it work. I would not have said unexpected error but that is not that important.

PR #796, opened 2022-12-06
MCP-2096 Claim Info Endpoint

ianwolson:
## What was the problem?
 We need to search claims using a particular claimSubmissionId or  a veteranIcn.  We also need to see the all the processed claims stored in the database using pagination. 

- [x] Find claim with claimSubmissionId
- [x] Find claims with veteranIcn
- [x] Find all claims using pagination
- [x] Remove fetch-claims endpoint

- [MCP-2096](https://amida.atlassian.net/browse/MCP-2096)

## How does this fix it?[^1]
Creates endpoints which will display claims based on the input parameters.

## How to test this PR
- Step 1: Run full health assessment with claimSubmissionId = 1234 
- Step 2:  Execute the claim-info endpoint using claimSubmissionId 1234
- Step 3: Execute the claim-info-for-veteran endpoint using veteranIcn = 
- Step 4: run full-health-assessment a few times with different claim submission ids each time.
- Step 5: Execute the claim-info-with-pagination endpoint and specify the offset as 0 and the pageSize as 10.
- Step 6: Verify all the claims are returned successfully.


au2 (changes requested):
1) Do not return a list for a single claim. It should be an object. If there are multiple information for a single claim make those lists.
2) There should be no field called errorMessgae in these jsons. Errors are handled in GlobalExceptionHandler with an independent Json,
3) offset and pageSize should not be in the path. Those are query parameters. The end point should be claim-info not claim-info-with-pagination.
4) Same with claim-info-for-veteran. It should be claim-info with veteranIcn as query parameter.

app/src/test/java/gov/va/vro/controller/VroControllerTest.java
```diff
@@ -201,21 +201,6 @@ void fullHealthAssessmentMalformedJson() {
...
-    String url = "/v1/fetch-claims";
-    String sampleRequestBody = "{ \"one\":\"one\", \"two\":\"two\",}";
-    var getResponseEntity = get(url, headers, ClaimProcessingError.class);
-    var postResponseEntity = post(url, sampleRequestBody, headers, ClaimProcessingError.class);
-    var putResponseEntity = put(url, sampleRequestBody, headers, ClaimProcessingError.class);
-    assertEquals(HttpStatus.OK, getResponseEntity.getStatusCode());
-    assertEquals(HttpStatus.METHOD_NOT_ALLOWED, postResponseEntity.getStatusCode());
-    assertEquals(HttpStatus.METHOD_NOT_ALLOWED, putResponseEntity.getStatusCode());
-  }
-
```
alan-amida, line 218: This was added to test the fix from sending 500 Internal Server Error to 405 Unsupported Method -- I didn't realize we were removing `/fetch-claims` -- so rather than removing this, can we just update the URL to something else like `/full-health-data-assessment`?

alan-amida (changes requested):

app/src/test/java/gov/va/vro/controller/VroControllerTest.java
ianwolson: yep, no problem!

alan-amida:
> 1. Do not return a list for a single claim. It should be an object. If there are multiple information for a single claim make those lists.
> 
>     2. There should be no field called errorMessgae in these jsons. Errors are handled in GlobalExceptionHandler with an independent Json,
> 
>     3. offset and pageSize should not be in the path. Those are query parameters. The end point should be claim-info not claim-info-with-pagination.
> 
>     4. Same with claim-info-for-veteran. It should be claim-info with veteranIcn as query parameter.

@ianwolson let me know if you have questions about Afsin's feedback here. I can help provide some context where I can.

api/src/main/java/gov/va/vro/api/resources/VroResource.java
```diff
@@ -258,9 +256,14 @@ ResponseEntity<ClaimInfoListResponse> claimInfoForVeteran(
             content = @Content(schema = @Schema(hidden = true)))
       })
   @ResponseStatus(HttpStatus.OK)
-  @Timed(value = "claim-info-with-pagination")
+  @Timed(value = "claim-info")
   @Tag(name = "Claim Metrics")
+  @RequestMapping(
+      value = "/claim-info",
+      params = {"offset", "pageSize"},
+      method = RequestMethod.GET)
```
alan-amida, line 264: I don't think this is necessary to redefine -- was it giving problems when you tried to redefine the original endpoint?

like instead of:
```
  @Timed(value = "claim-info")
  @Tag(name = "Claim Metrics")
  @RequestMapping(value = "/claim-info", params = "veteranIcn", method = RequestMethod.GET)
  ResponseEntity<ClaimInfoListResponse> claimInfoForVeteran(
      @RequestParam(name = "veteranIcn", required = false) String veteranIcn)
      throws MetricsProcessingException;

  @ApiResponses(
      value = {
        @ApiResponse(responseCode = "201", description = "Successful"),
        @ApiResponse(
            responseCode = "400",
            description = "Bad Request",
            content = @Content(schema = @Schema(hidden = true))),
        @ApiResponse(
            responseCode = "401",
            description = "Unauthorized",
            content = @Content(schema = @Schema(hidden = true))),
        @ApiResponse(
            responseCode = "500",
            description = "Claim Metrics Server Error",
            content = @Content(schema = @Schema(hidden = true)))
      })
  @ResponseStatus(HttpStatus.OK)
  @Timed(value = "claim-info")
  @Tag(name = "Claim Metrics")
  @RequestMapping(
      value = "/claim-info",
      params = {"offset", "pageSize"},
      method = RequestMethod.GET)
  ResponseEntity<ClaimInfoListResponse> claimInfoWithPagination(
      @RequestParam(name = "offset", required = false) int offset,
      @RequestParam(name = "pageSize", required = false) int pageSize)
      throws Exception;
```

did you try doing this?
```
  @Timed(value = "claim-info")
  @Tag(name = "Claim Metrics")
  @RequestMapping(
  value = "/claim-info",
  params = {"veteranIcn", "offset", "pageSize"},  // <--- just threw them all here
  method = RequestMethod.GET)
  ResponseEntity<ClaimInfoListResponse> claimInfo( // <--- method name changed to be more generic, since all of these query params are "optional" and handled further downstream (it looks like from here?) then I don't think we would need to explicitly define new paths based on veteranIcn or pagination meta being passed in or not
      @RequestParam(name = "veteranIcn", required = false) String veteranIcn)
      @RequestParam(name = "offset", required = false) int offset, // <--- added here
      @RequestParam(name = "pageSize", required = false) int pageSize)  // <--- added here
      throws MetricsProcessingException;

  @ApiResponses(
      value = {
        @ApiResponse(responseCode = "201", description = "Successful"),
        @ApiResponse(
            responseCode = "400",
            description = "Bad Request",
            content = @Content(schema = @Schema(hidden = true))),
        @ApiResponse(
            responseCode = "401",
            description = "Unauthorized",
            content = @Content(schema = @Schema(hidden = true))),
        @ApiResponse(
            responseCode = "500",
            description = "Claim Metrics Server Error",
            content = @Content(schema = @Schema(hidden = true)))
      })
  @ResponseStatus(HttpStatus.OK)
```

alan-amida: From Slack, for other observers' awareness:

I feel like since we're not doing anything special in our definition of our endpoint "signature", that we do not need to declare divergent paths right there. Especially if we're handling each case differently further downstream like that (or even if we're not tbh).
In theory, is it possible that a single veteranIcn should return enough results to warrant paginating?

ianwolson: Hey Alan! The only reason the paths are that way is because there are two different methods, one for finding claim info for a veteran ICN and one for pagination.

I could try and make it all one method where the pagination parametrs are required and the veteran ICN is not, this way there could be some logic where if a veteranIcn is present we just get the claims for that veteran.  Otherwise, we get all the claims. Let me know what you guys think.

alan-amida: I suppose for now we can let it pass, but then write up a refactor ticket later. If @au2 agrees, then I can approve. No need to hold up this functionality if it works as-is.

au2 (changes requested):
1) As it is claim-info get gives internal error. Parameter conditions "veteranIcn" OR "offset, pageSize" not met for actual request parameters. We should be able to call this with nothing specified. You will need to merge those two seperate cases.
2) You do not need the claim node in the response. It does not add any value. Just move what is under claim to top level. {
  **"claim"**: {
    "claimSubmissionId": "7001",
    "veteranIcn": "1012666073V986297",
    "contentionsCount": 1,
    "assessmentResultsCount": 1,
    "evidenceSummaryDocumentsCount": 1,
    "contentions": [
      "7101"
    ],
    "evidenceSummary": {
      "recentBpReadings": "3",
      "medicationsCount": "7",
      "totalBpReadings": "8"
    }
  }
}

au2 (changes requested):
Also please remove all commented out code.

au2:
@ianwolson Also I would change the the description of claim-info end points. "Retrieves claim specific metrics" and "Retrieves claim specific metrics for all claims". Do away with parameters in the description. Also you need defaults for paging parameters. 0 and 50 makes sense.

au2:
@ianwolson @alan-amida I started another branch to clean this up and ended up making too many changes. I am opening a PR for that branch. Please review that branch; i think it is much cleaner now with a lot of more testing. I am closing this one up.

PR #799, opened 2022-12-07
Replace VRO_NON_PROD with SWAGGER_ENABLED

yoomlam:
## What was the problem?

`VRO_NON_PROD` is too generic a name for how it is used. It is only used to disable Swagger UI for the prod deploys.

## How does this fix it?

Replace VRO_NON_PROD with SWAGGER_ENABLED. Set its value to `true` by default, so we don't have to set it unnecessarily (Reminder to remove `VRO_NON_PROD` from `setenv.sh` in the secrets repo). Remove extraneous references to it.

## How to test this PR
```
❯ helm install -n va-abd-rrd-dev --dry-run --debug abd-vro helmchart | grep SWAGGER_ENABLED -C 3
...
            - name: SWAGGER_ENABLED
              # enabled when not in prod
              value: true
...
```

On [this line](https://github.com/department-of-veterans-affairs/abd-vro/pull/799/files#diff-26474d778a95df886b201f2062af38f6b8534d875c411efe228b4415c97e96c3R267), change 
* `value: {{ ne .Values.environment "prod" }}` to
* `value: {{ eq .Values.environment "prod" }}`
and re-run the helm command above to see `value: false`.

### DIFF
--- a/.github/workflows/end2end-test.yml
+++ b/.github/workflows/end2end-test.yml
@@ -15,7 +15,6 @@ on:
 
 env:
   GITHUB_ACCESS_TOKEN: ${{ secrets.ACCESS_TOKEN }}
-  SWAGGER_ENABLED: true
   COMPOSE_PROJECT_NAME: vro
   LH_TOKEN_URL: https://sandbox-api.va.gov/oauth2/health/system/v1/token
   LH_ASSERTION_URL: https://deptva-eval.okta.com/oauth2/aus8nm1q0f7VQ0a482p7/v1/token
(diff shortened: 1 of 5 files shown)


advcoder (approved):

au2 (approved):
I will approve this but I think environment specific application.properties is a better solution. We already have ENV defined and that will define what application.properties to use. There is some info here https://www.baeldung.com/spring-profiles although I am not sure latest and greatest recommendation is. You can put that in after iMVP.

yoomlam:
> We already have ENV defined and that will define what application.properties to use.

In the code, I don't see how `ENV` affects the `application.properties/yml` file to use.

But it appears we should set `SPRING_PROFILES_ACTIVE` based on https://docs.spring.io/spring-boot/docs/current/reference/html/howto.html#howto.properties-and-configuration.set-active-spring-profiles

PR #805, opened 2022-12-07
Update server response for POSTing nonexistent ICN from 500 to 201

alan-amida:
## What was the problem?

- [MCP-2075](https://amida.atlassian.net/browse/MCP-2075)

## How to test this PR
- Step 1
- Step 2[^secrel]

### DIFF
--- a/app/src/test/java/gov/va/vro/controller/VroControllerTest.java
+++ b/app/src/test/java/gov/va/vro/controller/VroControllerTest.java
@@ -163,7 +163,7 @@ class VroControllerTest extends BaseControllerTest {
 
     var responseEntity =
         post("/v1/full-health-data-assessment", request, ClaimProcessingError.class);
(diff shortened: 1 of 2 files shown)


alan-amida:
Thinking about this more, I have reservations about issuing a `201 Created` response for something, even if we're POSTing it, that isn't found. I feel like the proper response here should be `404 Not Found` -- but this also assumes a standardized REST endpoint (where we would have something like a `/v1/veteran/${icn}` or something) where it feels way too late in the game to discuss.

au2:
Alan, you do not need this pr. This has been already been addressed in others. I think one of the prs assigned to you to review. 404 caused confusion since empty data is not an error. We are not getting data for the claim here, we are assessing if there is enough data for the claim. I understand the the difference is subtle but 404 caused confusion among the users of the API (and I would have been confused as well) hence the fix. But I get your point about 201. Maybe we should have used 200 for this and the happy path. We can discuss that later.

Issue #428, opened 2022-09-29
IncomingClaims: VRO receives hypertension CFI claims from MAS/VBAAP


yoomlam:
**User Story**

As MAS (a VRO client), I want to send single-issue hypertension CFI (claims for increase) and presumptive (new) claims to VRO, so that VRO can fast-track the claim if possible.

This ticket only provides the API endpoint and quick claim checks in order to return a response.

Requires integration: proxy VRO API through Lighthouse API so that MAS (external to VA's network) can call VRO (which has a non-static IP within LHDI) -- https://developer.va.gov/api-publishing (contact Premal to initiate this)

**Acceptance Criteria**

* VBAAP will send single-issue hypertension CFI (claims for increase) and presumptive (new) claims to VRO
* VRO checks if claim is within scope (shown as an `inScope?` decision point in [the workflow diagram](https://github.com/department-of-veterans-affairs/abd-vro/wiki/VRO-v2-Roadmap#workflow-diagram)).
  - "within scope" for VRO = claim is single-issue hypertension, INCREASE or NEW
  ```
  diagnosticCode: 7101 (hypertension)
  disabilityActionType: "INCREASE" or "NEW"
  ```
  - 11/16/2022 UPDATE: removed `disabilityClassificationCode: 3460 (hypertension) or 3370 (high blood pressure)`
  * If not in scope, VRO ignores it and sends Slack notification to channel `#benefits-vro-alerts`. Preferably respond to VBAAP with an error message.
     * Consider creating an `offrampClaim` Camel route, where claims can be fed whenever VRO cannot fast-track the claim for whatever reason. The route should use the [Slack component](https://camel.apache.org/components/3.18.x/slack-component.html) to send the Slack notification, and then route the claim to the `CompleteProcessing` component (described in #431) to double-check the claim and send it downstream.
     * When a claim is offramped, that offramp status and `offrampReason` should be saved to the DB. When #431 is implemented, consider saving to the DB once in the `CompleteProcessing` component.
  * If claim is within scope, it checks for 3 "anchors" - data elements that should already be set for claims coming from VBAAP:
     * claim's "station" (specifically `temporary station of jurisdiction` (TSOJ)) is set to `398`
     * claim's contention has special issue `Rating Decision Review - Level 1`
     * claim's contention has special issue `RRD`
     * If these are not present, send a Slack notification to channel `#benefits-vro-alerts` with details and stop processing. Preferably respond to VBAAP with an error message.
[...]

PR #806, opened 2022-12-08
MCP-2094 create end to end tests for mas

dimitri-amida:
## What was the problem?

Sometimes the application would break but we had no way to detect it.

- [MCP-2094](https://amida.atlassian.net/browse/MCP-2094)

## How does this fix it?[^1]

These post-docker tests help catch some problems during CI.

However, since the endpoints run asynchronously, the test do not cover a lot of backend scenarios. The pre-docker integration tests cover some of that. I also have some ideas about calling endpoints that execute synchronously, but they need a bit more thought and planning.

## How to test this PR
- N/A


alan-amida (commented):
Ran tests locally, code looks good to me. Approved.

This question is most likely out of scope for this PR:
To help me understand here, are we still returning `200 OK` responses even if what they try to POST is invalid in the scenarios of it being "out of scope" and "missing anchor?" -- because I see it looks like we're at least returning a response to the user via `masResponse.getMessage()` in these situations

Are we accepting the good parts and only partially rejecting? Or are we rejecting the entire request? Wondering why `200 OK` and not `400 Bad Request` in these cases

wlin2019 (approved):
I don't have much knowledge about this end2end test, and camel stuff. So I don't see anything wrong.

alan-amida (approved):

dimitri-amida:
> Ran tests locally, code looks good to me.
> 
> This is most likely out of scope for this PR: To help me understand here, are we still returning `200 OK` responses even if what they try to POST is invalid in the scenarios of it being "out of scope" and "missing anchor?" -- because I see it looks like we're at least returning a response to the user via `masResponse.getMessage()` in these situations
> 
> Are we accepting the good parts and only partially rejecting? Or are we rejecting the entire request? Wondering why `200 OK` and not `400 Bad Request` in these cases

It's a good question. The spec does not specify the exact behavior. Since there are several checks that result in the claim being rejected for autoprocessing, my interpretation is that such a rejection is not really a "bad request", just a request that does not meet the criteria. https://github.com/department-of-veterans-affairs/abd-vro/issues/428

PR #822, opened 2022-12-13
Remove env vars FLYWAY_TABLE and FLYWAY_TABLESPACE

yoomlam:
## What was the problem?

Environment variables `FLYWAY_TABLE` and `FLYWAY_TABLESPACE` were being passed around unnecessarily, adding to more maintenance burdens.

## How does this fix it?

Remove unnecessary env vars `FLYWAY_TABLE` and `FLYWAY_TABLESPACE`

## How to test this PR
Make sure VRO continues to run locally and deploy to LHDI


db-init/src/main/resources/flyway.conf
```diff
@@ -6,8 +6,8 @@ flyway.connectRetries=10
 flyway.defaultSchema=${FLYWAY_SCHEMA}
 flyway.schemas=${FLYWAY_SCHEMA}
 flyway.createSchemas=true
-flyway.table=${FLYWAY_TABLE}
-flyway.tablespace=${FLYWAY_TABLESPACE}
+flyway.table=schema_history
+flyway.tablespace=pg_default
```
yoomlam, line 10: This is the only place these configuration setting are needed, and the setting values do not need to be shared with other containers.

PR #831, opened 2022-12-14
Use @ActiveProfiles consistently

yoomlam:
## What was the problem?

Some tests with `@SpringBootTest` used `@ActiveProfiles("test")` but there was no corresponding `app/src/test/resources/application-test.yml`. The default `app/src/test/resources/application.yml` was being used.

Some tests with `@SpringBootTest` did not have `@ActiveProfiles("test")` so it was using `app/src/test/resources/application.yml`

`app/src/test/resources/application.yml` repeats many of the same settings as `app/src/main/resources/application.yml`. If the one under `main` changes, we'd have to remember to change the one under `test` -- if we forget, the main and test configurations will become unintentionally different.

## How does this fix it?

Have all tests with `@SpringBootTest` use `@ActiveProfiles("test")`.
Move `app/src/test/resources/application.yml` to `app/src/test/resources/application-test.yml`.
This way, the default `main/resources/application.yml` will be loaded, followed by `app/src/test/resources/application-test.yml`.

Since we want tests to be as close as possible to the deployed app, don't create `test/resources/application.yml`. If `test/resources/application.yml` exists, it will override `main/resources/application.yml`. 

As configured in `app/src/test/resources/application-test.yml`, `application-nonprod.yml` and `application-local.yml` will also be loaded because tests are nonprod and local.

## How to test this PR
All PR checks pass and [end2end test pass](https://github.com/department-of-veterans-affairs/abd-vro/actions/runs/3690792542).


yoomlam:
Merging PR before more conflicts

PR #836, opened 2022-12-14
MCP-2135 - Fix server response from 500 to 405 Unsupported on disallowed HTTP metthod for endpoint

alan-amida:
## What was the problem?

It looks like we are currently capturing Unsupported HTTP Request Method scenarios[*], but returning 500 Internal Server Error responses.

- [MCP-2135](https://amida.atlassian.net/browse/MCP-2135) (subtask of [MCP-2119](https://amida.atlassian.net/browse/MCP-2119))

## How does this fix it?[^1]

This fixes it so we return 405 Method Not Allowed instead

## How to test this PR
1. `curl` or postman any number of endpoints such as `/v1/fetch-claims` or `/v1/full-health-data-assessment`
2. Using the swagger docs as a reference (`http://localhost:8080/swagger`), test a Happy Path case and ensure you get the correct server response status code back from the server -- i.e. GET `/v1/fetch-claims` should return `200 OK`
3. Now test an Unhappy Path case and ensure you get `405 Method Not Allowed` -- i.e. POST `/v1/fetch-claims` should return `405 Method Not Allowed`
4. Test a couple of other routes to be sure

NOTE that the current `/v2/...` MAS routes will, unless authed, default to `401 Unauthorized` no matter which method you use, which is desirable behavior in this case.

### DIFF
--- a/controller/build.gradle
+++ b/controller/build.gradle
@@ -11,6 +11,9 @@ dependencies {
 
   implementation "gov.va.starter:exceptions"
 
+  // https://mvnrepository.com/artifact/javax.servlet/javax.servlet-api
+  compileOnly group: 'javax.servlet', name: 'javax.servlet-api', version: '4.0.1'
+
   // Needed?
   // implementation 'gov.va.starter:test-data'
   // implementation "gov.va.starter:entity-lifecycle-notifier"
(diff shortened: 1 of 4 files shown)


controller/build.gradle
```diff
@@ -11,6 +11,9 @@ dependencies {
 
   implementation "gov.va.starter:exceptions"
 
+  // https://mvnrepository.com/artifact/javax.servlet/javax.servlet-api
+  compileOnly group: 'javax.servlet', name: 'javax.servlet-api', version: '4.0.1'
+
```
alan-amida, line 16: This was necessary to pull in for `org.springframework.web.HttpRequestMethodNotSupportedException` to register with the app. Otherwise, without this, it would explain about `error: cannot access ServletException`

dimitri-amida (approved):
lgtm

ianwolson (approved):

PR #870, opened 2022-12-24
MCP-2164 Helper Cleanup

anthonyvelazquez:
## What was the problem?

Consolidate the additional lines that cleanup the PDF into the appropriate helper functions. Will make it easier to work with the logic when adding the new WeasyPrint package

- [MCP-2164](https://amida.atlassian.net/browse/MCP-2164)

## How does this fix it?

Move code out of pdf generator and into appropriate helper function

## How to test this PR
- Run containers and generate a PDF. There should be no differences


au2 (approved):
veteran info is gone. not sure if it is due to this. will approve and write a ticket

PR #871, opened 2022-12-24
MCP-2110 PDF Generate + Fetch Queue

anthonyvelazquez:
## What was the problem?

No issue exactly just adding an extra queue that generates + fetches the PDF all in 1 endpoint

- [MCP-2110](https://amida.atlassian.net/browse/MCP-2110)

## How does this fix it?

New endpoint that behaves the same as the generate and fetch endpoints combined. You can still use this endpoint and fetch endpoint to get the same PDF again if needed

## How to test this PR
- Start the containers and on Swagger, use the new `generate-fetch` endpoint.


service-python/pdfgenerator/src/lib/queues.py
```diff
@@ -57,6 +58,36 @@ def on_fetch_callback(channel, method, properties, body):
     channel.basic_publish(exchange=EXCHANGE, routing_key=properties.reply_to, properties=pika.BasicProperties(correlation_id=properties.correlation_id), body=json.dumps(response))
 
 
+def on_generate_fetch_callback(channel, method, properties, body):
+    try:
+        redis_client = RedisClient(redis_config)
+        pdf_generator = PDFGenerator(pdf_options)
+
```
au2, line 65: Can you modularize this? It looks like we literally copied two methods here. I would make commonalities functions?

controller/src/main/java/gov/va/vro/controller/VroController.java
```diff
@@ -99,6 +99,44 @@ public ResponseEntity<Object> fetchPdf(String claimSubmissionId) throws ClaimPro
...
+    log.info(
+        "Generating pdf for claim: {} and diagnostic code {}",
+        request.getClaimSubmissionId(),
+        request.getDiagnosticCode());
+    try {
+      GeneratePdfPayload model = generatePdfRequestMapper.toModel(request);
+      log.info(model.toString());
+      String response = camelEntrance.generateFetchPdf(model);
+      FetchPdfResponse pdfResponse = objectMapper.readValue(response, FetchPdfResponse.class);
+      log.info("RESPONSE from fetchPdf returned status: {}", pdfResponse.getStatus());
```
au2, line 114: Can we move commonalities with the existing endpoints to common methods?

api/src/main/java/gov/va/vro/api/resources/VroResource.java
```diff
@@ -105,6 +105,41 @@ ResponseEntity<GeneratePdfResponse> generatePdf(
   ResponseEntity<Object> fetchPdf(@PathVariable String claimSubmissionId)
       throws MethodArgumentNotValidException, ClaimProcessingException;
 
+  @Operation(
```
au2, line 108: I do not think we need a full new end point. Can we add a query param to the existing end point? You also need a "produces" here (take a look at the GET) or in the older post if you add query params.  If you are going to keep this end point can you change  it to a name instead of verb. Do not know what. "immediate-evidence-pdf" maybe?

au2 (changes requested):
PLease also add some tests, both to python and java.

service-python/pdfgenerator/src/lib/queues.py
anthonyvelazquez: Removed the extra function. Added a check in the generate function to decide if it needs to fetch right after.

controller/src/main/java/gov/va/vro/controller/VroController.java
anthonyvelazquez: Cleaned up into a `fetchPdfProcess` function that gets used for the fetch and immediate pdf routes

api/src/main/java/gov/va/vro/api/resources/VroResource.java
anthonyvelazquez: Kept the extra route since it would complicate the `generatePdf` java function and have to parse the 2 different types of responses. Renamed it to `immediate-evidence-pdf`

jbregstone (approved):
Endpoint works as expected! We should make a ticket for updating automatedClaim to use this route.

au2 (approved):

PR #876, opened 2022-12-27
MCP-2111 WeasyPrint PDF Package

anthonyvelazquez:
## What was the problem?

WKHTMLTOPDF runs on Webkit 2.2 which means a lot of the newer HTML features don’t work. There’s also issues with the engine it uses that causes a DPI issue and leads to inconsistency between the design and HTML page.

- [MCP-2111](https://amida.atlassian.net/browse/MCP-2111)

## How does this fix it?[^1]

Adds a new `pdfPackage` optional field to the PDF Generation request. If the field is not passed, it defaults to `wkhtmltopdf` which is what its currently running using. You can pass `weasyprint` to generate the same PDF using the new engine.

NOTE: This PR is just to add the functionality to use a different package. Since the end goal is to move over, subsequent PRs will work on fixing the templates so they render properly using WeasyPrint

## How to test this PR
- Build and run the containers. Use the swagger example for the PDF generate route and add `pdfPackage: "weasyprint"` to the request. The request should complete and you should be able to download the new PDF how you normally would using the fetch route. Note: The new PDF will not visually match the old version

### DIFF
--- a/api/src/main/java/gov/va/vro/api/requests/GeneratePdfRequest.java
+++ b/api/src/main/java/gov/va/vro/api/requests/GeneratePdfRequest.java
@@ -26,6 +26,10 @@ public class GeneratePdfRequest {
   @Schema(description = "Diagnostic code", example = "7101")
   private String diagnosticCode;
 
+  @NotBlank
+  @Schema(description = "PDF package", example = "wkhtmltopdf")
+  private String pdfPackage;
+
   @NotNull
   @Schema(description = "Veteran data for the pdf")
   private VeteranInfo veteranInfo;
--- a/service-python/pdfgenerator/src/lib/pdf_generator.py
+++ b/service-python/pdfgenerator/src/lib/pdf_generator.py
@@ -1,3 +1,4 @@
+import io
 import json
 import logging
 import os
@@ -5,6 +6,7 @@ import os
 import pdfkit
 from dateutil import parser
 from jinja2 import Environment, PackageLoader, select_autoescape
+from weasyprint import HTML
 
 from .helper_functions import *  # noqa: F403
 
@@ -13,8 +15,14 @@ lib_dir = os.path.dirname(__file__)
 
 class PDFGenerator:
 
-    def __init__(self, options):
+    def __init__(self, options, message_data=None):
         self.options = options
+        logging.info(message_data)
+        if message_data and "pdfPackage" in message_data and message_data["pdfPackage"].lower() in ["wkhtmltopdf", "weasyprint"]:
+            self.package = message_data["pdfPackage"]
+        else:
+            self.package = "wkhtmltopdf"
+        logging.info(f"Using PDF Package: {self.package}")
 
     def generate_template_variables(self, template_name: str, pdf_data: dict) -> dict:
         placeholder_variables = json.load(open(os.path.join(lib_dir, f"template_variables/{template_name}.json")))
@@ -43,11 +51,21 @@ class PDFGenerator:
         return generated_html
 
     def generate_pdf_from_string(self, template_name: str, html: str, data, output=False) -> bytes or bool:
-        base_toc_file_path = os.path.join(lib_dir, f"templates/{template_name}/base_toc.xsl")
-        if os.path.isfile(base_toc_file_path):
-            # Call a helper function that make adjustments to toc before creating
-            generated_toc_file_path = toc_helper_all(base_toc_file_path, data) # noqa: F405, E261
-            toc = {'xsl-style-sheet': generated_toc_file_path}
-            return pdfkit.from_string(html, output, options=self.options, toc=toc, verbose=True)
+        if self.package == "wkhtmltopdf":
+            base_toc_file_path = os.path.join(lib_dir, f"templates/{template_name}/base_toc.xsl")
+            if os.path.isfile(base_toc_file_path):
+                # Call a helper function that make adjustments to toc before creating
+                generated_toc_file_path = toc_helper_all(base_toc_file_path, data) # noqa: F405, E261
+                toc = {'xsl-style-sheet': generated_toc_file_path}
+                return pdfkit.from_string(html, output, options=self.options, toc=toc)
+            else:
+                return pdfkit.from_string(html, output, options=self.options)
         else:
-            return pdfkit.from_string(html, output, options=self.options, verbose=True)
+            buffer = io.BytesIO()
+            html_obj = HTML(string=html)
+            if output:
+                html_obj.write_pdf(output)
+                return
+            else:
+                html_obj.write_pdf(buffer)
+                return buffer.getbuffer()
--- a/app/src/test/java/gov/va/vro/controller/VroControllerTest.java
+++ b/app/src/test/java/gov/va/vro/controller/VroControllerTest.java
@@ -247,6 +247,7 @@ class VroControllerTest extends BaseControllerTest {
   void generatePdfInvalidInput() {
     var generatePdf = new GeneratePdfRequest();
     generatePdf.setClaimSubmissionId("1234");
+    generatePdf.setPdfPackage("wkhtmltopdf");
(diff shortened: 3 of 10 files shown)


api/src/main/java/gov/va/vro/api/requests/GeneratePdfRequest.java
```diff
@@ -26,6 +26,10 @@ public class GeneratePdfRequest {
   @Schema(description = "Diagnostic code", example = "7101")
   private String diagnosticCode;
 
+  @NotBlank
```
au2, line 29: Not sure why this marked @NotBlank. Isn't this supposed to be optional?

au2 (changes requested):
I cannot build. Looks like it is because test coverage issues. If I need to install something please specify in the pr description. Also please fix the conflicts.

 I would rather add this as a query parameter. It looks to be an odd thing to add to input data for me. Or even as an environment variable? Are we expecting these two to work side by side and selectable by the user? The pdf generator end point will not even be used directly in the MAS work flow.

jbregstone:
I am getting an error while building, I think It has to do with how my environment is configured using `pyenv`. I am going to look into it. https://github.com/Kozea/WeasyPrint/issues/1556

anthonyvelazquez:
> I cannot build. Looks like it is because test coverage issues. If I need to install something please specify in the pr description. Also please fix the conflicts.
> 
> I would rather add this as a query parameter. It looks to be an odd thing to add to input data for me. Or even as an environment variable? Are we expecting these two to work side by side and selectable by the user? The pdf generator end point will not even be used directly in the MAS work flow.

This was originally meant to replace `WKHTMLTOPDF` but have either selectable so we can compare the old vs new pdf render using the endpoint. Once everything is a match which [this PR](https://github.com/department-of-veterans-affairs/abd-vro/pull/1012) partially addresses, then we will remove the old library or keep it but remove it as a selectable option.

At the moment, the PDF defaults to `WKHTMLTOPDF` library which the design is already finalized for so adding this will not affect the current state of the PDF Generator.

After the final PDF discussions for what might come in the future, using `WeasyPrint` will be make the work across teams much easier. So its not a requirement for iMVP but will replace `WKHTMLTOPDF` in the future

alan-amida:
Seconding the test coverage errors, but looking into them, the root is WeasyPrint is having trouble importing external libraries:
```
OSError: cannot load library 'gobject-2.0-0': dlopen(gobject-2.0-0, 0x0002): tried: 'gobject-2.0-0' (no such file), '/System/Volumes/Preboot/Cryptexes/OSgobject-2.0-0' (no such file), '/usr/lib/gobject-2.0-0' (no such file, not in dyld cache), 'gobject-2.0-0' (no such file), '/usr/local/lib/gobject-2.0-0' (no such file), '/usr/lib/gobject-2.0-0' (no such file, not in dyld cache).  Additionally, ctypes.util.find_library() did not manage to locate a library called 'gobject-2.0-0'
```

```
WeasyPrint could not import some external libraries. Please carefully follow the installation steps before reporting an issue:
                https://doc.courtbouillon.org/weasyprint/stable/first_steps.html#installation
                https://doc.courtbouillon.org/weasyprint/stable/first_steps.html#troubleshooting 
```

etc

I see there are some recommendations for Windows user, but I'm on MacOS.

jbregstone:
> Seconding the test coverage errors, but looking into them, it seems that WeasyPrint is having trouble importing external libraries:
> 
> ```
> OSError: cannot load library 'gobject-2.0-0': dlopen(gobject-2.0-0, 0x0002): tried: 'gobject-2.0-0' (no such file), '/System/Volumes/Preboot/Cryptexes/OSgobject-2.0-0' (no such file), '/usr/lib/gobject-2.0-0' (no such file, not in dyld cache), 'gobject-2.0-0' (no such file), '/usr/local/lib/gobject-2.0-0' (no such file), '/usr/lib/gobject-2.0-0' (no such file, not in dyld cache).  Additionally, ctypes.util.find_library() did not manage to locate a library called 'gobject-2.0-0'
> ```
> 
> ```
> WeasyPrint could not import some external libraries. Please carefully follow the installation steps before reporting an issue:
>                 https://doc.courtbouillon.org/weasyprint/stable/first_steps.html#installation
>                 https://doc.courtbouillon.org/weasyprint/stable/first_steps.html#troubleshooting 
> ```
> 
> etc
> 
> I see there are some recommendations for Windows user, but I'm on MacOS.

I am on MacOS as well. @anthonyvelazquez shared this [documentation](https://xvik.github.io/gradle-use-python-plugin/latest/guide/configuration/#python-location) for the python Gradle plug-in we are using. The plug-in pulls in a local version of python to do the operation of setting up a virtualenv and installing packages. It seems that the weasyprint package is particular about how python was installed (homebrew. pyenv etc). My understanding is that is relies on packages included in Homebrew python.

alan-amida:
@anthonyvelazquez @jbregstone 

Gonna need to read up on `virtualenv`... but in the meantime, as a work-around, can we just straight up pull in the Python libs it expects rather than have it be dependent on the individual host environment? Assuming none of the many checks and tests in place wouldn't prevent us from doing so, we can put it off as tech debt to implement a more robust solution later.

anthonyvelazquez:
@alan-amida @jbregstone I updated the build config to use docker instead of your local python setup. This should fix the dependency issues but if you run into the same or new issues, let me know

jbregstone:
I encountered a problem when I tried to build.

`== BUILD FAILED ==
[docker] container 'python:3.10.6' (/crazy_cohen) stopped in 237ms

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':service-python:checkPython'.
> Python call failed: ../.gradle/python/bin/python -c exec("import sys;ver=sys.version_info;print(str(ver.major)+'.'+str(ver.minor)+'.'+str(ver.micro));print(sys.prefix);print(sys.executable)")
  
        OUTPUT:
                OCI runtime exec failed: exec failed: unable to start container process: exec: "../.gradle/python/bin/python": stat ../.gradle/python/bin/python: no such file or directory: unknown
`

service-python/pdfgenerator/src/lib/pdf_generator.py
```diff
@@ -12,8 +14,14 @@
 
 class PDFGenerator:
 
-    def __init__(self, options):
+    def __init__(self, options, message_data=None):
         self.options = options
+        logging.info(message_data)
```
jbregstone, line 19: Remove log before merging

alan-amida:
@anthonyvelazquez @jbregstone 

```
Mine built just fine.
./gradlew clean
source scripts/setenv.sh
./gradlew build check docker
```

Looks like it's spinning up a new container` testcontainers` ?

Mine is still running after it's done with it, looks like. Can we tear it down if this is reproducible behavior?

jbregstone:
I am getting an isort related error on build.

> 
> * What went wrong:
> Execution failed for task ':service-python:pyisort'.
>  Python call failed: python3 -m isort . --check --skip tests --skip .venv
>   
>         OUTPUT:
>                 Skipped 3 files
>                 ERROR: /usr/src/abd_vro/deactivate/bin/pip Imports are incorrectly sorted and/or formatted.
>                 ERROR: /usr/src/abd_vro/deactivate/bin/pip-3.8 Imports are incorrectly sorted and/or formatted.
> ``

anthonyvelazquez:
When you get the chance, give it another try.
Run `./gradlew service-python:cleanPython` then run your `./gradlew build check docker` as normal. The plugin was expecting to have access to `/bin/sh` which caused part of the issue and left the testcontainers orphaned

jbregstone:
> 

That worked for me, do you think you could include `./gradlew service-python:cleanPython` in `./gradlew clean` so that it is not necessary to add a step to the build process? 

Thanks, this PR is going to be really useful!

jbregstone (approved):

au2:
@anthonyvelazquez Please resolve conflicts

PR #883, opened 2023-01-03
MCP-2165 Veteran Info Fix

anthonyvelazquez:
## What was the problem?

After making some updates to the helper functions there was an issue with missing veteran info in the V1 PDFs.

- [MCP-2165](https://amida.atlassian.net/browse/MCP-2165)

## How does this fix it?

Originally, we had a `veteran_info` field that was being used for the templates and then a `veteranInfo` field that was getting provided from the data. Updated all V1 references of `veteran_info` to be `veteranInfo` which is what the Health Assessment endpoint provides

## How to test this PR
- Generate a PDF and verify that the veteran info is being displayed


au2 (changes requested):
hyperversion v2 appears to be referring to veteran_info still.

anthonyvelazquez:
> hyperversion v2 appears to be referring to veteran_info still.

Yep, this PR is only for V1. I'm leaving that one as is since I'm working on another PR thats fixing/adding more template functionality to V2.

au2 (approved):

PR #890, opened 2023-01-04
MCP-2174 Race Condition + Minor Cleanup

anthonyvelazquez:
## What was the problem?

WKHTMLTOPDF requires a file to generate the Table of Contents and it only accepts it in file format. This would lead to a race condition if multiple PDFs got requested at once since they would each read the `base_toc` template and then create a `generated_toc` file but would overwrite.

**Minor Cleanup:** Turning `verbose` off in the pdfgenerator since we don't need the extra messages and it slows down the generation process

- [MCP-2174](https://amida.atlassian.net/browse/MCP-2174)

## How does this fix it?

Rather than make a custom version of WKHTMLTOPDF that accepts the ToC in another format, we prepend `claimSubmissionId` to the `generated_toc` so there are no overwrite issues.

**Minor Cleanup:** Turning `verbose` off in the pdfgenerator since we don't need the extra messages and it slows down the generation process

## How to test this PR
- Try to create 2 PDFs at the same time with different `claimSubmissionId`
or
- After creating a PDF, enter the pdfgenerator container to see the file that was generated for the ToC to verify that it has the `claimSubmissionId`

### DIFF
--- a/service-python/pdfgenerator/src/lib/helper_functions.py
+++ b/service-python/pdfgenerator/src/lib/helper_functions.py
@@ -32,7 +32,7 @@ def toc_helper_all(toc_file_path, data):
     file_data = file_data.replace("{{file}}", data['fileIdentifier'])
     file_data = file_data.replace("{{date}}", f"{pytz.utc.localize(datetime.now()).strftime('%b. %d, %Y')}")
 
-    generated_toc_path = toc_file_path.replace("base_toc", "generated_toc")
+    generated_toc_path = toc_file_path.replace("base_toc", f"{data['claimSubmissionId']}_toc")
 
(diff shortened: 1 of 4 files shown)


anthonyvelazquez:
Just a note, this fix is temporary and will be removed once: https://github.com/department-of-veterans-affairs/abd-vro/pull/876 gets merged in. Using WeasyPrint removes all issues of race conditions since everything can be done in memory rather than actual files

jbregstone (changes requested):
Please add Claim ID in the logging message "Saved PDF"

au2 (approved):

jbregstone (approved):

PR #899, opened 2023-01-04
MCP-2137 Add sufficientEvidence flag to assessment results.

ianwolson:
MCP-2137 Add sufficientEvidence flag to assessment results.

## What was the problem?
We need to add a sufficient evidence flag to claims after MAS automated claim processes the claims.

- [MCP-2137](https://amida.atlassian.net/browse/MCP-2137)

## How to test this PR
- Step 1 Run MAS automated claim endpoint
- Step 2 Allow the MAS endpoint to do its thing, it takes about a minute.
- Step 3 Check the assessment result in the VRO database for the correct flag, [] = false and [v] = true.


au2 (changes requested):
I am assuming we should follow what we did for version 1 path and move this flag to contentions. I was under the impression that database was designed for multiple contentions per claim for future multiple issue claims. Either way this is a assesment result item not claim since it is being produced by assessment,

dimitri-amida:
@ianwolson in latest development, I created MasAssessmentResultProcessor. I think this is the best place to update the sufficiency flag, at the same time as inserting the assessment.

service/db/src/main/java/gov/va/vro/service/db/SaveToDbServiceImpl.java
```diff
@@ -105,6 +108,28 @@ public void insertEvidenceSummaryDocument(GeneratePdfPayload request, String doc
     claimRepository.save(claim);
   }
 
+  @Override
+  public void updateSufficientEvidenceFlag(
+      String claimSubmissionId, Boolean flag, String diagnosticCode) {
+    ClaimEntity claim = claimRepository.findByClaimSubmissionId(claimSubmissionId).orElse(null);
+    if (claim == null) {
+      log.warn("Could not find claim with given claimSubmissionId.");
```
wlin2019, line 116: Is it considered as a problem if the claim is not found? If yes, should it be handled other than just logging here?

service/provider/src/main/java/gov/va/vro/service/provider/services/MasAssessmentResultProcessor.java
```diff
@@ -24,5 +24,9 @@ public void process(Exchange exchange) throws Exception {
       return;
     }
     saveToDbService.insertAssessmentResult(evidence, diagnosticCode);
+    String claimSubmissionId = evidence.getClaimSubmissionId();
+    Boolean sufficiencyFlag = evidence.isSufficientForFastTracking();
+    saveToDbService.updateSufficientEvidenceFlag(
```
wlin2019, line 29: It looks like that the two variables created here don't do much. Just using evidence get methods in next line can reduce a couple of lines.

wlin2019 (approved):
I'm not familiar with the data model used, and the flow of setting up this flag. I have no objection to merge this PR. But here's some concerns.
1. I feel the DB update logic might be a bit too complicated. But if the data model is set that way, it probably won't be able to find a much better solution.
2. It doesn't seem have proper consideration for error handling.

au2:
@ianwolson is this pr still alive or are the new developments are taken over?

PR #902, opened 2023-01-05
MCP-2116 - Fix status code for missing collectionId field in requestBody

alan-amida:
## What was the problem?

When POSTing to `/v2/examOrderingStatus` endpoint:

- If collectionId field is missing, the response status code is 200 OK but the expected result  status code should be 400 Bad Request for missing required parameter.

- [MCP-2116](https://amida.atlassian.net/browse/MCP-2116)

## How does this fix it?[^1]

The previous Java field was using the `int` primitive with a `NotNull` validation. However, since `int` cannot be null, we must declare the field as `Integer` (which can be null) to account for this scenario.

## How to test this PR
- Step 1: Submit a POST request to `/v2/examOrderingStatus` with a payload that omits `collectionId`
  - Verify you receive a `400 Bad Request` with a message indicating that the field is required
- Step 2: Submit a valid POST request to `/v2/examOrderingStatus` with a valid payload
  - Verify you receive a `200 OK`


alan-amida:
`end2endTest` failing with the enforcement of the required field... will investigate in the morning.

alan-amida:
Huh. Test is working now. Will re-run it again a couple times in case of unexpected intermittent issues.

au2 (approved):

PR #908, opened 2023-01-06
MCP-2218 Add category for conditions

jbregstone:
## What was the problem?

Sufficiency decision should only consider encounter-diagnosis conditions. Previously `automatedClaim` route was not sending conditions from lighthouse to downstream services. 

As per design request, if there is no date we should leave the field blank or write "no associated date" on the PDF. 

- [MCP-2218](https://amida.atlassian.net/browse/MCP-2218)

## How to test this PR
- Step 1 audit messages going through automatedClaim routes.
- Step 2[^secrel]


service-python/assessclaimdc7101/src/lib/conditions.py
```diff
@@ -21,9 +21,13 @@ def conditions_calculation(request_body):
     for condition in veterans_conditions:
         condition_code = condition["code"]
         if "recordedDate" in condition.keys():
-            date = datetime.strptime(condition["recordedDate"], "%Y-%m-%d").date()
-            condition["dateFormatted"] = date.strftime("%m/%d/%Y")
-            condition_with_date.append(condition)
+            if condition["recordedDate"]:
+                date = datetime.strptime(condition["recordedDate"], "%Y-%m-%d").date()
```
dimitri-amida, line 25: what should be the expected behavior if the date cannot be parsed? As it is now, it will throw an uncaught exception

dimitri-amida (commented):
Looks OK to me but I think we should handle badly formatted dates the same way we handle missing dates.

service-python/assessclaimdc7101/src/lib/conditions.py
jbregstone: I will add a try and except

anthonyvelazquez (approved):

dimitri-amida (approved):
looks good

PR #918, opened 2023-01-09
Fix typo: slackExceptionWebhook

yoomlam:
## What was the problem?

`slackExceptionChannel != null` is checked twice -- copy-paste error?

## How does this fix it?

Replace second `slackExceptionChannel` with `slackExceptionWebhook`


dimitri-amida (approved):
good catch

PR #926, opened 2023-01-10
MCP-2226 PDF Template Naming Convention

anthonyvelazquez:
## What was the problem?

Originally PDFs were named based on the actual diagnosis name like `asthma`, `hypertension`, etc. This PR is to simplify that and match the naming convention of the assessment services

- [MCP-2226](https://amida.atlassian.net/browse/MCP-2226)

## How does this fix it?

Update the template file names so they're codes like `7101-v1.html`, etc...

## How to test this PR
- Create a PDF, there should be no issues with generation and fetching


jbregstone:
I am trending more in the direction of keeping the diagnosis names. See discussion [here](https://github.com/department-of-veterans-affairs/abd-vro/discussions/832)

anthonyvelazquez:
Ah I understand now, thanks. I'll close this PR for now

PR #930, opened 2023-01-11
MCP-2233 Python Health Check

anthonyvelazquez:
## What was the problem?

Added Python health check to `docker-compose` but its supposed to be added to the `Dockerfile` to comply with SDE

- [MCP-2233](https://amida.atlassian.net/browse/MCP-2233)


jbregstone (approved):

au2 (approved):
This is good for now. I future version should check on the health of the rabbit queues as well but not important for now. @anthonyvelazquez please write a low priority jira ticket.

PR #939, opened 2023-01-13
MCP-2240 Un-parsable date fix 

jbregstone:
## What was the problem?

If the code to parse a date failed then the date we show on the PDF was set to an empty string. 

- [MCP-2240](https://amida.atlassian.net/browse/MCP-2240)
- https://amida.slack.com/archives/C01Q7979Z7D/p1672172009573269

## How does this fix it?[^1]

In the case of an un-parsable date it is passed directly to the PDF without being parsed
## How to test this PR
- Step 1
- Step 2[^secrel]


anthonyvelazquez:
So we will need to change the date to something like `unknown-{date}`. Passing the date as it is, I won't know if I need to fill in the table cell with `No associated date / Non-confident date {date}` or if I should just display the date. Maybe adding a prefix/suffix or passing an additional field that lets me know its a bad date?

anthonyvelazquez:
Wait, I can do the parse on my side if needed or should we keep that logic in your service?

jbregstone:
> Passing the date as it is, I won't know if I need to fill in the table cell with No associated date / Non-confident date {date} or if I should just display the date.

I see that could be an issue. If it simplifies things, then we could move the parsing to the PDF side. I thought it would be better to have it in here since the code is already going through and reading each date. I could change `dateformatted` to start with with the string "unparsed". In the pdf service it could check if the `dateformatted` starts with that string and then handle the case accordingly. Keep in mind this is an edge case so I don't want to minimize designing around it.

anthonyvelazquez:
Since you already have the logic on your side, I think just passing `unparsed` is fine. I have helper functions for each code so I'll add that logic on my side in there so it doesnt affect future PDFs

jbregstone:
The four cases to account for in the pdf helper logic:

1. If there is no date at all -> `dateFormatted` is an empty string
2. If there is well formatted date -> `dateFormatted` looks like 04/06/1950
3. If there is a date but its empty -> `dateFormatted` is 'unparsed ()`
4. if there is a poorly formatted date (like "1999/m/d")-> 'dateFormatted` is 'unparsed (1999/m/d)`

jbregstone:
There is a possibility that unparsable dates will fail earlier in the system

anthonyvelazquez (approved):

au2 (approved):

PR #967, opened 2023-01-23
Add try catch to catch exception for BIP API service.

wlin2019:
Add code to handle BipException for automatedClaim process.

## What was the problem?
The BIP Claim API service can throw exceptions. The current code doesn't catch the exception and handle it. 
See error reported in MCP-2283.

MCP-2283

## How does this fix it?[^1]

Add handler for BipException in GlobalExceptionHandler.

## How to test this PR
- Step 1
- Step 2[^secrel]


service/provider/src/main/java/gov/va/vro/service/provider/bip/service/BipClaimService.java
```diff
@@ -46,31 +46,36 @@ public class BipClaimService {
...
+      var specialIssues =
+          contentions.stream()
+              .filter(BipClaimService::hasSpecialIssues)
+              .map(ClaimContention::getSpecialIssueCodes)
+              .flatMap(Collection::stream)
+              .map(String::toLowerCase) // Ignore case
+              .collect(Collectors.toSet());
+      return specialIssues.contains(SPECIAL_ISSUE_1) && specialIssues.contains(SPECIAL_ISSUE_2);
+    } catch (BipException e) {
+      log.error("BIP service call responds with exception. {}", e.getMessage(), e);
```
dimitri-amida, line 76: This is incorrect in my opinion. If the exception is thrown, a slack message will be sent and the end-user will know that the processing did not complete. Also the exception is stored in the database. If instead, we return false, meaning that the claim does not have the right anchors, which may not be true, we are hiding the real cause of the problem.

wlin2019: Per our conversation, I changed the approach and modified the PR description.

dimitri-amida (approved):
👍

rajesh-Krishnasamy (approved):

PR #1006, opened 2023-01-27
Feature/add prodtest config

alan-amida:
Adds an `application-prodtest.yml` file that we will use for PRODTEST environment


app/src/main/resources/application-prodtest.yml
```diff
@@ -0,0 +1,25 @@
...
+slack:
+  exception:
+    webhook: "${SLACK_EXCEPTION_WEBHOOK:#{null}}"
+    channel: benefits-vro-alerts
+
+springdoc:
+  swagger-ui:
+    enabled: true
+
+lhAPIProvider:
```
alan-amida, line 24: I assume this is how we can override other common-environment variables as defined in `application.yml`

au2: Yes.

app/src/main/resources/application-prodtest.yml
```diff
@@ -0,0 +1,29 @@
...
+  validateToken: NO
```
alan-amida, line 25: I assume this is how we can override other common-level environment variables as defined in application.yml

rajesh-Krishnasamy (approved):

au2 (approved):
Looks good to me.

PR #1018, opened 2023-01-30
MCP-1919 Rename RabbitMq queues from diagnostic code to diagnosis

ianwolson:
## What was the problem?
Rename the queues from 7101 → “hypertension” and 6602 → “asthma”, since we are adopting that naming convention for future conditions.

- [MCP-1919](https://amida.atlassian.net/browse/MCP-1919)

## How does this fix it?[^1]
Changes the queue names and adjusts tests to accept new queue names

## How to test this PR
- Start the app
- Navigate to http://localhost:15672/#/queues and confirm that the queue names contain the diagnoses and not their diagnostic codes


jbregstone:
@ianwolson This is not going to work unless we change the way messages are routed. Right now they are routed by their diagnostic code, to one of these queues using a string method. Look at `MasProcessingObject processClaim`. In the ticket I suggested looking into the issue to better understand the scope, I am not sure if it is actionable at the moment.

service-python/assessclaimdc6602v2/src/lib/settings.py
```diff
@@ -1,4 +1,4 @@
 queue_config = {
     "exchange_name": "health-assess-exchange",
-    "service_queue_name": "health-assess.6602v2",
+    "service_queue_name": "health-assess.asthma",
```
jbregstone, line 3: I would not change this yet until we decide on handling versions

service/provider/src/main/java/gov/va/vro/service/provider/camel/SlipClaimSubmitRouter.java
```diff
@@ -19,6 +20,17 @@ public class SlipClaimSubmitRouter {
   private static final long DEFAULT_REQUEST_TIMEOUT = 60000;
   public static final String NO_DIAGNOSTIC_CODE_ERROR = "No diagnostic code in properties.";
 
+  public static final Map<String, String> diagnoses;
+
+  static {
```
jbregstone, line 25: Can you use the diagnosis lookup map `java/gov/va/vro/service/provider/services/DiagnosisLookup.java`?

jbregstone (changes requested):
I like the direction this is going! I left comments on a couple of things

service/provider/src/main/java/gov/va/vro/service/provider/services/DiagnosisLookup.java
```diff
@@ -7,7 +7,7 @@ public final class DiagnosisLookup {
   private DiagnosisLookup() {}
 
   private static final Map<String, String> diagnosisMap =
-      Map.of("7101", "Hypertension", "6602", "Asthma");
+      Map.of("7101", "Hypertension", "6602", "Asthma", "1233", "1233");
```
jbregstone, line 10: can you try changing the test that says 1233 instead, making that 6602 and 7101?

ianwolson: One of them tests if the claim is out of scope and needs to be offramped, so I think Dimitri did this on purpose.  I did fix one of the tests though.

dimitri-amida: Is 1233 an actual value or just for testing? It it is for testing, it should not appear in the real Lookup

service-python/tests/assessclaim/test_queues.py
```diff
@@ -21,9 +21,9 @@
 
 @pytest.mark.parametrize(
     "queue, service_queue_name", [
-        (q6602v2, "6602v2"),
-        (q6602, "6602"),
-        (q7101, "7101"),
+        (q6602v2, "6202v2"),
```
jbregstone, line 24: Not sure how 6602 got changed to 6202

jbregstone (approved):
Nice job! I saw two small things and fixed them.

au2 (approved):

PR #1041, opened 2023-02-01
Update flyway Dockerfile image to 9.14-alpine

alan-amida:
Update flyway Dockerfile image to 9.14-alpine since previous 9.12.0-alpine version was removed from the vendor repository


alan-amida:
Added @ianwolson to kick off "when reviewer added" GitHub action workflows

anthonyvelazquez (approved):

PR #1107, opened 2023-02-14
Last Minute Database Improvements

hayleymdawson:
## What was the problem?
Yoom requested via Slack that the following items be added and it was agreed the PR would be broken out from current work to keep the size manageable.

- [x] Store the claimDetail.conditions.disabilityClassificationCode (provided in the /automatedClaim payload) in the DB as classification_code in the contention table

- [x] Store the claimDetail.conditions.name (provided in the /automatedClaim payload) in the DB as condition_name in the contention table

- [x] For the /examOrderingStatus endpoint, store examOrderDateTime field of the payload as ordered_at. This is for *all* statuses as clarified by subsequent thread discussion. 

(https://amida.slack.com/archives/C01Q7979Z7D/p1675982532875369)

## How does this fix it?
The work above is done to the spec requested. Audit details are added since we are already persisting those fields by request in the database

## How to test this PR
Run the /automatedClaim and /examOrderingStatus endpoints, validate new data is persisted in database. Claim payloads did not need modification beyond exposing fields to the rest of the package.

### DIFF
--- a/app/src/end2endTest/java/gov/va/vro/end2end/VroV2Tests.java
+++ b/app/src/end2endTest/java/gov/va/vro/end2end/VroV2Tests.java
@@ -228,7 +228,7 @@ public class VroV2Tests {
             "collectionStatus",
             "DRAFT",
             "examOrderDateTime",
-            "2022-12-08T17:45:61Z",
+            "2022-12-08T17:45:59Z",
             "eventId",
             "None");
     return objectMapper.writeValueAsString(payload);
--- a/app/src/end2endTest/resources/test-mas/claim-350-7101-outofscope.json
+++ b/app/src/end2endTest/resources/test-mas/claim-350-7101-outofscope.json
@@ -16,7 +16,7 @@
     "claimSubmissionDateTime": "2018-11-04T17:45:59Z",
     "claimSubmissionSource": "VA.GOV",
     "conditions": {
-      "name": "string",
+      "name": "blood pressure",
       "diagnosticCode": "7101",
       "disabilityActionType": "DECREASE",
       "disabilityClassificationCode": "3460",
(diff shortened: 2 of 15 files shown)


service/db/src/main/java/gov/va/vro/service/db/SaveToDbServiceImpl.java
```diff
@@ -152,8 +156,8 @@ public void insertOrUpdateExamOrderingStatus(ExamOrder examOrder) {
             .findByCollectionId(examOrder.getCollectionId())
             .orElseGet(() -> createExamOrder(examOrder));
     if (null != examOrderEntity) {
-      examOrderEntity.setCollectionId(examOrder.getCollectionId());
```
hayleymdawson, line 155: Since we just looked up the examOrderEntity by this key, and keys should not change, i removed this. It technically would be equal and not corrupt data but I felt it better to ditch it.

app/src/end2endTest/java/gov/va/vro/end2end/VroV2Tests.java
```diff
@@ -218,7 +218,7 @@ private String getOrderingStatusValidRequest() {
             "collectionStatus",
             "DRAFT",
             "examOrderDateTime",
-            "2022-12-08T17:45:61Z",
+            "2022-12-08T17:45:59Z",
```
hayleymdawson, line 221: ISO 1806 dates don't permit 61 seconds (no date standard does)
This was never valid we just didn't use it. It's now valid.

app/src/end2endTest/resources/test-mas/claim-350-7101-outofscope.json
```diff
@@ -16,7 +16,7 @@
     "claimSubmissionDateTime": "2018-11-04T17:45:59Z",
     "claimSubmissionSource": "VA.GOV",
     "conditions": {
-      "name": "string",
+      "name": "blood pressure",
```
hayleymdawson, line 19: Made this *anything* but the java class name to ensure that it really does populate the database as expected.

yoomlam (approved):
Code looks good! Did not test.

alan-amida (approved):

PR #1060, opened 2023-02-07
Database changes and Java logic

ianwolson:
## What was the problem?

Associated tickets or Slack threads:

- [MCP-2303](https://amida.atlassian.net/browse/MCP-2303)

## How does this fix it?[^1]
Populates the veteran flash ID table

## How to test this PR
- Run the automated claim endpoint using a valid integer represented as a string for veteranFlashIds.
- Check the database and verify that flash ids are populated in the flash id table and that the veteranIcn matches the request.

[^1]: [Pull-Requests guidelines](https://github.com/department-of-veterans-affairs/abd-vro/wiki/Pull-Requests). If PR is significant, update [Current Software State](https://github.com/department-of-veterans-affairs/abd-vro/wiki/Current-Software-State) wiki page.
[^secrel]: To check if a PR will succeed in the SecRel workflow, [test PRs in the SecRel pipeline](https://github.com/department-of-veterans-affairs/abd-vro-internal/wiki/Secure-Release-process#to-test-prs-in-the-secrel-pipeline).

PR #1046, opened 2023-02-03
Add DB schema enhancement changes

alan-amida:
DB Schema changes based on Dimitri's [DB Enhancement Proposa](https://github.com/department-of-veterans-affairs/abd-vro/wiki/VRO-Database#proposed-redesign)l and Yoom's [feedback](https://github.com/department-of-veterans-affairs/abd-vro/wiki/VRO-Database#feedback-on-proposal)

## Current order of operations (subject to change based on review/feedback):
1. Create `veteran_flash_id` table to stored a veteran's associated `flashId`s
  a. currently only processing one flashId for Agent Orange exposure
2. Create `exam_order` table to store the status of `/v2/examOrderingStatus` requests
3. Update `audit_event` table's `details` column type from VARCHAR to JSONB
4. Update our `claim` table with columned fields stored as JOSNB in `audit_event.details` (no replacement):
  a. add `vbms_id` field
  b. add `off_ramp_reason` field
  c. add `presumptive_flag` field
  d. add `disability_action_type` field
  e. add `in_scope` field
  f. add `submission_source` field
  g. add `submission_date` field
5. Update our `veteran` table to include a `icn_timestamp` field
6. Update our `evidence_summary_document` table to include a `location` field

## In-progress operations:
- Implementing Java Entities and necessary other Java code complete the "from Data hitting our API to our DB" path

I want to make sure the Model Schema changes are good with everyone before we proceed with spending any time writing the associated Java.

## Open Questions:
- `claim.id` is our internal DB primary key for the claim, but `claim.claim_submission_id` is an external number provided by [some external caller of our API]?
- Is it better to think of `claim` table as "current claim state" vs `claim_submission` table as the new "claim audit event history"?
  - This might mean we can continue to store `details` as recorded history in `claim_submission`, but continue to maintain current state in `claim` table. Ultimately must consult with Yoom, but might be better to err on the side of redundancy than otherwise.
  - It feels like the root of this ambiguity is because there may be a misunderstanding on (or conflation with) claim history tracking vs current state
- Do the data types look okay?
- Where do we want any more indexing to happen?
- What else am I missing?

## Some clarifications from Dimitri:
- Leave `audit_event` as-is, change `details` to JSONB, just add fields to `claim` table. We can leave out:
  - ~~requestType: CLAIM_SUBMISSION, AUTOMATED_CLAIM, OTHER~~
  - ~~status: PROCESSED, REJECTED, ERROR (or something like this)~~
[...]

PR #1112, opened 2023-02-15
Feature/database improvements clean

alan-amida:
@yoomlam This is a mirror of PR #1046 , which I am closing due to the comments littering it everywhere -- I reiterate, I have cloned this PR to make this easier to read. The previous PR is full of back-and-forth comment noise... but they are still there for anyone who wants to review all of that backstory.

This is now ready for review. Yoom, please review this for ensuring that this addresses your concerns and feedback (see the below Overall plan for DB changes section as a reference guide).

I will be adding @au2 @rajesh-Krishnasamy @dimitri-amida @wlin2019  to review our work for sanity and to ensure we are not breaking any expected INPUT or OUTPUT

THIS PR IS HUGE. We apologize for this. This turned out to be a much bigger PR than we intended, but by the time work had begun, we were somewhat locked into a development cycle until we could finally get feedback from Yoom. We ended up creating smaller secondary PRs (#1060 ) that were merged into the primary PR (#1046), which is now this.

### Note

- Most everything here is adding DB tables and columns which are geared toward capturing more Claim, ClaimSubmission, Veteran, etc information
- No Camel Routes were touched
- This SHOULD NOT affect any existing API input or output, and we are depending on everyone reviewing this PR to verify this -- i.e. this SHOULD NOT affect any ongoing testing efforts, but you know how these things can go

This PR is already huge, and this work will be done in a much smaller PR (the work is currently residing in a branch pushed up by Hayley)

# Overall plan for DB changes
### Schema
- [x] KEEP THE `audit_event` table
  - This is strictly for internal system auditing
  - [x] update the `details` column to JSONB for easier querying if needed
- [x] Create `claim_submission` table:
  - This represents claim submission request states, understanding that a claim might be submitted more than once (failed, retried, etc)
  - [x] add `reference_id` field (replaces `collection_id` currently in `claim` table)
  - [x] add `id_type` field
  - [x] add `incoming_status` field
  - ~~add `submission_source` field~~ (audit_event concern)
  - ~~add `submission_date` field~~ (audit_event concern)
  - [x] Add `rfd_flag` to `claim` table
  - [x] add `off_ramp_reason` field
- [x] Create `veteran_flash_id` table to store a veteran's associated `flashId`s
  - currently only processing one `flashId` for Agent Orange exposure
  - [x] Map Foreign Key to `veteran` Primary Key (id)
  - [x] Add `status` field
- [x] Create `exam_order` table to store the status of `/v2/examOrderingStatus` requests
  - [x] Map Foreign Key to `claim_submission` Primary Key (id) (`findByReferenceId` via `collectionId` from JSON payload)
  - [x] Add `status` field
- [x] Update our `claim` table:
  - [x] add `vbms_id` field
  - ~~add `off_ramp_reason` field~~
  - [x] add `presumptive_flag` field
  - [x] add `disability_action_type` field
  - [x] add `in_scope` field
  - ~~add `submission_source` field~~ (audit_event concern)
  - ~~add `submission_date` field~~ (audit_event concern)
- [x] Update our `veteran` table to include a `icn_timestamp` field
- [x] Update our `evidence_summary_document` table to include a `location` field
- [x] Remove `collection_id` from `claim` table

### Java Code
- [x] Update logic to accommodate JSONB
  - This is strictly for internal system auditing
- [x] Create `ClaimSubmissionEntity`
  - [x] Add logic to create `ClaimSubmission` record
  - [x] Update `collectionId` field logic going to `ClaimEntity` with `referenceId` field going to `ClaimSubmissionEntity`
  - [x] Update logic for `idType` field going to `ClaimEntity` to `ClaimSubmissionEntity`
  - [x] Update logic for `incomingStatus` field going to `ClaimEntity` to `ClaimSubmissionEntity`
  - [x] Add logic to create `rfd_flag` to `ClaimSubmissionEntity`
- [x] Add logic to capture `veteranFlashIds` to go into `VeteranFlashIdEntity`
- [x] Create `ExamOrderEntity`
  - [x] Add logic to create `ExamOrder` record
  - [x] `findByReferenceId` via `collectionId` from JSON payload
  - [x] `collectionStatus` will be `status`
- [x] Update `ClaimEntity`:
  - [x] Add logic to capture `vbmsId` value
  - [x] Add logic to capture `off_ramp_reason` field
  - [x] Add logic to capture `presumptive_flag` field
  - [x] Add logic to capture `disability_action_type` field
  - [x] Add logic to capture `in_scope` field
- [x] Add logic to capture `icnTimestamp` to go into `veteran` table

### Moving to the next PR
- Add logic to capture `location` info for `evidence_summary_document` table


app/src/test/java/gov/va/vro/controller/VroControllerTest.java
```diff
@@ -122,9 +120,8 @@ void postFullHealthAssessment() throws Exception {
     assertEquals(request.getDiagnosticCode(), response2.getDiagnosticCode());
     assertEquals(request.getVeteranIcn(), response2.getVeteranIcn());
 
-    Optional<ClaimEntity> claimEntityOptional =
-        claimRepository.findByClaimSubmissionIdAndIdType("1234", "va.gov-Form526Submission");
-    assertTrue(claimEntityOptional.isPresent());
+    var claim = claimRepository.findByVbmsId(request.getClaimSubmissionId());
```
yoomlam, line 123: For MAS requests, the claimSubmissionId (referenceId) is the collectionId.
For claims, the vbmsID is the benefitsClaimId, as provided by MAS.
Thus, the vbmsID is not comparable to the claimSubmissionId.

See https://dsva.slack.com/archives/C01Q7979Z7D/p1674145386628059?thread_ts=1674144499.310049&cid=C01Q7979Z7D and https://dsva.slack.com/archives/C01Q7979Z7D/p1674237329439659?thread_ts=1674144499.310049&cid=C01Q7979Z7D

db-init/src/main/resources/database/migrations/V4.3__Update_Claim_Table.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE claim 
+ADD COLUMN vbms_id VARCHAR,
+ADD COLUMN off_ramp_reason VARCHAR, 
+ADD COLUMN presumptive_flag BOOLEAN, 
+ADD COLUMN disability_action_type VARCHAR,
+ADD COLUMN in_scope BOOLEAN,
```
yoomlam, line 6: Should be in claim_submission. See https://dsva.slack.com/archives/C01Q7979Z7D/p1675959892218709?thread_ts=1675912916.651659&cid=C01Q7979Z7D

db-init/src/main/resources/database/migrations/V4.3__Update_Claim_Table.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE claim 
+ADD COLUMN vbms_id VARCHAR,
+ADD COLUMN off_ramp_reason VARCHAR,
```
yoomlam, line 3: Should be in claim_submission. See https://dsva.slack.com/archives/C01Q7979Z7D/p1675959892218709?thread_ts=1675912916.651659&cid=C01Q7979Z7D

service/provider/src/main/java/gov/va/vro/service/provider/bip/service/BipClaimService.java
```diff
@@ -146,6 +149,7 @@ public MasProcessingObject markAsRfd(MasProcessingObject payload) {
 
     try {
       bipApiService.updateClaimStatus(claimId, ClaimStatus.RFD);
+      saveToDbService.updateRfdFlag(String.valueOf(claimId), true);
```
wlin2019, line 152: I see there are two places to update the claim status to ClaimStatus.RFD. Is it intended to save the info only in this place (not for completeProcessing)?

db-init/src/main/resources/database/migrations/V4.3__Update_Claim_Table.sql
hayleymdawson: Per the referenced slack discussion to clarify this comment
presumptive_flag, and disability_action_type fields should exist in the claim table -- these values shouldn't change if the same claim was resubmitted. 

values for off_ramp_reason and in_scope need to be moved.

wlin2019 (commented):
I am not familiar with the database in VRO. I don't think I have enough knowledge to approve or disapprove this PR. I left some comment.

hayleymdawson:
Review is pausing as of Feb 15th 12:30 with a video call with Yoom, Rajesh, Ian, Myself, and Alan.
The following will be happening immediately regarding "claimSubmissionId":

The Mas Automated Claim process which goes through a series of variable renames will refer to benefitClaimId as benefitClaimId up until we refer to it as vbmsId which is how it is stored in the claim postgres table. It will *not* be referred to as "claimSubmissionId" as it is currently because that is not what it is.

THe other entry points from V1 that refer to claimSubmissionId (equivalent to reference_id), will need to look up the claim_submission table first then use the reference to find the actual claim. They will not equate claimSubmissionId and vbmsId (nor our internal Ids that are not exposed)

A future PR will clean up claimSubmissionId and referenceId naming, but this PR will make sure that there is only one use of claimSubmissionId and it means reference id.

hayleymdawson:
Work that was holding up this review is now finished and this is ready for review.

service/spi/src/main/java/gov/va/vro/service/spi/model/Claim.java
```diff
@@ -26,10 +27,19 @@ public class Claim {
...
+  // Either benefitClaimId or CollectionId must be filled out
+  // CollectionId is equivalent to reference_id on the claim submission table
+  private String benefitClaimId;
 
   private String collectionId;
 
+  // Backwards compatibility with v1 routes. On the way in mappers put claimsubmission id into
+  // collection id
+  // Both ways tie to the reference_id on claim submission table.
+  public String getClaimSubmissionId() {
```
hayleymdawson, line 39: Ideally there would be a v1 and a v2 claim entity object. However due to time constraints there is no possible way to do that.

app/src/test/java/gov/va/vro/TestDataSupplier.java
```diff
@@ -13,26 +17,31 @@ public class TestDataSupplier {
...
+  public static ClaimEntity createClaim(
+      String benefitClaimId, VeteranEntity veteran, String referenceId) {
     ClaimEntity claim = new ClaimEntity();
-    claim.setClaimSubmissionId(id);
-    claim.setIdType(idType);
+    claim.setVbmsId(benefitClaimId);
     claim.setVeteran(veteran);
+    ClaimSubmissionEntity claimSubmission = new ClaimSubmissionEntity();
+    claimSubmission.setReferenceId(referenceId);
+    claimSubmission.setIdType(Claim.DEFAULT_ID_TYPE);
```
yoomlam, line 43: What is the value of idType when MAS submits a request to the `v2/automationClaim` endpoint? I wasn't able to find it in the code. The value should not be `Claim.DEFAULT_ID_TYPE`, which is for `v1` endpoints.

db-init/src/main/resources/database/migrations/V4.1__Create_ExamOrder_Table.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE IF NOT EXISTS exam_order (
+    id uuid DEFAULT uuid_generate_v4 () NOT NULL,
+    collection_id VARCHAR,
+    status VARCHAR,
```
yoomlam, line 4: Expecting `status_at timestamp` to be added in another PR to satisfy from the PR description:
> 3. For the /examOrderingStatus endpoint, ~how is the examOrderDateTime field of the payload stored in the DB? If it's not stored, please store it as ordered_at only if the exam was successfully ordered.~ (regardless of status) store the status string and a status_at timestamp.

db-init/src/main/resources/database/migrations/V4.3__Update_Claim_Table.sql
yoomlam: Done in V4.7 and V4.8 migrations.

db-init/src/main/resources/database/migrations/V4.3__Update_Claim_Table.sql
yoomlam: in_scope moved in V4.8 and V4.11.

db-init/src/main/resources/database/migrations/V4.7__Add_Claim_FK_On_Claim_Submission.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE claim_submission
+    ADD COLUMN IF NOT EXISTS claim_id UUID DEFAULT uuid_generate_v4 () NOT NULL;
```
yoomlam, line 2: If it's null, it shouldn't generate a new random UUID. It should use the UUID from the claim table.

db-init/src/main/resources/database/migrations/V4.9__Add_Sufficient_Evidence_Flag.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE assessment_result ADD COLUMN sufficient_evidence_flag BOOLEAN;
```
yoomlam, line 1: I want to confirm that `sufficient_evidence_flag` can be null, so that we can treat values True, False, and Null as distinct values like so:
image

persistence/model/src/main/java/gov/va/vro/persistence/model/AssessmentResultEntity.java
```diff
@@ -21,6 +21,8 @@ public class AssessmentResultEntity extends BaseEntity {
 
   @ManyToOne private ContentionEntity contention;
 
+  private Boolean sufficientEvidenceFlag;
```
yoomlam, line 24: Please add a unit test to allow `sufficientEvidenceFlag` to be Null (when its null in the DB), as well as False. The intention of the test is to prevent this field from being changed to a primitive `boolean`.

persistence/model/src/main/java/gov/va/vro/persistence/model/ClaimSubmissionEntity.java
```diff
@@ -0,0 +1,32 @@
...
+import javax.persistence.Table;
+
+@Entity
+@Getter
+@Setter
+@Table(name = "claim_submission")
+public class ClaimSubmissionEntity extends BaseEntity {
+  @ManyToOne private ClaimEntity claim;
+
+  private String referenceId;
```
yoomlam, line 18: Yesss!

persistence/model/src/main/java/gov/va/vro/persistence/model/VeteranFlashIdEntity.java
```diff
@@ -0,0 +1,31 @@
...
+
+  @Id
+  @GeneratedValue(generator = "uuid")
+  @GenericGenerator(name = "uuid", strategy = "uuid2")
+  @Column(columnDefinition = "BINARY(16)")
+  private UUID id;
+
+  @ManyToOne private VeteranEntity veteran;
+
+  @NotNull private Integer flashId;
```
yoomlam, line 30: I'm not familiar with this in the code, but there was talk of a String version of the flash. Did you come across a mapping of these `flashId` Integers to flash strings?

persistence/model/src/main/java/gov/va/vro/persistence/repository/ClaimSubmissionRepository.java
```diff
@@ -0,0 +1,22 @@
...
+import java.util.UUID;
+
+@Repository
+public interface ClaimSubmissionRepository extends JpaRepository<ClaimSubmissionEntity, UUID> {
+
+  List<ClaimSubmissionEntity> findByReferenceIdAndIdType(String referenceId, String idType);
+
+  // For the few cases where we really don't know an idType, this is equivalent to the prior Claim
+  // find byReferenceId not specifying a type.
+  Optional<ClaimSubmissionEntity> findFirstByReferenceIdOrderByCreatedAtDesc(String referenceId);
```
yoomlam, line 18: Reminder to self: what value of idType is being used when only the referenceId is provided?

yoomlam (commented):
Submitting first batch of comments.

service/db/src/main/java/gov/va/vro/service/db/SaveToDbServiceImpl.java
```diff
@@ -87,14 +139,29 @@ private void insertAssessmentResult(
     claimRepository.save(claimEntity);
   }
 
+  @Override
+  public void setOffRampReason(Claim claimWithOffRamp) {
+    Optional<ClaimSubmissionEntity> claimSubmission =
+        claimSubmissionRepository.findFirstByReferenceIdAndIdTypeOrderByCreatedAtDesc(
+            String.valueOf(claimWithOffRamp.getCollectionId()), Claim.DEFAULT_ID_TYPE);
```
yoomlam, line 146: This would be a bug for v2 endpoints.
```suggestion
            String.valueOf(claimWithOffRamp.getCollectionId()), claimWithOffRamp.getIdType());
```

service/db/src/main/java/gov/va/vro/service/db/SaveToDbServiceImpl.java
```diff
@@ -87,14 +139,29 @@ private void insertAssessmentResult(
...
   @Override
   public void insertEvidenceSummaryDocument(GeneratePdfPayload request, String documentName) {
-    ClaimEntity claim =
-        claimRepository.findByClaimSubmissionId(request.getClaimSubmissionId()).orElse(null);
-    if (claim == null) {
-      log.warn("Could not find claim by claimSubmissionId, exiting.");
+    // ClaimSubmissionId in v1 calls is equal to ReferenceId on ClaimSubmission in VRO
+    Optional<ClaimSubmissionEntity> claimSubmission =
+        claimSubmissionRepository.findFirstByReferenceIdAndIdTypeOrderByCreatedAtDesc(
+            request.getClaimSubmissionId(), Claim.DEFAULT_ID_TYPE);
```
yoomlam, line 157: This would be a bug for v2 endpoints.
```suggestion
            request.getClaimSubmissionId(), claim. getIdType());
```

service/db/src/main/java/gov/va/vro/service/db/SaveToDbServiceImpl.java
```diff
@@ -87,14 +139,29 @@ private void insertAssessmentResult(
...
+    claimSubmissionRepository.save(claimSubmissionEntity);
+  }
+
   @Override
   public void insertEvidenceSummaryDocument(GeneratePdfPayload request, String documentName) {
-    ClaimEntity claim =
-        claimRepository.findByClaimSubmissionId(request.getClaimSubmissionId()).orElse(null);
-    if (claim == null) {
-      log.warn("Could not find claim by claimSubmissionId, exiting.");
+    // ClaimSubmissionId in v1 calls is equal to ReferenceId on ClaimSubmission in VRO
```
yoomlam, line 154:
```suggestion
    // For v1 endpoints, ClaimSubmissionId = ClaimSubmission.referenceId
```

service/db/src/main/java/gov/va/vro/service/db/SaveToDbServiceImpl.java
```diff
@@ -105,6 +172,95 @@ public void insertEvidenceSummaryDocument(GeneratePdfPayload request, String doc
...
+  @Override
+  @Transactional
+  public void insertOrUpdateExamOrderingStatus(ExamOrder examOrder) {
+    ExamOrderEntity examOrderEntity =
+        examOrderRepository
+            .findByCollectionId(examOrder.getCollectionId())
+            .orElseGet(() -> createExamOrder(examOrder));
+    if (null != examOrderEntity) {
+      examOrderEntity.setCollectionId(examOrder.getCollectionId());
+      examOrderEntity.setStatus(examOrder.getStatus());
```
yoomlam, line 184: In subsequent PR, remember to update `status_at` with the payload's `examOrderDateTime`.

service/db/src/main/java/gov/va/vro/service/db/SaveToDbServiceImpl.java
```diff
@@ -105,6 +172,95 @@ public void insertEvidenceSummaryDocument(GeneratePdfPayload request, String doc
...
+    claimRepository.save(claimEntity);
+  }
+
+  @Override
+  public void updateSufficientEvidenceFlag(
+      String claimSubmissionId, Boolean flag, String diagnosticCode) {
+    // ClaimSubmissionId in v1 = ReferenceId on ClaimSubmission in VRO
+    Optional<ClaimSubmissionEntity> claimSubmission =
+        claimSubmissionRepository.findFirstByReferenceIdAndIdTypeOrderByCreatedAtDesc(
+            claimSubmissionId, Claim.DEFAULT_ID_TYPE);
```
yoomlam, line 224: Somehow, you'll need to get the claim or `idType`.
```suggestion
            claimSubmissionId, claim.getIdType());
```

service/db/src/main/java/gov/va/vro/service/db/SaveToDbServiceImpl.java
```diff
@@ -105,6 +172,95 @@ public void insertEvidenceSummaryDocument(GeneratePdfPayload request, String doc
...
+    if (claimSubmission.isEmpty()) {
+      log.warn(
+          "Claim Submission not found for claim submission id = {} and id type = {} in update sufficient evidence flag",
+          claimSubmissionId,
+          Claim.DEFAULT_ID_TYPE);
```
yoomlam, line 229:
```suggestion
          claim.getIdType());
```

service/db/src/main/java/gov/va/vro/service/db/SaveToDbServiceImpl.java
```diff
@@ -105,6 +172,95 @@ public void insertEvidenceSummaryDocument(GeneratePdfPayload request, String doc
...
+      log.warn("Could not find contention with given diagnostic code.");
+      return;
+    }
+    Optional<AssessmentResultEntity> result =
+        assessmentResultRepository.findFirstByContentionIdOrderByCreatedAtDesc(contention.getId());
+    if (result.isEmpty()) {
+      log.warn("Could not match assessment result to this contention id.");
+      return;
+    }
+    if (flag == null) {
```
yoomlam, line 244: Good -- the flag can be `null`.

service/spi/src/main/java/gov/va/vro/service/spi/model/Claim.java
hayleymdawson: Oh, sorry. I apparently misworded that comment. I meant, a v1 and v2 Claim *spi model* that we pass around the code. You are absolutely correct we definitely do not want 2 claim entities or table.

service/db/src/main/java/gov/va/vro/service/db/SaveToDbServiceImpl.java
```diff
@@ -159,6 +315,24 @@ private ClaimEntity createClaim(Claim claim, VeteranEntity veteranEntity) {
     return claimRepository.save(claimEntity);
   }
 
+  private ExamOrderEntity createExamOrder(ExamOrder examOrder) {
+    // Currently ExamOrders only come from MAS
+    Optional<ClaimSubmissionEntity> claimSubmission =
+        claimSubmissionRepository.findFirstByReferenceIdAndIdTypeOrderByCreatedAtDesc(
+            examOrder.getCollectionId(), Claim.DEFAULT_ID_TYPE);
```
yoomlam, line 322: If we assume this method is only called for v2 endpoints (which is a valid assumption since v1 doesn't involve exam ordering), then
```suggestion
            examOrder.getCollectionId(), Claim.V2_ID_TYPE);
```

service/spi/src/main/java/gov/va/vro/service/spi/model/Claim.java
```diff
@@ -26,10 +27,19 @@ public class Claim {
 
   private UUID recordId;
 
-  @NotNull private String claimSubmissionId;
+  // Either benefitClaimId or CollectionId must be filled out
```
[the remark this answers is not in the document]
hayleymdawson: I'll add more commentary here, but, yes in MAS we get benefitClaimId and in v1 we get claimSubmissionId

There is code that maps v1 claimSubmissionId to collectionId in this file that we pass around. Both tie to the reference_id field. (that's part of why ideally the objects wouldn't be the same but that turned into a serious set of changes in the camel routes)

What the comment means is that we should either get a benefitClaimId from MAS, or v1 needs to give us a claimSubmissionId that we set as collectionId here.

service/spi/src/main/java/gov/va/vro/service/spi/model/Claim.java
```diff
@@ -26,10 +27,19 @@ public class Claim {
...
+  // CollectionId is equivalent to reference_id on the claim submission table
+  private String benefitClaimId;
 
   private String collectionId;
 
+  // Backwards compatibility with v1 routes. On the way in mappers put claimsubmission id into
+  // collection id
```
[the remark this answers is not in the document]
hayleymdawson: You are absolutely correct on what that comment should read and the code means. I will make the correction.

persistence/model/src/main/java/gov/va/vro/persistence/model/VeteranFlashIdEntity.java
hayleymdawson: I did not come across that mapping. @alan-amida and maybe @ianwolson might know the history on this one.

yoomlam (commented):
I suggest first renaming `Claim.DEFAULT_ID_TYPE` to `Claim.V1_ID_TYPE`, add
`Claim.V2_ID_TYPE = "MAS"`, and addressing my comments related to finding a ClaimSubmissionEntity.

If addressing some of the comments become too challenging, we can assume that v1 endpoints will not be called. This assumption can be removed as a future tech debt ticket.

persistence/model/src/main/java/gov/va/vro/persistence/repository/ClaimSubmissionRepository.java
hayleymdawson: Re: the comment above for v2 types and this it does look like the routes aren't setting the type correctly, and I will make them set the type to "MAS". 

As far as removing this method, i left it in place for claim metrics since that is a v1 api and it can only specify the reference id but not the id type unless I change the api contract. That is the only place it's used. If that's not what we want, I can certainly do any of the following:
1. change the claim metrics API point to ask for the id type
2. change the claim metrics API point to only search one type or the other (presumably the v1 type). 
3. Leave it, with the slight risk of this method's existence, but note that it is only called and should only ever be called by the metrics api.

service/db/src/main/java/gov/va/vro/service/db/SaveToDbServiceImpl.java
```diff
@@ -167,7 +341,11 @@ private ContentionEntity createContention(ClaimEntity claim, String diagnosticCo
   }
 
   private VeteranEntity findOrCreateVeteran(String veteranIcn) {
-    return veteranRepository.findByIcn(veteranIcn).orElseGet(() -> createVeteran(veteranIcn));
+    VeteranEntity veteranEntity =
+        veteranRepository.findByIcn(veteranIcn).orElseGet(() -> createVeteran(veteranIcn));
+    Date date = new Date();
+    veteranEntity.setIcnTimestamp(date);
```
yoomlam, line 347: Yes!

service/db/src/main/java/gov/va/vro/service/db/SaveToDbServiceImpl.java
```diff
@@ -28,21 +39,60 @@ public class SaveToDbServiceImpl implements SaveToDbService {
...
+    claimEntity.addClaimSubmission(claimSubmissionEntity);
+    claimSubmissionRepository.save(claimSubmissionEntity);
+    claimRepository.save(claimEntity);
     claim.setRecordId(claimEntity.getId());
     return claim;
   }
 
+  private ClaimSubmissionEntity createClaimSubmission(Claim claim) {
+    ClaimSubmissionEntity claimSubmission = new ClaimSubmissionEntity();
+    claimSubmission.setReferenceId(claim.getCollectionId());
```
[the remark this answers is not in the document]
hayleymdawson: So it is. The Claim object here is the SPI model, not the Claim Entity nor the raw payload. On the way in, claimSubmissionId becomes collectionId which gets set to reference Id.

persistence/model/src/main/java/gov/va/vro/persistence/repository/ClaimSubmissionRepository.java
yoomlam: I prefer bullet 1 if it's not too burdensome. The Claim Metrics API is actually relevant to both v1 and v2, but it's not a critical API and only used for our reporting purposes.
Alternatively, you can do bullet 2 using `idType="MAS"` (for v2 endpoints) -- I know, a bit confusing (sorry).

service/spi/src/main/java/gov/va/vro/service/spi/model/Claim.java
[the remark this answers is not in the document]
yoomlam: Ah! Got it. Suggested change:
```suggestion
  // we should either get a benefitClaimId from MAS for v2 endpoints, or v1 endpoints provide a claimSubmissionId that we use to set the collectionId
```

persistence/model/src/main/java/gov/va/vro/persistence/model/VeteranFlashIdEntity.java
yoomlam: Here's one Slack thread https://dsva.slack.com/archives/C0309PZT20Z/p1675966464451669?thread_ts=1675710347.845039&cid=C0309PZT20Z:
> Could you pls share the "short name" of the (Agent Orange Exposure Verified flash (ID: 266)) in the notification to VRO .i.e automatedClaim API request

Response:
> this is the flash name we will send. "Agent Orange Exposure Verified". We are not aware of different types of names (short/long etc.) for flashes.

This may be irrelevant but I wan't to bring it up, just in case.

yoomlam: Actually now that I'm reading [this Slack thread](https://dsva.slack.com/archives/C0309PZT20Z/p1675271527188669?thread_ts=1668694256.106189&cid=C0309PZT20Z):
> We do not have any flash ids. We can send the short names of the applicable flashes as a list of strilngs.

and looking at the request object schema:
image

I think we should change the `flashId` into a String -- this is the safest bet.

service/provider/src/main/java/gov/va/vro/service/provider/services/EvidenceSummaryDocumentProcessor.java
```diff
@@ -16,11 +18,15 @@ public class EvidenceSummaryDocumentProcessor implements Processor {
 
   @Override
   public void process(Exchange exchange) {
+    String body = exchange.getIn().getBody(String.class);
+    log.info(">>>>> exchange > body: " + body);
```
yoomlam, line 22: Let's comment these out in case there's any PII in the `body`.
```suggestion
    //log.info(">>>>> exchange > body: " + body);
```

service/provider/src/main/java/gov/va/vro/service/provider/services/EvidenceSummaryDocumentProcessor.java
```diff
@@ -16,11 +18,15 @@ public class EvidenceSummaryDocumentProcessor implements Processor {
...
     GeneratePdfPayload payload = exchange.getIn().getBody(GeneratePdfPayload.class);
+    log.info(">>>>> exchange > payload~1: " + payload);
```
yoomlam, line 24:
```suggestion
    // log.info(">>>>> exchange > payload~1: " + payload);
```

service/provider/src/main/java/gov/va/vro/service/provider/services/EvidenceSummaryDocumentProcessor.java
```diff
@@ -16,11 +18,15 @@ public class EvidenceSummaryDocumentProcessor implements Processor {
...
     if (payload == null) {
       log.warn("Payload is empty, returning...");
       return;
     }
+    log.info(">>>>> exchange > payload~2: " + Objects.toString(payload));
```
yoomlam, line 29:
```suggestion
    // log.info(">>>>> exchange > payload~2: " + Objects.toString(payload));
```

shared/domain/src/main/java/gov/va/vro/model/mas/MasAutomatedClaimPayload.java
```diff
@@ -90,7 +90,7 @@ public Boolean isPresumptive() {
   }
 
   @JsonIgnore
-  public Integer getClaimId() {
+  public Integer getBenefitClaimId() {
     return claimDetail == null ? null : Integer.parseInt(claimDetail.getBenefitClaimId());
   }
```
yoomlam, line 95: I don't think the types match. `vbmsId` (aka BenefitClaimId) is a String. We shouldn't be coercing it into an Integer. Any uses of this method is likely to produce a bug.

persistence/model/src/main/java/gov/va/vro/persistence/model/VeteranFlashIdEntity.java
alan-amida: Agreed, we'll change it to String.

service/provider/src/main/java/gov/va/vro/service/provider/services/EvidenceSummaryDocumentProcessor.java
hayleymdawson: Agreed, and done.

shared/domain/src/main/java/gov/va/vro/model/mas/MasAutomatedClaimPayload.java
hayleymdawson: Corrected, and good catch.

app/src/test/java/gov/va/vro/controller/VroControllerTest.java
hayleymdawson: Resolved by larger code commits later in the code.

db-init/src/main/resources/database/migrations/V4.9__Add_Sufficient_Evidence_Flag.sql
hayleymdawson: As per below comments, it absolutely can be null. Resolving this comment. Will resolve other comment once unit test is in.

alan-amida:
@yoomlam 

As I'm sure you agree, we are at a point in development that we must focus on what is paramount for iMVP.

My view is for @hayleymdawson and @ianwolson to focus entirely on schematic feedback first, and any other findings about data flowing through the stack can wait provided it 1) does not break I/O and 2) does not break putting the expected data where we want it in the DB

Time is running short and we simply cannot fix all of VRO in this PR. We will address any important but not-direly-mission-critical findings after the code-freeze (which, as I understand it, will allow us to resolve bugs and critical issues in the following week).

I will have Hayley summarize what she has addressed and what she feels would be a risk of running over time if she were to tackle it here.

In the meantime, the work continues.

yoomlam:
@alan-amida Sounds good!

hayleymdawson:
@yoomlam @alan-amida @rajesh-Krishnasamy 

What has been done:
Claim Type is now different between v1 and v2. 
Default is renamed to V1, V2 is declared in mas processing code with a string that was requested by Rajesh in preference over just MAS
Flash id is now string
Swagger docs are updated for clarity 
Comments updated for clarity
Claim metric get claimInfo now has an optional query parameter. It accepts v1, or v2 to specify the kind of claim. It defaults to v2. Other methods use the claimRepo entry point so they can see everything without a parameter to specify, it is only for a specific claim id we need this. The info object returns the type just in case there's ever a collision. (this can be altered easily but seemed reasonable to do). 
The integer coercion of benefitClaimId is removed

If we would prefer I can make the claimMetric API take the same string we store as idType for it's query parameter instead for input/output symmetry but I erred on the side of easy use for a first go at it.

What has not been done:
1) items intended for next PR by prior agreement
2) ABDEvidenceWithSummary *does* coerce the sufficient flag to a primitive. However, changing it here caused v1 breaks in areas of the code that I am not familiar with and due to time constraints have not fixed here. @ianwolson is investigating this for a subsequent PR. 
3) There is an additional method that coerces benefitClaimId to a long. this interacts with BIP service code in an area that I have not yet looked at to be familiar with the flow of. It will also have to be fixed in a subsequent PR for time reasons.

service/provider/src/main/java/gov/va/vro/service/provider/mas/MasProcessingObject.java
```diff
@@ -18,16 +20,20 @@ public int getCollectionId() {
...
 
-  public String getClaimId() {
+  public String getBenefitClaimId() {
     return claimPayload.getClaimDetail().getBenefitClaimId();
   }
 
-  public long getClaimIdAsLong() {
-    String claimIdString = getClaimId();
+  public long getBenefitClaimIdAsLong() {
+    String claimIdString = getBenefitClaimId();
```
yoomlam, line 28:
```suggestion
    // benefitClaimId (aka vbmsId)
    // TODO: verify that vbmsId is always a long or check the API spec where this long is submitted
    String claimIdString = getBenefitClaimId();
```

service/provider/src/main/java/gov/va/vro/service/provider/camel/MasIntegrationProcessors.java
```diff
@@ -86,7 +88,8 @@ private static GeneratePdfPayload getGeneratePdfPayload(MasProcessingObject tran
     MasAutomatedClaimPayload claimPayload = transferObject.getClaimPayload();
     GeneratePdfPayload generatePdfPayload = new GeneratePdfPayload();
     generatePdfPayload.setEvidence(transferObject.getEvidence());
-    generatePdfPayload.setClaimSubmissionId(claimPayload.getClaimDetail().getBenefitClaimId());
+    generatePdfPayload.setClaimSubmissionId(String.valueOf(claimPayload.getCollectionId()));
```
yoomlam, line 91: Should be a String:
```suggestion
    generatePdfPayload.setClaimSubmissionId(claimPayload.getCollectionId());
```

shared/domain/src/main/java/gov/va/vro/model/mas/MasAutomatedClaimPayload.java
```diff
@@ -46,6 +47,8 @@ public class MasAutomatedClaimPayload implements Auditable {
 
   @NotNull @Valid private ClaimDetail claimDetail;
 
+  @Builder.Default @NotNull private String idType = CLAIM_V2_ID_TYPE;
+
```
yoomlam, line 51: This comment is for line 44 above. It should be:
```
   private String collectionId;
```
See my comment for [MasIntegrationProcessors.java](https://github.com/department-of-veterans-affairs/abd-vro/pull/1112/files#diff-af0561bfcedd9b27f11cfb67d867cf4ffc9f7a4685f4848fe2483986f3d31d89) line 91.

service/db/src/test/java/gov/va/vro/service/db/util/ClaimMetricsTestCase.java
```diff
@@ -58,14 +60,28 @@ private GeneratePdfPayload getPdfPayload(AbdEvidence evidence) {
...
+    claim.setCollectionId(claimSubmissionId);
     claim.setVeteranIcn(icn);
     claim.setDiagnosticCode("7101");
     service.insertClaim(claim);
 
-    ClaimEntity claimEntity = repo.findByClaimSubmissionId(claimSubmissionId).orElseThrow();
+    ClaimEntity claimEntity;
+
+    // v1 Some calls send us claimSubmissionId which is the same as reference_id on the
+    // claim_submission table (which is collectionId as well)
```
yoomlam, line 73:
```suggestion
    // v1 endpoints provide claimSubmissionId, which maps to reference_id on the
    // claim_submission table (which is used as collectionId)
```

service/db/src/test/java/gov/va/vro/service/db/SaveToDbServiceImplTest.java
```diff
@@ -40,71 +35,146 @@ class SaveToDbServiceImplTest {
...
+    ClaimSubmissionEntity claimSubmission2 = csEntities.iterator().next();
+    assertEquals(claim.getOffRampReason(), claimSubmission2.getOffRampReason());
+  }
+
   @Test
   void persistEvidenceSummaryDocument() throws Exception {
     // Save claim
     Claim claim = new Claim();
-    claim.setClaimSubmissionId("1234");
+    claim.setBenefitClaimId("787878");
```
yoomlam, line 175: For clarity
```suggestion
    vbmsId="787878"
    claim.setBenefitClaimId(vbmsId);
```
and change line 194 below to use the `vbmsId` variable.

service/db/src/test/java/gov/va/vro/service/db/SaveToDbServiceImplTest.java
```diff
@@ -113,11 +183,15 @@ void persistEvidenceSummaryDocument() throws Exception {
...
+    // Payload gives us claimsubmissionId which is the same as the reference_id on the claim
+    // submission table. Type should match the previously submitted claim. This is normally set by
+    // the endpoint processing.
+    input.setIdType(MasAutomatedClaimPayload.CLAIM_V2_ID_TYPE);
     String diagnosis = "Hypertension";
     String documentName = GeneratePdfPayload.createPdfFilename(diagnosis);
     // Save evidence summary document.
     saveToDbService.insertEvidenceSummaryDocument(input, documentName);
-    ClaimEntity result = claimRepository.findByClaimSubmissionId("1234").orElseThrow();
+    ClaimEntity result = claimRepository.findByVbmsId("787878").orElseThrow();
```
yoomlam, line 194:
```suggestion
    ClaimEntity result = claimRepository.findByVbmsId(vbmsId).orElseThrow();
```

service/db/src/test/java/gov/va/vro/service/db/SaveToDbServiceImplTest.java
```diff
@@ -40,71 +35,146 @@ class SaveToDbServiceImplTest {
...
+    assertEquals(assessmentResult.getSufficientEvidenceFlag(), false);
 
     long c = assessmentResultRepository.count();
     assertEquals(1, c);
   }
 
+  @Test
+  void persistOffRampReason() {
+    Claim claim = new Claim();
+    claim.setBenefitClaimId("1234");
```
yoomlam, line 149: For clarity:
```suggestion
    vbmsId="1234"
    claim.setBenefitClaimId("1234");
```
and change line 155 below to use the `vbmsId` variable.

service/db/src/test/java/gov/va/vro/service/db/SaveToDbServiceImplTest.java
```diff
@@ -40,71 +35,146 @@ class SaveToDbServiceImplTest {
...
+    claim.setCollectionId("collection1");
+    claim.setIdType(MasAutomatedClaimPayload.CLAIM_V2_ID_TYPE);
+    claim.setVeteranIcn("v1");
+    claim.setDiagnosticCode("7101");
+    saveToDbService.insertClaim(claim);
+    ClaimEntity result1 = claimRepository.findByVbmsId("1234").orElseThrow();
```
yoomlam, line 155:
```suggestion
    ClaimEntity result1 = claimRepository.findByVbmsId(vbmsId).orElseThrow();
```

yoomlam (changes requested):
Lookin' good! Just minor suggestions.

app/src/test/java/gov/va/vro/controller/VroControllerTest.java
```diff
@@ -106,7 +105,7 @@ void postFullHealthAssessment() throws Exception {
     request.setDiagnosticCode("7101");
 
     var responseEntity1 =
-        post("/v2/health-data-assessment", request, FullHealthDataAssessmentResponse.class);
+        post("/v1/full-health-data-assessment", request, FullHealthDataAssessmentResponse.class);
```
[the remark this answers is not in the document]
hayleymdawson: Yes, so, this test while using some mocked items, tests the a v1 path. Previously because we didnt set id type to anything but default the v1 and v2 paths behaved identically so the test could have been argued to cover both (wrongly)

Once we corrected the id types to be correct for v1 vs v2 paths, this started to matter in the asserts. I actually missed a problem during development because I thought the test was fine and didn't discover it until I manually tested endpoints and checked the database. Once I fixed the problem I updated the strings so the test would be more clear what path it's going through.

service/provider/src/main/java/gov/va/vro/service/provider/camel/MasIntegrationProcessors.java
hayleymdawson: CollectionId is an integer in the mas payload so it is not a string here. At this point we are out of time I think to make more changes throughout the code for this PR to make it a string earlier. At least without serious risk to the deadline there is another PR stacked behind this one with more schema changes.

service/db/src/main/java/gov/va/vro/service/db/SaveToDbServiceImpl.java
```diff
@@ -105,6 +173,94 @@ public void insertEvidenceSummaryDocument(GeneratePdfPayload request, String doc
...
+      examOrderEntity.setStatus(examOrder.getStatus());
+      examOrderRepository.save(examOrderEntity);
+    }
+  }
+
+  @Override
+  public void insertFlashIds(List<String> veteranFlashIds, String veteranIcn) {
+    var veteran = veteranRepository.findByIcn(veteranIcn);
+    if (veteran.isEmpty()) {
+      log.warn("Could not find a Veteran with this ICN. Could not attach flash IDs.");
```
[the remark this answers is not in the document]
hayleymdawson: this has been changed to an error rather than change behavior more in this PR.

service/db/src/test/java/gov/va/vro/service/db/SaveToDbServiceImplTest.java
```diff
@@ -40,71 +35,146 @@ class SaveToDbServiceImplTest {
...
     Claim claim = new Claim();
-    claim.setClaimSubmissionId("claim1");
+    claim.setCollectionId("collection1");
+    claim.setIdType(Claim.V1_ID_TYPE);
     claim.setVeteranIcn("v1");
     claim.setDiagnosticCode("1234");
     var result = saveToDbService.insertClaim(claim);
     assertNotNull(result.getRecordId());
-    assertEquals(claim.getClaimSubmissionId(), result.getClaimSubmissionId());
+    assertEquals(claim.getBenefitClaimId(), result.getBenefitClaimId());
```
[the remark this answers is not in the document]
hayleymdawson: I was intentionally ensuring that the values were null. That said, result no longer *has* the claimSubmissionId on it in this line. it was moved to the other table so that line *must* be removed or changed. I think further rename refinements really need to wait to a later PR. 

I have added a comment however that this *really is* ensuring that null is a thing.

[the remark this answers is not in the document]
hayleymdawson: The subsequent line where we check claimSubmissionRepository for the collectionId or throw, is testing the equivalent of this line that we set the correct value on the correct table.

service/spi/src/main/java/gov/va/vro/service/spi/model/GeneratePdfPayload.java
```diff
@@ -27,6 +27,8 @@
 public class GeneratePdfPayload {
   @NonNull private String claimSubmissionId;
 
+  @Builder.Default @NotNull private String idType = Claim.V1_ID_TYPE;
```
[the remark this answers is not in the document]
hayleymdawson: Yes it's used in both V1 and v2 workflows. It is currently set to default and then v2 explicitly sets it.
I don't entirely disagree that this could be set explicitly in v1 but ensuring I get all the places this should be set at this late state is more risky than addressing it later imo.

shared/domain/src/main/java/gov/va/vro/model/mas/MasAutomatedClaimPayload.java
hayleymdawson: see earlier comment re: risk to timelines while not disagreeing. This PR didn't alter that field and I don't believe at this point it should.

app/src/test/java/gov/va/vro/service/MasProcessingServiceTest.java
```diff
@@ -16,6 +18,10 @@ public class MasProcessingServiceTest extends BaseIntegrationTest {
 
   @Autowired MasProcessingService masProcessingService;
 
+  @Autowired ClaimSubmissionRepository claimSubmissionRepository;
+
+  public static final String DEFAULT_ID_TYPE = "va.gov-Form526Submission";
```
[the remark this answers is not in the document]
hayleymdawson: Good catch. This actually needs to be the Claim v2 type. There was a silent error in the test where claimsubmissions may not be found, but iterating on an empty set doesn't cause an error. This has been corrected in both places now.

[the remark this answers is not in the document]
yoomlam: That makes much more sense given that this is a MAS-related test.

yoomlam (approved):
🎉

alan-amida:
@au2 and @rajesh-Krishnasamy , can y'all give this a little bit of time to look for and identify any potential gotchas that may impact iMVP PRODTEST/PROD?

If you don't find anything, please approve and lets get this merged in ASAP.

alan-amida:
oh @yoomlam would you mind terribly if you attach the associated GH issues that this PR would solve (strictly to your satisfaction) when you have a free moment? even if after this PR is merged is fine.

yoomlam:
Associated tickets are listed [here](https://github.com/department-of-veterans-affairs/abd-vro/pull/1046#issuecomment-1421096134)

rajesh-Krishnasamy (approved):

au2 (changes requested):
I ran the testAutomatedClaim in end 2 end test. This simulates a full iMVP run.
1)  AssesmentResult is not populated although I can see in there was an assessment in logs of vro-svc-assessor-dc7101-1. There appears to be nothing related to assessment in the event table either.
2) This test currently orders and exam but there is nothing in the exam order table.
3) What exactly is icn_timestamp? Icn is a unique medical id assigned during vha registration. Conceptually that does not make any sense. And it appears to be identical to updated_at (within miliseconds)
4) Participant id appears to be null although it was given in payload. Although it is "string" I expected to see it.
5) Filenumber is the main id we are using Did we make a conscious decision not to store Filenumber? 
6) I do not think we are getting flash_id's anymore. I need to double check how IBM sending this but it is not id anymore it is the name. I think they are putting into a different field.

Some cosmetic things
7) There is really no such thing called "vbms_id".  This is an id in CorpDb and I think column name is BNFT_CLAIM_ID.  I would just call it benefit_claim_id or corpdb_benefit_claim_id.
8) v1 vs v2 claim does not make much sense to me. I rather have differentiator to be id type. If va.gov calls health_assessment end point with claim_submission_id it will NOT be a "v2" claim id. It will still be "v1" id.

yoomlam:
Some responses:

> 3. What exactly is icn_timestamp? Icn is a unique medical id assigned during vha registration. Conceptually that does not make any sense. And it appears to be identical to updated_at (within miliseconds)

An ICN maps to 1 veteran. It's possible for a veteran to have multiple ICNs. See https://dsva.slack.com/archives/C01Q7979Z7D/p1676404743127689?thread_ts=1675982532.875369&cid=C01Q7979Z7D
and  https://dsva.slack.com/archives/C01Q7979Z7D/p1674237329439659?thread_ts=1674144499.310049&cid=C01Q7979Z7D:
> 7. Add icnTimestamp -- see [comment for 428](https://github.com/department-of-veterans-affairs/abd-vro/issues/428#issuecomment-1270578262):
> > VRO should NOT be seen as the source-of-truth for the Veteran's ICN, so we should make that more explicit with this icnTimestamp DB field.

(If we add comments along with the DB migrations, we can reduce some of these questions.)

> 5. Filenumber is the main id we are using Did we make a conscious decision not to store Filenumber?

Filenumber is PII so it should not be stored in the DB.

> 6. I do not think we are getting flash_id's anymore. I need to double check how IBM sending this but it is not id anymore it is the name. I think they are putting into a different field.

See https://github.com/department-of-veterans-affairs/abd-vro/pull/1112#discussion_r1109193380
We need to receive the flashId -- it is used near the beginning of the workflow for checking if a new claim is eligible of processing as a presumptive claim.
If we're getting it in a different field, we'll need to update the [API specs](https://github.com/department-of-veterans-affairs/abd-vro/wiki/MAS-api-spec-%28VRO-hosted-api%29) as well as the code.

> 
> Some cosmetic things 
> 7) There is really no such thing called "vbms_id". This is an id in CorpDb and I think column name is BNFT_CLAIM_ID. I would just call it benefit_claim_id or corpdb_benefit_claim_id. 

I worked on another VA project that called it the VBMS id. It can be renamed benefit_claim_id to be consistent with the payload or shorter corpdb_id.

alan-amida:
I'm going to merge this in now, noting the following:
- ExamOrder logic etc is in [this PR](https://github.com/department-of-veterans-affairs/abd-vro/pull/1107), which will also be merged in shortly
- We must look into the AssessmentResult not populating in the DB
- We must look into `participant_id` not populating in the Veteran table
- We must verify how IBM is sending us `flash_id` information
  - I know that they are sending us the "short name", but I did not know they were sending it in a different field in a different location within the schema.... please verify this information to us so we can make the changes quickly @au2 @rajesh-Krishnasamy 
  - either way, accommodating this change should be minimally invasive as long as it's still a part of the initial payload they send to us
- The field rename from VBMS to something else will happen as a patch update after code-freeze but before iMVP after we all agree on a naming convention
- v1/v2 claim update can likewise be updated after code-freeze but before iMVP after further discussion between Afsin and Yoom on this behavior (@au2 @yoomlam) -- but it sounds like we can update the logic to set the version with a new condition against the `idType` but still maintain much of the logic we already have (CC: @hayleymdawson @ianwolson)

I think there is a point about how `updated_at` could feasibly be used instead of `icn_timestamp` -- but I prefer to keep this logical separation in the rare event that a Veteran record gets updated for *other* reasons/columns/etc (especially if we end up adding more columns to the Veteran table for any unforeseen reason)

alan-amida:
@yoomlam Note that merging this has automatically closed those issues that I went in and attached to this PR -- please verify the truthiness of this. If they are not completed to your satisfaction, they'll need to be re-opened.

yoomlam:
> please verify the truthiness of this

@alan-amida There are many requirements in those tickets -- I can spot check some of them, but I believe QA is ultimately responsible for verifying that the requirements are completely fulfilled.

PR #1126, opened 2023-02-16
MCP-2371: Patch insufficient health data to order exam

jbregstone:
## What was the problem?

If sufficientForFastTracking is null it will probably cause an unexpected error. Without touching the camel routes we can patch this by creating an 'errorMessage' in the assessment services, as a temporary solution.

- [MCP-2371](https://amida.atlassian.net/browse/MCP-2371)

## How does this fix it?[^1]

More gracefully handle the case that sufficientForFastTracking is null by injecting an `errorMessage`. 

Error is caught in `service/provider/src/main/java/gov/va/vro/service/provider/services/HealthEvidenceProcessor.java`

```
if (evidence.getErrorMessage() != null) {
      log.error("Health Assessment Failed");
      throw new MasException("Health Assessment Failed with error:" + evidence.getErrorMessage());
    } else { ... }
```

## How to test this PR
- Step 1
- Step 2[^secrel]


yoomlam:
Is there a related PR that sets the offramp reason to `insufficientHealthDataToOrderExam`?

jbregstone:
> 

There is not, I am not familiar enough with camel make that change.

service-python/assessclaimdc7101/src/lib/main.py
```diff
@@ -83,6 +83,8 @@ def assess_sufficiency(event: Dict):
                     sufficient = True
             if bp_calculation["recentElevatedBpReadings"] >= 1 and bp_calculation["twoYearsBpReadings"] >= 3:
                 sufficient = True
+        if sufficient is None:
+            response_body["errorMessage"] = "insufficientHealthDataToOrderExam"
```
yoomlam, line 87: My concern is that this workflow path is not an error in the workflow -- it is a valid workflow path where VRO offramps the claim. Using `errorMessage` for this situation is not intuitive, esp. since `errorMessage` is used for errors like `error validating request message data`. An error suggests something needs to be fixed.

jbregstone: I agree, it is simply a workaround that will prevent uncaught exceptions. I can add a comment in the code with a to-do.

service-python/assessclaimdc7101/src/lib/main.py
```diff
@@ -83,6 +83,9 @@ def assess_sufficiency(event: Dict):
                     sufficient = True
             if bp_calculation["recentElevatedBpReadings"] >= 1 and bp_calculation["twoYearsBpReadings"] >= 3:
                 sufficient = True
+        # To DO: remove the following conditional. This should be handled in the camel routes. (HealthEvidenceProcessor)
```
yoomlam, line 86: IntelliJ highlights the `TODO` string
```suggestion
        # TODO: remove the following conditional. This should be handled in the camel routes. (HealthEvidenceProcessor)
```

rajesh-Krishnasamy (approved):

service-python/assessclaimdc7101/src/lib/main.py
rajesh-Krishnasamy: Please make the following change in the HealthEvidenceProcessor, if the errorMessage is set to "insufficientHealthDataToOrderExam".


public class HealthEvidenceProcessor implements Processor {

  /** Set the flag "sufficient for fast-tracking". */
  @Override
  public void process(Exchange exchange) {
    MasProcessingObject masTransferObject = (MasProcessingObject) exchange.getProperty("payload");

    AbdEvidenceWithSummary evidence = exchange.getMessage().getBody(AbdEvidenceWithSummary.class);

    if (evidence.getErrorMessage() != null) {
      **if ("insufficientHealthDataToOrderExam".equalsIgnoreCase(evidence.getErrorMessage())) {
        exchange.setProperty("sufficientForFastTracking", null);**
      } else {
        log.error("Health Assessment Failed");
        throw new MasException("Health Assessment Failed with error:" + evidence.getErrorMessage());
      }
    } else {
      exchange.setProperty("sufficientForFastTracking", evidence.isSufficientForFastTracking());
      log.info(
          " MAS Processing >> Sufficient Evidence >>> " + evidence.isSufficientForFastTracking());
      masTransferObject.setEvidence(getValidEvidence(evidence.getEvidence()));
      exchange.getMessage().setBody(masTransferObject);
    }
  }

rajesh-Krishnasamy (changes requested):

PR #1130, opened 2023-02-17
For non existant claimid return 400 not 500.

au2:
## What was the problem?

When MAS end point called with a claim id that does not exist we were returning 500.

- [MCP-2372](https://amida.atlassian.net/browse/MCP-000)

## How does this fix it?[^1]

MAS end point returns 400 if no claim.

## How to test this PR

There is a end to test added. Make sure that passes/

- Set ENV=end2end-test and build the application locally. 
- ./gradlew end2endTest


controller/src/main/java/gov/va/vro/controller/exception/GlobalExceptionHandler.java
```diff
@@ -141,14 +141,16 @@ public ResponseEntity<ClaimProcessingError> handleException(Exception exception)
...
+   * Handles BIP Exceptions.
+   *
+   * @param exception the exception
+   * @return returns response
+   */
   @ExceptionHandler(BipException.class)
   public ResponseEntity<ClaimProcessingError> handleBipClaimException(BipException exception) {
     log.error("BIP exception: {}", exception.getMessage(), exception);
     ClaimProcessingError cpe = new ClaimProcessingError(exception.getMessage());
-    return new ResponseEntity<>(
```
wlin2019, line 148: BIP does return 404 (NOT_FOUND) when it cannot find the claim for the given claim ID. It throws the exception for claim not found ONLY in some cases. If you remove this part of the code, it will make automatedClaim return 404 instead of 400 for those BIP return 404.

wlin2019 (commented):
For fixing 404 to 400, I think it fixes unusual cases but break more cases.

controller/src/main/java/gov/va/vro/controller/exception/GlobalExceptionHandler.java
au2: That is OK. It is worse than now which always throws 500.

wlin2019: This part has nothing to do with 500. It only changes 404 to 400 as you requested before. But it is really that important anyway.

wlin2019 (approved):

PR #1141, opened 2023-02-21
Fix sufficient for fast tracking when null and RFD when true

rajesh-Krishnasamy:
## What was the problem?
Order Exam only when Sufficient for Fast Tracking flag is false, Off Ramp if the value is null

- [MCP-2368](https://amida.atlassian.net/browse/MCP-2368)

## How does this fix it?[^1]
Order Exam only when Sufficient for Fast Tracking flag is false, Off Ramp if the value is null

## How to test this PR
- Step 1
- Step 2[^secrel]


service/provider/src/main/java/gov/va/vro/service/provider/services/HealthEvidenceProcessor.java
```diff
@@ -22,12 +22,16 @@ public void process(Exchange exchange) {
     AbdEvidenceWithSummary evidence = exchange.getMessage().getBody(AbdEvidenceWithSummary.class);
 
     if (evidence.getErrorMessage() != null) {
-      log.error("Health Assessment Failed");
-      throw new MasException("Health Assessment Failed with error:" + evidence.getErrorMessage());
+      if ("insufficientHealthDataToOrderExam".equalsIgnoreCase(evidence.getErrorMessage())) {
```
jbregstone, line 25: Instead of passing an error message from the assessment service, what if "null" was passed in as "sufficientForFastTracking" would we need to set the exchange property or would it already be set?

rajesh-Krishnasamy: Null can be passed as I changed the type of this field from primitive to Boolean object. The error message "insufficientHealthDataToOrderExam" is equal to Null being passed and is more verbose

wlin2019 (approved):
LGTM

au2 (changes requested):
Will update the end test. Do not merge.

au2:
e2e manually passed

PR #1143, opened 2023-02-21
Save Participant ID in Veteran Table

hayleymdawson:
## What was the problem?
Asfin noted that in v2 end to end testing that the participant id field in the table is not getting filled out. 

https://amida.slack.com/archives/C04NA99RLNT/p1676911117477589

## How does this fix it?[^1]
V2 now passes the participant id through mas processing. 
CAVEAT: If an ICN is re-used from a v1 test/call and in the database, participant ID will remain Null. 
V1 never sends participant id so this field is absolutely nullable.
If ICN data is not reused or the tables are cleared out you will see v2 populate the participant ids correctly. noted here
https://amida.slack.com/archives/C039KH7HY79/p1676997807160709

## How to test this PR
You can run the end to end full suite and at least one veteran in the current data will populate.
If you use a new ICN via the mas endpoint you will also see participant id filled out.
If you clear the database beforehand you can just run the Vro2 end to end tests manually and it also shows the data filling out.


yoomlam (approved):
LGTM. Didn't test the code.

alan-amida (approved):
LGTM 👍

hayleymdawson:
MAS-2383 Is the associated Jira Ticket.

PR #1187, opened 2023-02-27
MCP-2409 PDF Template Updates

anthonyvelazquez:
## What was the problem?

Amy requested further updates to the PDF Templates

- Re-name the Page(s) column to be: Page number(s) / Event ID(s)

- If the document ID is empty for "VAMC Other Output Reports" documents (the HDR record), display Pending upload to VBMS eFolder in the Document ID column. If it's not empty, display the Document ID.

- For health evidence data we're getting from Lighthouse, display n.a. - data pulled directly from VistA in the Document ID column. This will replace “n.a. - see VistA/Capri to verify”.

- [Adjustments to the column widths](https://www.figma.com/file/gENy7VJcPxq35DmqCit5av/iMVP-designs?node-id=509%3A269&t=CEzF0pnNYODxntbH-1) to accommodate "event IDs" that are being displayed in the page column. See linked Figma file for specified column widths.

- [MCP-2409](https://amida.atlassian.net/browse/MCP-249)


service-python/pdfgenerator/src/lib/templates/hypertension-v2.html
```diff
@@ -306,7 +306,23 @@ <h2 class="m-0 mb-13"><b>Schedular evidence for hypertension</b></h2>
             <td {% if medication.organization == "" and medication.dataSource == "MAS" %}class="empty" {% endif %}>{% if not medication.organization and medication.dataSource == "MAS" %}n.a.{% elif not medication.organization and medication.dataSource == "LH" %} VAMC record {% elif medication.organization and medication.dataSource == "LH" %} VAMC record - {{ medication.organization }}{% else %}{{ medication.organization }}{% endif %}</td>
             <td {% if medication.receiptDate == "" %}class="empty" {% endif %}>{% if medication.receiptDate == ""%}n.a.{% else %}{{ medication.receiptDate }}{% endif %}</td>
             <td {% if medication.page == "" %}class="empty" {% endif %}>{% if medication.page == ""%}n.a.{% else %}{{ medication.page }}{% endif %}</td>
-            <td {% if medication.document == "" %}class="empty" {% endif %}>{% if medication.document == ""%}n.a.{% else %}{{ medication.document }}{% endif %}</td>
+            <td {% if medication.document == "" %}class="empty" {% endif %}>
+              {% if medication.dataSource == "LH" %}
+                {% if medication.organization == "VAMC Other Output Reports" and medication.document == "" %}
```
jbregstone, line 311: I believe  "VAMC Other Output Reports" is coming from MAS not LH. 

image (4)

jbregstone (changes requested):

jbregstone:
#1201

service-python/pdfgenerator/src/lib/templates/hypertension-v2.html
```diff
@@ -306,7 +306,23 @@ <h2 class="m-0 mb-13"><b>Schedular evidence for hypertension</b></h2>
...
+                {% elif medication.document == "" %}
+                  n.a.
+                {% else %}
+                  {{ medication.document }}
+                {% endif %}
+              {% else %}
+                {% if medication.document == ""%}
+                  n.a.
+                {% else %}
+                  {{ medication.document }}
```
jbregstone, line 322: I believe we need to implement this logic for BP readings as well as medication. I am not sure if it is needed for conditions.

au2 (changes requested):
VAMC Other Output Reports  still showing Pulled directly from EHR

au2 (changes requested):
Now LH data rows show "na" in the document id. That needs to be fixed with the new message.

au2 (changes requested):
Also wrong is the VAMC Other output Reports document id in "Other medical Evidence for hypertension"

service-python/pdfgenerator/src/lib/templates/hypertension-v2.html
```diff
@@ -271,7 +271,22 @@ <h2 class="m-0 mb-13"><b>Schedular evidence for hypertension – blood pressure
             <td {% if bp_reading.organization == "" and bp_reading.dataSource == "MAS"  %}class="empty" {% endif %}>{% if not bp_reading.organization and bp_reading.dataSource == "MAS" %}n.a.{% elif not bp_reading.organization and bp_reading.dataSource == "LH" %} VAMC record {% elif bp_reading.organization and bp_reading.dataSource == "LH" %} VAMC record - {{ bp_reading.organization }}{% else %}{{ bp_reading.organization }}{% endif %}</td>
             <td {% if bp_reading.receiptDate == "" %}class="empty" {% endif %}>{% if bp_reading.receiptDate == ""%}n.a.{% else %}{{ bp_reading.receiptDate }}{% endif %}</td>
             <td {% if bp_reading.page == "" %}class="empty" {% endif %}>{% if bp_reading.page == ""%}n.a.{% else %}{{ bp_reading.page }}{% endif %}</td>
-            <td {% if bp_reading.document == "" %}class="empty" {% endif %}>{% if bp_reading.document == ""%}n.a.  - see VistA/Capri to verify{% else %}{{ bp_reading.document }}{% endif %}</td>
+            <td {% if bp_reading.document == "" %}class="empty" {% endif %}>
+              {% if bp_reading.dataSource == "MAS" %}
```
jbregstone, line 275:
```suggestion
```

jbregstone: I think the logic could be simplified if this `if` statement is removed. Then there would be no need for the else block because it would be handled the same by the `elif` and `else` in this block.

au2 (changes requested):
I do not think Amy specified what to do when MAS document other than VAMC Other Output Reports has no document Id. Probably will never happen but can  we put "Pending upload to VBMS eFolder" instead of "data pulled directly from VistA". If it happens they will not be confused again.

jbregstone (approved):

service-python/pdfgenerator/src/lib/templates/hypertension-v2.html
anthonyvelazquez: After Afsin's text update request, now the conditional branches are different so I'll leave it for now. I'll remove if there's another update that make the both branches the same again

au2 (approved):

PR #1210, opened 2023-03-01
End to End test of Order Exam path

hayleymdawson:
## What was the problem?
No end to end testing of the order exam path

- [MCP-2423](https://amida.atlassian.net/browse/MCP-2423)

## How does this fix it?
Fixes V2 test issues due to upstream data changes
Fixes V1 test issues due to upstream data changes
Adds end to end testing for the positive test path. negative path will come in MCP-2424

## How to test this PR
Run End to end Tests


app/src/end2endTest/resources/test-6602-01/assessment.json
```diff
@@ -1,165 +1,89 @@
 {
-  "veteranIcn": "9000682",
```
hayleymdawson, line 2: Changed this file to what we get back from upstream

mock-lighthouse-api/src/test/java/gov/va/vro/mocklh/ApplicationTest.java
```diff
@@ -223,7 +223,7 @@ void icn9000682MedicationRequestTest() {
     TestSpec spec = TestSpec.builder().icn("9000682").resourceType("MedicationRequest").build();
 
     JsonNode bundle = getBundle(spec);
-    verifyBundle(bundle, 14, 7);
+    verifyBundle(bundle, 11, 8);
```
hayleymdawson, line 226: Fixes for upstream data

mock-lighthouse-api/src/main/java/gov/va/vro/mocklh/config/AppConfig.java
```diff
@@ -27,7 +27,10 @@ public MockBundleStore muckBundleStore() throws IOException {
     MockBundleStore store = new MockBundleStore();
 
     MockBundles mb1 = MockBundles.of("mock-bundles/mock1012666073V986297");
-    store.put("mock1012666073V986297", mb1);
+    store.put("1012666073V986297", mb1);
```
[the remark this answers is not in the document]
au2: i used mock as part of icn to indicate it is a mock. this is correct and was working. something else is broken here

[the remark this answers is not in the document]
au2: 376 is not working locally for me probably because of this change.

au2 (changes requested):

mock-lighthouse-api/src/main/java/gov/va/vro/mocklh/config/AppConfig.java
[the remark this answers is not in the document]
hayleymdawson: For future reference of anyone:

Mock services pass through to a sandbox if the id does not exist in the mock. Test 376 and a few others *intentionally* do this. The mock preprend prevents a collision.

377 and 378 will update to use a "mock<icn>" everywhere as the correct way to handle this.

app/src/end2endTest/java/gov/va/vro/end2end/VroV2Tests.java
```diff
@@ -134,6 +152,28 @@ private void testAutomatedClaimFullPositive(String collectionId, boolean expecte
     String expectedMessage = String.format("Received Claim for collection Id %s.", collectionId);
     assertEquals(expectedMessage, masResponse.getMessage());
 
+    boolean successOrdering = false;
+    if (expectedExamOrder != null) {
```
alan-amida, line 156: I feel like this piece can be pulled out into a reusable private helper method... but don't feel strongly about it enough to make a formal change request.

alan-amida (approved):
Code looks good. Found only one minor thing which I can live without.

app/src/end2endTest/java/gov/va/vro/end2end/VroV2Tests.java
```diff
@@ -218,6 +258,21 @@ void testAutomatedClaim_missingAnchor() {
         masResponse.getMessage());
   }
 
+  // Test Case that ensures that exam order *is* callled
+  // The data underlying follows the NEW claim, one relevant condition, not enough information path.
+  @Test
+  void testAutomatedClaim_orderExamNewClaim() {
+    testAutomatedClaimFullPositive("377", false, true);
+  }
```
au2, line 266: If you give the second parameter false the check for pdf will not be done.  I would reorganize these parameter or modularize testAutomatedClaimFullPositive to check pdf generation as well since ordering exam also puts a pdf in efolder so we want to check that.

app/src/end2endTest/java/gov/va/vro/end2end/VroV2Tests.java
```diff
@@ -218,6 +258,21 @@ void testAutomatedClaim_missingAnchor() {
...
+
+  // Test case that ensures the exam order *is* callled
+  // The data underlying follows the "increase" claim path where not enough blood pressure readings
+  // exist.
+  @Test
+  void testAutomatedClaim_orderExamIncreaseClaim() {
+    testAutomatedClaimFullPositive("378", false, true);
+  }
```
au2, line 274: See the comment above.

app/src/end2endTest/resources/test-mas/claim-378-7101.json
```diff
@@ -0,0 +1,29 @@
...
+  "collectionId": 378,
+  "veteranIdentifiers": {
+    "icn": "mock1012666073V986377",
+    "ssn": "111-11-1377",
+    "veteranFileId": "9999377",
+    "edipn": "1010234-377",
+    "participantId": "88888377"
+  },
+  "claimDetail": {
+    "benefitClaimId": "1040",
```
au2, line 15: Now that we have mock lighhouse server I would not use the same benefit id for different end to end tests.  If nothing else to test database population is correct. Looking at this ICN's are identical to 377 as well so it looks ike copy and paste error. Looks like this is increase and the previous one presuptive. I would use different patients and make sure the  to choose a condition that does not intersects. One of them looks three years and other looks at two years I think.

mock-lighthouse-api/src/main/java/gov/va/vro/mocklh/config/AppConfig.java
```diff
@@ -29,6 +29,9 @@ public MockBundleStore muckBundleStore() throws IOException {
     MockBundles mb1 = MockBundles.of("mock-bundles/mock1012666073V986297");
     store.put("mock1012666073V986297", mb1);
 
+    MockBundles mb2 = MockBundles.of("mock-bundles/mock1012666073V986377");
+    store.put("mock1012666073V986377", mb2);
+
```
au2, line 34: Yeah, i would use different patients for different end to end tests.

mock-mas-api/src/main/java/gov/va/vro/mockmas/controller/MockMasController.java
```diff
@@ -38,6 +42,8 @@ public class MockMasController {
   private final MasApiService apiService;
   private final CollectionStore store;
 
+  @Autowired private final ExamOrderStore examOrderStore;
```
au2, line 45: You do not need @Autowired here. "private final" will cause constructor injection since we have the @RequiredArgsConstructor as well. Constructor injection is preferred.

app/src/end2endTest/java/gov/va/vro/end2end/VroV2Tests.java
```diff
@@ -134,6 +152,28 @@ private void testAutomatedClaimFullPositive(String collectionId, boolean expecte
...
+                request.getCollectionId());
+          }
+        }
+      } else {
+        log.info("Negative test case for exam ordering TBD via database check. Skipping polling");
+      }
+      assertEquals(successOrdering, expectedExamOrder);
+    }
+
     if (!expectedStatusUpdate) {
```
au2, line 177: This should not happen for cases which orders exam since we also have pdf generator adnd we want to test that.

au2 (changes requested):
I added some comments inline. Also consider adding something like restTemplate.delete(ORDER_EXAM_CHECK_END_POINT +CollectionId); for the exam flag similar to other deletes at the start. It comes handy if the end to end test is being used for testing and debugging locally  and run multiple times. That way it does not give you false positives from previous runs.

app/src/end2endTest/resources/test-mas/claim-378-7101.json
```diff
@@ -0,0 +1,29 @@
...
+  "collectionId": 378,
+  "veteranIdentifiers": {
+    "icn": "mock1012666073V986378",
+    "ssn": "111-11-1378",
+    "veteranFileId": "9999378",
+    "edipn": "1010234-378",
+    "participantId": "88888378"
+  },
+  "claimDetail": {
+    "benefitClaimId": "1040",
```
au2, line 15: You need to change either this or the previous 1041. Whatever the intention was.

au2 (changes requested):
Looks like 1041 claim is created but now used.

au2 (approved):

PR #1239, opened 2023-03-07
Save Exam Order Status when VRO initializes order to MAS

hayleymdawson:
## What was the problem?
Previously the exam order table did not get an entry until MAS called for a status from us. 
Now VRO will put an entry into the table when we send the order to MAS.
If MAS asks us for a status we are not aware of we still save the row in the table. 

- [MCP-2470](https://amida.atlassian.net/browse/MCP-2470)
- https://amida.slack.com/archives/C01Q7979Z7D/p1678153769230959

## How does this fix it?
Hooked into the code where we send the order and create an object to save to the database. Requires changes only in that file as the database file can already handle this call. 

## How to test this PR
Run end to end tests for VroV2. 
Claim id 377, 378, and 350 will all land in ORDER_SUBMITTED status as we initiate the order.
This can also be verified in the vro logs as the log message has changed to indicate we are both making the order and saving to the database at that point.

### DIFF
--- a/service/db/src/main/java/gov/va/vro/service/db/SaveToDbServiceImpl.java
+++ b/service/db/src/main/java/gov/va/vro/service/db/SaveToDbServiceImpl.java
@@ -373,7 +373,8 @@ public class SaveToDbServiceImpl implements SaveToDbService {
   }
 
   private ExamOrderEntity createExamOrder(ExamOrder examOrder) {
-    // Currently ExamOrders only come from MAS or after VRO successfully requests an exam order from MAS
+    // Currently ExamOrders only come from MAS or after VRO successfully requests an exam order from
+    // MAS
     Optional<ClaimSubmissionEntity> claimSubmission =
         claimSubmissionRepository.findFirstByReferenceIdAndIdTypeOrderByCreatedAtDesc(
             examOrder.getCollectionId(), examOrder.getIdType());
--- a/service/db/src/main/java/gov/va/vro/service/db/SaveToDbServiceImpl.java
+++ b/service/db/src/main/java/gov/va/vro/service/db/SaveToDbServiceImpl.java
@@ -373,7 +373,7 @@ public class SaveToDbServiceImpl implements SaveToDbService {
   }
 
   private ExamOrderEntity createExamOrder(ExamOrder examOrder) {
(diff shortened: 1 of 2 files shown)


ianwolson (approved):

service/db/src/main/java/gov/va/vro/service/db/SaveToDbServiceImpl.java
```diff
@@ -373,7 +373,7 @@ private ClaimEntity createClaim(Claim claim, VeteranEntity veteranEntity) {
   }
 
   private ExamOrderEntity createExamOrder(ExamOrder examOrder) {
-    // Currently ExamOrders only come from MAS
+    // Currently ExamOrders only come from MAS or VRO sending to MAS
```
yoomlam, line 376:
```suggestion
    // Currently ExamOrders only come from MAS or after VRO successfully requests an exam order from MAS
```

service/provider/src/main/java/gov/va/vro/service/provider/MasOrderExamProcessor.java
```diff
@@ -42,7 +49,15 @@ public void process(Exchange exchange) {
...
       masOrderExamRequest.setConditions(List.of(masOrderExamConditions));
       var response = masApiService.orderExam(masOrderExamRequest);
-      log.info("Order Exam Response :  " + response);
+      log.info("Order Exam Response :  " + response + " saving as ORDER_SUBMITTED");
+      ExamOrder examOrder =
+          ExamOrder.builder()
+              .collectionId(Integer.toString(claimPayload.getCollectionId()))
+              .idType(MasAutomatedClaimPayload.CLAIM_V2_ID_TYPE)
+              .status(SUBMITTED_STATUS)
+              .examOrderDateTime(OffsetDateTime.now())
```
yoomlam, line 58: If I understand correctly, the `examOrderDateTime` should be populated with the datetime from MAS when it successfully orders the exam. This place in the code represents when VRO requests the exam to be ordered.
```suggestion
              .examOrderDateTime(OffsetDateTime.now())
```

yoomlam (changes requested):
Change requested if I understand correctly.

yoomlam (approved):
Thanks for the quick turnaround!

PR #1241, opened 2023-03-07
Expose Exam Order Metric Table for QA/End to End Testing

hayleymdawson:
## What was the problem?
End to End testing to verify exam order entries required a developer to check the database. QA testers need an endpoint to check this. 

- [MCP-2488](https://amida.atlassian.net/browse/MCP-2488)

## How does this fix it?
New claim metric endpoint that pages through all exam order entries. 

## How to test this PR
Deploy, use swagger as the easiest method to test the end point. 

Still in Draft Due to:
Needs unit tests, want feedback on any functionality changes earlier.


service/db/src/main/java/gov/va/vro/service/db/mapper/ExamOrderInfoResponseMapper.java
```diff
@@ -0,0 +1,37 @@
...
+      target = "hasAssociatedClaimSubmission",
+      expression = "java(examOrderEntity.getClaimSubmission() != null)")
+  @Mapping(
+      target = "orderedAt",
+      expression = "java(convertOrderedAt(examOrderEntity.getOrderedAt()))")
+  ExamOrderInfoResponse toExamOrderInfoResponse(ExamOrderEntity examOrderEntity);
+
+  List<ExamOrderInfoResponse> toExamOrderInfoResponses(Iterable<ExamOrderEntity> examOrderEntities);
+
+  default LocalDateTime convertOrderedAt(OffsetDateTime offsetOrdered) {
```
alan-amida, line 31: I am wondering if MapStruct actually does this conversion automatically, especially since OffsetDateTime has its own built-in `toLocalDateTime()` method

If so this might kill the need to declare this default method and we would be able to just have a flat declaration like
```Java
@Mapping(
  target = "orderedAt",
  source = "orderedAt")
```
instead....

There's something about String-based method call expressions in Java that doesn't sit right with me. But if we must, we must.

alan-amida (approved):
Only one thing that *may* be a refactorable thing??? Would actually probably need to test it to find out if it would work the way I think it might. Otherwise, no biggie.

This looks good to me. 👍

service/db/src/main/java/gov/va/vro/service/db/mapper/ExamOrderInfoResponseMapper.java
hayleymdawson: So digging into the map struct documentation it can do quite a number of translations automatically around primatives, their corresponding class types, and things to strings.

The list of date compatibilities is long but unfortunately it looks to be centered more around older joda and java 7 dates to local date time in java 8, with nothing listed for offset date time. That leads me to believe it wont work, although we could verify by testing, that's a fair chunk of day to remove one line of code on an off chance.

alan-amida (approved):
LGTM 👍

au2 (approved):

PR #1245, opened 2023-03-08
Test off ramp End to End Testing

hayleymdawson:
## What was the problem?
Previously, the code could not check for the assessment result sufficient flag directly to know the end of processing had occurred and check that it is correct. 

- [MCP-2424](https://amida.atlassian.net/browse/MCP-2424)

## How does this fix it?
Wires up end to end testing against the claim-info api. 
Exposes the sufficient flag in the claim info api
And also solves a problem that I inadvertently discovered on database reads showing more results in only *some* cases than were actually in the database itself. 

## How to test this PR
Run end to end tests.
You can also use the claim-info api for a given claimsubmissionId

### DIFF
--- a/app/src/end2endTest/java/gov/va/vro/end2end/VroV2Tests.java
+++ b/app/src/end2endTest/java/gov/va/vro/end2end/VroV2Tests.java
@@ -234,7 +234,7 @@ public class VroV2Tests {
             log.error(
                 "CollectionId {} came back with more than one contention. Cannot determine which one to check",
                 collectionId);
-              break;
+            break;
           }
         }
       } catch (Exception exception) {
--- a/app/src/end2endTest/java/gov/va/vro/end2end/VroV2Tests.java
+++ b/app/src/end2endTest/java/gov/va/vro/end2end/VroV2Tests.java
@@ -396,7 +396,8 @@ public class VroV2Tests {
   @SneakyThrows
    @Test
   void testAutomatedSufficiencyIsNull() {
-    testAutomatedClaimFullPositive("500", true);
+    // Offramp claims do not go through pdf process per VRO workflow diagram.
+    testAutomatedClaimFullPositive("500", false);
     testClaimSufficientStatus("500", null);
   }
 }
--- a/app/src/end2endTest/java/gov/va/vro/end2end/VroV2Tests.java
+++ b/app/src/end2endTest/java/gov/va/vro/end2end/VroV2Tests.java
@@ -16,6 +16,7 @@ import gov.va.vro.model.claimmetrics.response.ClaimInfoResponse;
 import gov.va.vro.model.mas.request.MasAutomatedClaimRequest;
 import lombok.SneakyThrows;
 import lombok.extern.slf4j.Slf4j;
+import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 import org.springframework.http.HttpEntity;
 import org.springframework.http.HttpHeaders;
@@ -49,13 +50,6 @@ public class VroV2Tests {
 
   private final RestTemplate restTemplate = new RestTemplate();
 
-  // Authorization for Claim Info e2e since it cannot be accessed unauthorized.
-  private HttpEntity<Void> getAuthorizationHeader() {
-    HttpHeaders headers = new HttpHeaders();
-    headers.add("X-API-Key", "test-key-01");
-    return new HttpEntity<>(headers);
-  }
-
   /*
    * This test checks the RequestBodyAdvice sanitizing logic for disallowed characters.
    * Eventually we should refactor this out into its own test suite with other endpoints
@@ -64,7 +58,7 @@ public class VroV2Tests {
   @Test
   void testExamOrderingStatus_disallowedCharacters() {
     var request = getOrderingStatusDisallowedCharacters();
-    var requestEntity = getEntity(request);
+    var requestEntity = getBearerAuthEntity(request);
     try {
       restTemplate.postForEntity(EXAM_ORDERING_STATUS_URL, requestEntity, String.class);
       fail("Should have thrown exception");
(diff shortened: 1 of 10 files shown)


persistence/model/src/main/java/gov/va/vro/persistence/model/ClaimEntity.java
```diff
@@ -30,18 +32,25 @@ public class ClaimEntity extends BaseEntity {
 
   @ManyToOne private VeteranEntity veteran;
 
+  // Multiple collections need to be eagerly fetched without causing a cartesian join product with
```
hayleymdawson, line 35: This hasn't shown up int he testing so i thought since we only grab one claim at a time we avoided it. turns out we do not always, and this should be fixed sooner than later.

alan-amida:
https://github.com/department-of-veterans-affairs/abd-vro/actions/runs/4367100607/jobs/7637977766#step:5:3328
```
Execution failed for task ':persistence:model:jacocoTestCoverageVerification'.
> Rule violated for bundle model: instructions covered ratio is 0.7, but expected minimum is 0.8
```

app/src/end2endTest/java/gov/va/vro/end2end/VroV2Tests.java
```diff
@@ -43,6 +49,13 @@ public class VroV2Tests {
 
   private final RestTemplate restTemplate = new RestTemplate();
 
+  // Authorization for Claim Info e2e since it cannot be accessed unauthorized.
+  private HttpEntity<Void> getAuthorizationHeader() {
+    HttpHeaders headers = new HttpHeaders();
+    headers.add("X-API-Key", "test-key-01");
+    return new HttpEntity<>(headers);
+  }
```
alan-amida, line 57: Minor OCD thing: can we move this down to where Dimitri's `private static HttpHeaders getHttpHeaders()` method is (to keep like-minded code together in case we gotta move it somewhere else later) and probably rename his old one to something more descriptive like `getBearerAuthHeaders` while we're there? Looks like he's only using it in one place.

image

alan-amida (approved):
Minor thing. Otherwise LGTM! 👍

au2 (commented):
Not for this pr but consider something like this https://github.com/Skellington-Closet/slack-mock for future iterations since slack message is the last step in off-ramping. It also gives an opportunity to test/verify what is in the slack messages.

au2 (changes requested):
Not fully related to this pr but saw when starting reviewing. Can you change HttpStatus.BAD_GATEWAY return status in CalimMetricsController.claimInfoForClaimId. It appears to be a typo.

service/db/src/test/java/gov/va/vro/service/db/mapper/ClaimInfoResponseMapperTest.java
```diff
@@ -43,9 +43,11 @@ void testMain() {
     ContentionEntity contentionEntity = new ContentionEntity();
     contentionEntity.setDiagnosticCode(diagnosticCode);
 
+    final Boolean sufficientEvidence = true;
     AssessmentResultEntity assessmentResultEntity = new AssessmentResultEntity();
     Map<String, String> areEvidenceInfo = ImmutableMap.of("k1", "v1", "k2", "v2");
     assessmentResultEntity.setEvidenceCountSummary(areEvidenceInfo);
+    assessmentResultEntity.setSufficientEvidenceFlag(sufficientEvidence);
```
au2, line 50: This is good except it looks to me that we are not setting it in the real code; i.e SaveDbServiceImpl.insertAssessmentResult. So your end to end test will always pass since it is based on setSufficientEvidenceFlag.

au2 (changes requested):
You need to add setSufficientEvidenceFlag in real code I think. SaveDbServiceImpl.insertAssessmentResult. That is critical, others minor.

service/db/src/test/java/gov/va/vro/service/db/mapper/ClaimInfoResponseMapperTest.java
hayleymdawson: We do set it in real code. We simply do not set it on immediate creation like the mapper test does, but it's trying to test the mappings so I don't see any need to mimic an e2e process here.

As for where this update happens:

There is a method in SavetoDbServiceImpl.java with this signature:
  public void updateSufficientEvidenceFlag(AbdEvidenceWithSummary evidence, String diagnosticCode) {

Which is called by MasAssessmentResultProcessor

Now why we make 2 calls instead of all at once I do not know, possibly to not break other paths in v1.

app/src/end2endTest/java/gov/va/vro/end2end/VroV2Tests.java
```diff
@@ -189,6 +202,52 @@ public void testExamOrdered(String collectionId, boolean expectedExamOrder) {
...
+
+    String url = CLAIM_INFO_URL + collectionId;
+    AssessmentInfo foundAssessment = null;
+    HttpEntity<Void> requestEntity = getAuthorizationHeader();
+
+    log.info("Waiting for claim processing to finish and assessment database results");
+    for (int pollNumber = 0; pollNumber < 15; ++pollNumber) {
+      Thread.sleep(20000);
+      try {
+        ResponseEntity<ClaimInfoResponse> testResponse =
```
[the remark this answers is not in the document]
hayleymdawson: it is. i hadn't noticed the typo and thus was unsure what error codes i would receive.

service/db/src/test/java/gov/va/vro/service/db/mapper/ClaimInfoResponseMapperTest.java
hayleymdawson: it looks like there's only one use of both save to dbserviceimpl calls so i should just collapse them into one method however.

au2 (approved):

PR #1247, opened 2023-03-08
MCP-2038 move MAS related env variables

ianwolson:
## What was the problem?

- [MCP-2038](https://amida.atlassian.net/browse/MCP-2038)

## How to test this PR
- Step 1
- Step 2[^secrel]


au2 (changes requested):
You cannot just remove MAS_API_BAS_URL. It has a different value in prodtest.

alan-amida:
Ian, please follow up with Afsin's comment and bug him if you need clarification.

au2:
@ianwolson if you look at helm file deployment.yaml you will see that MAS_API_BAS_URL is set from the secrets in the deployments. The value of that secret (mas.masApiBaseUrl) is different in prodtest and prod than https://viccs-api-dev.ibm-intelligent-automation.com/pca/api/dev. So if you hardcode masAPIProvider.baseUrl in application.yml that is what is going to be used in prodtest and production and end to end tests will fail. So you cannot just hard code the MAS development url in application.yml.

You can try leaving it as it is (env variable with the default value) and see if that works for local deployment (after this 350 bug is fixed - that is the only place this setting is being used currently locally). If it does great that is the solution. If not you can hardcode in application-local.yml probably.

PR #1252, opened 2023-03-09
Use RabbitMqCamelUtils.fromRabbitmq

yoomlam:
## What was the problem?

There's a gotcha (potential bug) when a Camel route receives from RabbitMQ and later in the route sends to RabbitMQ. The `CamelRabbitmqExchangeName` and `CamelRabbitmqRoutingKey` headers (which Camel automatically adds) will override the exchange and routing key in the URI in a `to("rabbitmq:<someURI>")` call, which results in a loop.

I added this comment to the code:
```
        // Good practice to remove the CamelRabbitmqExchangeName and CamelRabbitmqRoutingKey so it
        // doesn't interfere with subsequent sending to rabbitmq endpoints
        // https://camel.apache.org/components/3.19.x/rabbitmq-component.html#_troubleshooting_headers:
        // > if the source queue has a routing key set in the headers, it will pass down to
        // > the destination and not be overriden with the URI query parameters.
```

## How does this fix it?

Replaces `from("rabbitmq:...")` code with a `RabbitMqCamelUtils.fromRabbitmq` utility that removes the headers.

Also promote domain-specific `VroCamelUtils.java` to a shared library `RabbitMqCamelUtils.java`.

## How to test this PR
Run VRO and exercise the routes with changed code.


yoomlam:
[PR: End-to-end test](https://github.com/department-of-veterans-affairs/abd-vro/actions/runs/4378502375) passed.

PR #1265, opened 2023-03-12
Gradle: remove org.springdoc as a default dependency

yoomlam:
## What was the problem?

New Java-based containers has to exclude `SpringDocHateoasConfiguration` b/c the Spring libraries are included be default.

## How does this fix it?

Remove `org.springdoc:*` as a default dependency. It's only needed in the main app.

Similarly, remove `spring-boot-starter-data-jpa` from mocks so that `DataSourceAutoConfiguration` doesn't need to be excluded.

### DIFF
--- a/buildSrc/src/main/groovy/local.std.java.application-conventions.gradle
+++ b/buildSrc/src/main/groovy/local.std.java.application-conventions.gradle
@@ -37,6 +37,12 @@ plugins {
 }
 
 dependencies {
+    implementation "org.zalando:problem-spring-web"
+    implementation "org.springdoc:springdoc-openapi-ui"
+    implementation "org.springdoc:springdoc-openapi-webmvc-core"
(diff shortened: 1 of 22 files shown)


buildSrc/src/main/groovy/local.std.java.application-conventions.gradle
```diff
@@ -37,6 +37,12 @@ plugins {
 }
 
 dependencies {
+    implementation "org.zalando:problem-spring-web"
+    implementation "org.springdoc:springdoc-openapi-ui"
+    implementation "org.springdoc:springdoc-openapi-webmvc-core"
+    //implementation "org.springdoc:springdoc-openapi-security"
+    implementation "org.springdoc:springdoc-openapi-data-rest"
```
yoomlam, line 44: Moved from `buildSrc/src/main/groovy/starter.java.deps-build-conventions.gradle`

PR #1307, opened 2023-03-20
MCP-2558 422 Status Code

anthonyvelazquez:
## What was the problem?

Previously returning 200 for scenarios where we are not able to process the request

- [MCP-2558](https://amida.atlassian.net/browse/MCP-2558)

## How does this fix it?[^1]

This code indicates VRO received a claim it cannot process.

These can be the reasons based on the payload:

- “diagnosticCode” is not equal to “7101”.
- “disabilityActionType” is not equal to one of “NEW” or “INCREASE”.
- “disabilityActionType” is “NEW” but “veteranFlashIds” does not include “Agent Orange Exposure Verified” or “266”.

These can be the reasons based on verification with “BIP Claims
API”:

- Claim temporary station of jurisdiction is not “398”.
- Claim contentions does not contain special issue codes “RRD” or “RDR1”
- Claim has multiple contentions.

MAS is not expected to send these cases in production and thus
is not expected to receive 422’s but should be prepared to take
back ownership of the claim and go forward with the processing
in case it happens.

## How to test this PR
- Step 1
- Step 2[^secrel]


service/provider/src/main/java/gov/va/vro/service/provider/mas/service/MasProcessingService.java
ianwolson, line 1: if (!bipClaimService.hasAnchors(claimId))
Inside this method is where we need to do the checks and return false

shared/domain/src/main/java/gov/va/vro/model/mas/request/MasAutomatedClaimRequest.java
```diff
@@ -59,4 +60,15 @@ public class MasAutomatedClaimRequest {
...
   @Schema(description = "Veteran Flash Ids", example = "[\"999\"]")
   private List<String> veteranFlashIds;
+
+  public Boolean canProcess() {
+    List<String> validDisabilityActionTypes = Arrays.asList("NEW", "INCREASE");
+    if (!claimDetail.getConditions().getDiagnosticCode().equals("7101")) return false;
+    else if (!validDisabilityActionTypes.contains(
+        claimDetail.getConditions().getDisabilityActionType())) return false;
+    else if (claimDetail.getConditions().getDisabilityActionType().equals("NEW")
+        && (!veteranFlashIds.contains("Agent Orange Exposure Verified")
```
au2, line 70: This case will need to be processed seperate. For this case we will off-ramp as before; will not return 422.

au2 (changes requested):
See an inline comment. Also all end to end tests that attend to these cases will need to be updated.

app/src/test/java/gov/va/vro/service/BipClaimServiceTest.java
```diff
@@ -61,11 +58,17 @@ void hasAnchorsMissingSpecialIssue() throws BipException {
                 createContention(List.of("RRD", "OTHER"))));
 
     BipClaimService claimService = new BipClaimService(claimProps, bipApiService, null, null);
-    assertFalse(claimService.hasAnchors(bipClaimId));
+    Exception exception =
+        assertThrows(
```
hayleymdawson, line 62: Just trying to match code to spec, what are "anchors?"

hayleymdawson (dismissed):
spec issue regarding offramp needs to be either fixed in ticket/documentation or in code.

shared/domain/src/main/java/gov/va/vro/model/mas/request/MasAutomatedClaimRequest.java
anthonyvelazquez: Fixed

app/src/test/java/gov/va/vro/controller/MasControllerTest.java
```diff
@@ -128,11 +128,11 @@ void automatedClaimOutOfScope() throws Exception {
 
     var request = MasTestData.getMasAutomatedClaimRequest();
     var responseEntity = post("/v2/automatedClaim", request, MasResponse.class);
-    assertEquals(HttpStatus.OK, responseEntity.getStatusCode());
-    assertTrue(offrampCalled.get());
-    assertTrue(completeCalled.get());
+    assertEquals(HttpStatus.UNPROCESSABLE_ENTITY, responseEntity.getStatusCode());
+    //    assertTrue(offrampCalled.get());
```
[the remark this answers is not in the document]
anthonyvelazquez: Fixed

controller/src/main/java/gov/va/vro/controller/MasController.java
```diff
@@ -44,12 +44,25 @@ public ResponseEntity<MasResponse> automatedClaim(MasAutomatedClaimRequest reque
     if (!hasValidClaimId(request)) {
       throw new BipException(
           HttpStatus.BAD_REQUEST, "The request does not have a valid BenefitClaimId.");
+    } else if (!request.canProcess()) {
+      throw new BipException(HttpStatus.UNPROCESSABLE_ENTITY, "The request cannot be processed.");
     }
+
+    if (!validDisabilityActionTypes.contains(
```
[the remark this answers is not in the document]
anthonyvelazquez: Updated the ticket requirements:
`“disabilityActionType” is “NEW” but “veteranFlashIds” does not include “Agent Orange Exposure Verified” or “266”` will still offramp

au2 (changes requested):
@anthonyvelazquez This PR has a lot of problems. I suggest you restart and  follow this path

1) Separate getOffRampReason into two methods. 2nd will have (payload.isPresumptive() != null && !payload.isPresumptive()) piece and the 1st will have all the other logic.
2) Separate processIncomingClaim into three methods. Piece before getOffRampReason should be called as is always  as a new method from controller.  The 2nd should use the 1st method of getOffRampReason methods and just return the message without offramp. The 3rd should use the 2nd method of getOfframpReason methods and work similar to current hetOfframpReason.
3) In the controller call 2nd of processIncomingClaim method  and raise 422 with the message. I would not use BipException. ClaimProcessingException is better I would say.
4) In the controller use the 3rd of processIncomingClaim method  as we are currently using processIncomingClaim.

Let me know if this is not clear and we can talk in the morning.

service/provider/src/main/java/gov/va/vro/service/provider/bip/service/BipClaimService.java
```diff
@@ -66,6 +67,46 @@ public boolean hasAnchors(long claimId) {
       return false;
     }
 
+    var contentions = bipApiService.getClaimContentions(claimId);
+    if (contentions == null) {
+      log.info("Claim with claim Id {} does not have contentions.", claimId);
+      return false;
+    }
```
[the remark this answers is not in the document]
anthonyvelazquez: Added

app/src/test/java/gov/va/vro/service/BipClaimServiceTest.java
```diff
@@ -80,7 +79,7 @@ void hasAnchors() throws BipException {
                 createContention(List.of(claimProps.getSpecialIssue1(), "OTHER"))));
 
     BipClaimService claimService = new BipClaimService(claimProps, bipApiService, null, null);
-    assertTrue(claimService.hasAnchors(bipClaimId));
+    assertFalse(claimService.hasAnchors(bipClaimId));
```
hayleymdawson, line 82: Did you mean to leave this in "hasAnchors"? or should there be two methods one for false and one for true? I am mildly confused. But it may just need a name change.....

app/src/test/java/gov/va/vro/controller/MasControllerTest.java
```diff
@@ -152,22 +149,10 @@ void automatedClaimInScope() throws Exception {
 
     var request = MasTestData.getMasAutomatedClaimRequest(567, "7101", "999");
     var responseEntity = post("/v2/automatedClaim", request, MasResponse.class);
-    assertEquals(HttpStatus.OK, responseEntity.getStatusCode());
+    assertEquals(HttpStatus.UNPROCESSABLE_ENTITY, responseEntity.getStatusCode());
     mockMasOfframpEndpoint.expectedMessageCount(1);
 
     verifyClaimPersisted(request);
```
hayleymdawson, line 155: I noticed the test above you return and dont verify persisted (assuming we dont persist?) but here we check. Wouldn't both 422 codes not persist (or both persist?)

hayleymdawson: this may be my misunderstanding of specification if that's not true. if not, maybe a comment?

app/src/test/java/gov/va/vro/service/MasProcessingServiceTest.java
```diff
@@ -28,7 +28,10 @@ void testClaimPersistence() {
     var diagnosticCode1 = "71";
     var request1 =
         MasTestData.getMasAutomatedClaimPayload(collectionId1, diagnosticCode1, claimId1);
-    masProcessingService.processIncomingClaim(request1);
+    //    masProcessingService.processIncomingClaim(request1);
```
hayleymdawson, line 31: I would remove the unused lines here.

anthonyvelazquez: Removed

app/src/end2endTest/java/gov/va/vro/end2end/VroV2Tests.java
```diff
@@ -483,9 +485,11 @@ void testAutomatedClaimOutOfScope() {
    * message, Slack message and removal of RDR1 special issue are verified.
    */
   @Test
+  @SneakyThrows
   void testAutomatedClaimMissingAnchor() {
     AutomatedClaimTestSpec spec = new AutomatedClaimTestSpec("20");
     spec.setPayloadPath("test-mas/claim-20-7101-noanchor.json");
+    spec.setExpectedStatusCode(HttpStatus.UNPROCESSABLE_ENTITY);
     spec.setExpectedMessage(
```
au2, line 493: This is not an off-ramp case anymore. Please remove the last line in this method (testAutomatedClaimOffRamp(spec)). That should have been failing.

au2 (changes requested):
Please see inline comments. We need and additional end to end test for not presumptive case which is the only one from the original that will off-ramp now but that should not necessarily hold this pr. You can do right after. Either way if you need help putting that test together please let me know.

app/src/test/java/gov/va/vro/service/BipClaimServiceTest.java
anthonyvelazquez: I need to add another test for that. I'll keep track of it and add it to a different PR where I will also add a new E2E test

au2 (approved):

PR #1300, opened 2023-03-17
Consolidate Dockerfiles

yoomlam:
## What was the problem?

Dockerfiles with the same content were copied in several locations.

## How does this fix it?

If `src/docker/Dockerfile` doesn't exist in a Gradle module, use the default `buildSrc/Dockerfile-jre`.

Also, use port 8080 for the `/actuator/health` endpoint; the port doesn't need to be exposed outside the container. The health check is used by docker-compose and K8s.

PR #1315, opened 2023-03-21
Add common Dockerfile for python container images

yoomlam:
Follow-on to  #1300

## What was the problem?

Dockerfiles with the same content were copied in several locations, making updates to Dockerfiles a pain.

## How does this fix it?

If `src/docker/Dockerfile` doesn't exist in a Gradle module, use the default `buildSrc/docker-python/Dockerfile`.

Also to have consistency, all files reside in the `/project` folder of the container, rather than `/home/docker` or some other folder.


yoomlam:
[PR: End-to-end test for v2](https://github.com/department-of-veterans-affairs/abd-vro/actions/runs/4482944657)

yoomlam:
These are the only Dockerfiles that don't use the default -- they require special setup:
```
./app/src/docker/Dockerfile
./postgres/src/docker/Dockerfile
./db-init/src/docker/Dockerfile
./.github/runner/Dockerfile
./console/src/docker/Dockerfile
```

The defaults are:
```
./buildSrc/docker-python/Dockerfile
./buildSrc/docker-jre/Dockerfile
```

PR #1321, opened 2023-03-23
MCP-2557 PDF Generator Table Break Logic

anthonyvelazquez:
## What was the problem?

Currently, the template has breakpoints after each table is defined regardless of its size. The design team asked if it would be possible to keep multiple tables on 1 page if they fit properly without any overflow.

- [MCP-2557](https://amida.atlassian.net/browse/MCP-2557)

## How does this fix it?[^1]

Remove all the manual breakpoints and then iterate over each table and compare the heights of them to see if they fit together or not. If they don't then we add the breakpoint

## How to test this PR
- Generate a PDF using the template `hypertension-v2-breakpoint`
- When generating the PDF, modify the amount of data in each table to see the different scenarios

### DIFF
--- a/service-python/pdfgenerator/src/lib/templates/hypertension-v2-breakpoint.html
+++ b/service-python/pdfgenerator/src/lib/templates/hypertension-v2-breakpoint.html
@@ -469,5 +469,26 @@
   <script src="https://cdn.jsdelivr.net/npm/bootstrap@5.0.2/dist/js/bootstrap.bundle.min.js"
     integrity="sha384-MrcW6ZMFYlzcLA8Nl+NtUVF0sA7MsXsP1UyJoMp4YLEuNSfAP+JcXn/tWtIaxVXM"
     crossorigin="anonymous"></script>
+  <script>
+    window.onload = function() {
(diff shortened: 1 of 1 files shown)


alan-amida (approved):
Did this one just get neglected by us, @anthonyvelazquez ?

anthonyvelazquez:
> Did this one just get neglected by us, @anthonyvelazquez ?

Not a requirement for iMVP but for after. Got the work done ahead of time but with our stable tests, I don't want to introduce something like this that might cause issues so close to go live. Will add more reviewers to get this merge after that

jbregstone (changes requested):
Logic looks good! I could not run the template as is because it does not have a corresponding template-variable file. I would switch the new logic with the old `hypertension-v2` if this is the way designers want to go. Also I would run a sample by the design people, the tables look a little too condensed to me but I'm not a designer. (example: Screenshot 2023-04-24 at 10 28 05 AM)
Maybe adding some padding under the table before the next one would work. Another consideration is having the header for the table consistently at top of page makes the document easier to skim, just my personal opinion.

jbregstone (approved):

PR #1320, opened 2023-03-22
Mcp 2550 - Adjust MAS BP annotation handling

wlin2019:
Using regex to handling MAS BP annotation handling

## What was the problem?
For some MAS BP annotation data, VRO cannot handle it gracefully, and create some crashing error messages. See MCP-2550.

Associated tickets or Slack threads:

- [MCP-000](https://amida.atlassian.net/browse/MCP-000)

## How does this fix it?[^1]

Using regular expression to handle the BP annotation data. It keeps the original approach to handle the missing BP reading value with default value (0). Consequently, I also adjusted Lighthouse service code to handle missing BP reading value to be consistent with MAS handling. See PR #1293.

## How to test this PR
- Step 1
- Step 2[^secrel]

[^1]: [Pull-Requests guidelines](https://github.com/department-of-veterans-affairs/abd-vro/wiki/Pull-Requests). If PR is significant, update [Current Software State](https://github.com/department-of-veterans-affairs/abd-vro/wiki/Current-Software-State) wiki page.
[^secrel]: To check if a PR will succeed in the SecRel workflow, [test PRs in the SecRel pipeline](https://github.com/department-of-veterans-affairs/abd-vro-internal/wiki/Secure-Release-process#to-test-prs-in-the-secrel-pipeline).

PR #1327, opened 2023-03-24
Mcp 2550 - Remove Exception.

wlin2019:
Removed the exception per PR #1320 comment

## What was the problem?

PR #1320 throws MasException when the blood pressure annotation data is invalid. Per Afsin, the MAS integration process in VRO cannot have exception. 

Please do not throw an exception. Just log and move on. Exceptions here will cause off-ramps after the next round off fixes. We do not want to off-ramp with one bad blood pressure data if other data is good. We already fixed on bug in FHIR (no data) not to throw exceptions. PLease remove exception throwing from the changes.

MCP-2550

## How does this fix it?[^1]

1. Remove the exception and log the error.
2. Filter out -/- (missing both systolic and diastolic blood pressure reading) to be consistent with LH BP reading.

## How to test this PR
- Step 1
- Step 2[^secrel]


au2 (approved):
I think the code could be better organized but leaving that to future

PR #1328, opened 2023-03-24
MCP-2574 Continue Processing when BIP CE is down

hayleymdawson:
## What was the problem?
When BIP CE API is down, previously, processing would just stop and fail. 

- [MCP-2574](https://amida.atlassian.net/browse/MCP-2574)

## How does this fix it?[
Now we notify slack, and continue processing successfully

## How to test this PR
Again, another doozy to test.

Set up your environment with mocks standing up, and whatever slack config you want to test. I use an end to end test config but override the application yml to have the non prod slack channel active.

Stand docker up, stop mock BIP CE container. *do not* take down BIP Claim API. Only CE.

Send MAS an automated processing claim for a collection ID setup for ordering an exam. Id 377 and 378 work.
You can use VROV2Tests.java for this, but you'll have to comment out this line in setup

    // restTemplate.delete(RECEIVED_FILES_URL + fileNumber); (line 288)

Then, run the test from there (or run it via another method).

The logs will show an error with BIP exception which will get handled, and a slack message with the appropriate info will get called. You'll also see the logs continue on and update the BIP Claim API.

Attached screenshot of what you get message wise in slack. 
Screen Shot 2023-03-24 at 11 54 00 PM


yoomlam (approved):
Did not test. It'd be great to have a unit test that simulates an exception -- for example, https://github.com/department-of-veterans-affairs/abd-vro/blob/757b33951e74bfb7c22169b3219205724d9691e7/domain-xample/workflows-xample/src/test/java/gov/va/vro/routes/xample/XampleRoutesTest.java#L139

au2 (changes requested):
1. Please address the conflicts. Looks like my changes yesterday conflicts. Let me know if you need help.
2. Not sure why we addressed only order exam. PDF is uploaded for RDF as well. That I separated yesterday BTW.
3. To that end I would address this at ENDPOINT_UPLOAD_PDF so it is handled for everything at the same place.

hayleymdawson:
@au2 understood. what i'll do is make it work the way i have it first (so you'll see a commit) and then spend some time seeing if I can do it the other way and ping you on slack when I've either managed it or hit a snag.

service/provider/src/main/java/gov/va/vro/service/provider/camel/MasIntegrationRoutes.java
```diff
@@ -203,8 +202,17 @@ private void configureMasProcessing() {
...
+        .doTry()
         .to(ENDPOINT_UPLOAD_PDF)
+        .endDoTry()
+        .doCatch(BipException.class)
+        // Mas Complete Processing code expects this to be the body of the message
+        .setBody(simple("${exchangeProperty.payload}"))
+        .wireTap(ENDPOINT_NOTIFY_AUDIT) // Send error notification to slack
+        .onPrepare(
+            slackEventProcessor(orderExamRouteId, "PDF upload failed after exam order requested."))
+        .endDoCatch()
```
au2, line 215: Based on this "Instead of end() you can use endDoTry() to end and return back to the try …​ catch scope." from camel documentation you need end I think. Exam order end to end tests are failing because processing stops right after the pdf is uploaded.  I do not think you need ".setBody(simple("${exchangeProperty.payload}"))" but I have not tested that yet. I will start trying with end on my end.

au2 (changes requested):

service/provider/src/main/java/gov/va/vro/service/provider/camel/MasIntegrationRoutes.java
hayleymdawson: The set body is for the MAS complete route that needs it. it's not set to the body at that point. I forget what is, but it's not that. It definitely fails without it.

As far as the end vs endDoTry() that's....confusing. It was my understanding endDoTry was only the end ofthe try block and end() ends the route itself.

hayleymdawson: that said if end() not endDoTry is right the lighthouse code may also be wrong i'll switch back to this branch in a moment.

au2 (approved):

PR #1351, opened 2023-03-28
Prep for bitnami base images

yoomlam:
We've been talking about using bitnami images to reduce SecRel alerts. Update our Dockerfile setup to follow their conventions in preparation for this migration. The original choices were made arbitrarily.

- Bitnami images use the convention of using a `/app` workdir.
- Replace creation of user `docker` with something more unique: I chose `tron` (`nonroot` was considered but it has `root` in it. `tron` has the same letters and is short).
- Also rename `entrypoint-default.sh` to `entrypoint-wrapper.sh` b/c it's more accurate in that containers should always run `entrypoint-wrapper.sh`, which will call `entrypoint.sh` if it exists.


yoomlam:
[PR: End-to-end test for v2](https://github.com/department-of-veterans-affairs/abd-vro/actions/runs/4538573083)

Issue #1312, opened 2023-03-20
Offramp claims in given error handling scenarios


dianagriffin:
### User Story

As a VRO stakeholder, I want to have confidence that when claims cannot complete processing in VRO due to errors, the claims are appropriately offramped so that they will be processed manually by claim adjudicators. 

**Error scenarios**

1. VRO identifies that a `NEW` claim is not presumptive (ie. does not have the `Agent Orange Exposure Verified` flash) - _note this is technically not an error, and should still return 200 from /automatedClaims_
   - Logs should record offramp reason: `newClaimMissingFlash266`
2. VRO cannot successfully retrieve annotation data via the MAS endpoint 
   - Logs should record offramp reason: `annotationDataRequestFailed`
4. VRO cannot successfully request an exam order via the MAS endpoint
   - Logs should record offramp reason: `examOrderFailed`
6. BIP Claim Evidence API is down _and_ the claim has sufficient evidence to be marked `RFD` (ie. did not request an exam)
   - Logs should record offramp reason: `docUploadFailed`

### Acceptance Criteria

1. For each of the error scenarios listed above, offramp the claim by setting it to the state described in #1286 
2. For each of the error scenarios listed above, log the error and trigger a Slack notification to #benefits-vro-alerts that includes in the message field:
   - the Claim ID 
   - the offrampReason

**Notes about work**
- https://github.com/department-of-veterans-affairs/abd-vro/wiki/iMVP-claim-and-contention-update-scenarios
- https://lighthouseva.slack.com/files/U03GVR68Q6L/F04UB2LNA86/mas-mcp_automated_claim_contract.docx
- [Error scenarios spreadsheet](https://dvagov.sharepoint.com/:x:/r/sites/vaabdvro/Shared%20Documents/VRO%20-%20Virtual%20Regional%20Office/technical%20reference/error_scenarios.xlsx?d=w38a7304fd531410d892dc4733fdb09fa&csf=1&web=1&e=Agr1Bl)

PR #1353, opened 2023-03-28
MCP-2581 and MCP-2582 Offramp in Error Scenarios

hayleymdawson:
## What was the problem?
Claims were stuck in the following two error cases
1. MAS Failed to order the exam
2. We did not order an exam, and failed to upload the PDF due to BIP CE api error

- [MCP-2581](https://amida.atlassian.net/browse/MCP-2581)
- [MCP-2582](https://amida.atlassian.net/browse/MCP-2582)
- https://github.com/department-of-veterans-affairs/abd-vro/issues/1312 (items #3 and #4)
- 
## How does this fix it?
Claims are now offramped in the above two cases. (See MCP tickets or github issue)

## How to test this PR
Run End to End Tests

I have left some review comments on this one, because there's some things that other devs probably care about.


service/db/src/main/java/gov/va/vro/service/db/SaveToDbServiceImpl.java
```diff
@@ -157,6 +157,13 @@ public void setOffRampReason(Claim claimWithOffRamp) {
     Optional<ClaimSubmissionEntity> claimSubmission =
         claimSubmissionRepository.findFirstByReferenceIdAndIdTypeOrderByCreatedAtDesc(
             claimWithOffRamp.getCollectionId(), claimWithOffRamp.getIdType());
+    if (claimSubmission.isEmpty()) {
+      log.info(
```
hayleymdawson, line 161: Before Yoom asks, yes this is info not error. There are cases we offramp and this doesn't exist yet. When i collapsed the MasIntegration routes for offramping I found that out.

service/provider/src/main/java/gov/va/vro/service/provider/camel/MasIntegrationProcessors.java
```diff
@@ -145,12 +146,19 @@ public static Processor auditProcessor(
...
+  public static Processor completionProcessor(
+      BipClaimService bipClaimService, MasProcessingService masProcessingService) {
     return exchange -> {
       MasProcessingObject payload = exchange.getIn().getBody(MasProcessingObject.class);
       MasCamelStage origin = payload.getOrigin();
       Boolean sufficient = exchange.getProperty("sufficientForFastTracking", Boolean.class);
-      MasCompletionStatus completionStatus = MasCompletionStatus.of(origin, sufficient);
+      String offRampError = exchange.getProperty("offRampError", String.class);
+      // Update our database with offramp reason.
+      if (offRampError != null) {
```
hayleymdawson, line 157: OffRampError is our key here on properties to know if we're offramping or not.

service/provider/src/main/java/gov/va/vro/service/provider/camel/MasIntegrationRoutes.java
```diff
@@ -178,52 +180,72 @@ private void configureMasProcessing() {
         .to(END_POINT_RFD)
         .otherwise()
         // Off ramp if the Sufficient For Fast Tracking is null
-        .to(ENDPOINT_ACCESS_ERR)
+        .setProperty("offRampError", constant("Sufficiency cannot be determined."))
+        .setProperty("sourceRoute", constant("assessorError"))
+        .log("Assessor Error. Off-ramping claim")
```
hayleymdawson, line 185: ENDPOINT_ACCESS_ERR was very close to what I needed from ENDPOINT_OFFRAMP_ERROR and in the interest of not causing too many camel routes and making the code harder, it's been collapsed and what wasn't contained moved here.

service/provider/src/main/java/gov/va/vro/service/provider/camel/MasIntegrationRoutes.java
```diff
@@ -178,52 +180,72 @@ private void configureMasProcessing() {
...
+        .doTry()
         .to(ENDPOINT_UPLOAD_PDF)
-        // Check and update claim
+        .doCatch(BipException.class)
+        // Completion code needs the MasProcessingObject as the body.
+        .setBody(simple("${exchangeProperty.payload}"))
+        .setProperty("offRampError", constant(pdfFailError))
+        .setProperty("sourceRoute", constant(rfdRouteId))
+        .to(ENDPOINT_OFFRAMP_ERROR)
+        .stop()
```
hayleymdawson, line 206: stop is now explicit to ensure that the last masComplete after the catch doesn't run.

hayleymdawson: This should also allow future ENDPOINT_MAS_COMPLETE error handling to stand independently of this try catch as needed.

service/provider/src/main/java/gov/va/vro/service/provider/camel/MasIntegrationRoutes.java
```diff
@@ -178,52 +180,72 @@ private void configureMasProcessing() {
...
-        .routeId(assessorErrorRouteId)
-        .log("Assessor Error. Off-ramping claim")
-        .process(masAccessErrProcessor)
-        .wireTap(ENDPOINT_NOTIFY_AUDIT)
-        .onPrepare(slackEventProcessor(assessorErrorRouteId, "Sufficiency cannot be determined."))
-        .to(ENDPOINT_MAS_COMPLETE);
+    // Wiretap does NOT let camel work as expected when placed directly inside doCatch()
+    // Thus it is broken out here, in the interest of letting normal flow/control happen.
+    from(ENDPOINT_OFFRAMP_ERROR)
+        .wireTap(ENDPOINT_NOTIFY_AUDIT) // Send error notification to slack
```
hayleymdawson, line 245: I realize I often call stop() explicitly rather than relying on this strange behavior. I think that would be easier for devs to follow in the future. And if camel ever *makes that work* the way that it seems it should.

service/provider/src/main/java/gov/va/vro/service/provider/camel/MasIntegrationRoutes.java
```diff
@@ -178,52 +180,72 @@ private void configureMasProcessing() {
...
+        .doCatch(MasException.class)
+        // Body is still the Mas Processing object.
+        .setProperty("sourceRoute", constant(orderExamRouteId))
+        .setProperty("offRampError", constant(orderFailMessage))
+        .to(ENDPOINT_OFFRAMP_ERROR)
+        .stop() // Offramp and don't continue processing
         .doCatch(BipException.class)
         // Mas Complete Processing code expects this to be the body of the message
         .setBody(simple("${exchangeProperty.payload}"))
+        // Wiretap will cause no code to execute after the end of the try. Intentional here.
```
hayleymdawson, line 236: I wanted to not rely on the wiretap side-effect behavior here which stops processing, but i'd have to make an entirely new route just to contain the wiretap in a *non* offramp scenario and somehow pass the onprepare the right variables differently. 
Due to time I commented that I am relying on the strange behavior here. (This case execution did not change in this ticket, i just made endDoCatch() end which is more correct in this case even with wiretap involved).

alan-amida (approved):
I personally don't see anything objectionable here, but I also don't know what's in the businesses' head like some others do. The code looks sound. I'll let others comment on design.

Also this is the thing Hayley is talking about with respect to Camel wiretap causing `doCatch()`s to fail:
https://access.redhat.com/solutions/6749521

You need an account to view the whole thing, but otherwise, it's there, it's known, and RedHat are the only ones so far who have an article about it. Good find, @hayleymdawson

anthonyvelazquez (approved):

service/provider/src/main/java/gov/va/vro/service/provider/camel/MasIntegrationRoutes.java
hdaws: Not quite but I will sync with you in a huddle to explain. :) easier i think

Issue #1341, opened 2023-03-27
Update VRO workflow to use BGS service to add automation outcome notes to claims processed by VRO


dianagriffin:
### **User Story**
As a RVSR, I want to use the claim notes to understand what was completed by automation so I know how to proceed with claim processing.

As a member of PA&I, I want to use the claim notes to report off of the effectiveness/deferral rate of the program overall.

### **Acceptance Criteria**
VRO adds a note(s) to all single issue hypertension claims (new and increase) that VRO assesses for automation.
- If the claim and contention is set to RFD, add a:
  >**Veteran-level note**: "Hypertension automated review summary document uploaded MM/DD/YYYY" (datetime the ARSD was uploaded to the eFolder)

  >**Claim-level note**: "Claim status updated to RFD via automation"

  >**Claim-level note**: "Please note, the automated review summary document (ARSD) has been completed in full. One or more contentions were determined to be RFD by current rating criteria. Necessary exams were requested or drafted."

- If an exam is requested, add a:
  >**Veteran-level note**: "Hypertension automated review summary document uploaded MM/DD/YYYY" (datetime the ARSD was uploaded to the eFolder)

  >**Claim-level note**: "Please note, the automated review summary document has been completed in full with the necessary exam(s) requested or drafted based on automation eligible contention(s). This claim/contention has been placed in the Open Status Development Life Cycle for further review by a claims processor."

- If the claim is OFFRAMPED with either offrampReason `newClaimMissingFlash266` (we've assessed the claim, and we can't confirm presumptive service connection) or offrampReason `insufficientHealthDataToOrderExam`, add:
  >**Claim-level note**: "Automation-eligible contentions reviewed, but no action taken based on existing evidence. This claim has been removed from the automation program and placed in Open status for development by a claims processor."

- If claim is offramped due to ARSD upload failure, add a: 
  >**Claim-level note**: "Claim was offramped from VBA automation platform when ARSD could not be uploaded to eFolder."

### **Assumptions**
All other notes related to eligibility, automation exam request issues, and HDR records will be handled by MAS, because it occurs before the claim is handed off to VRO or it's related to exam requests.

### **Not included in this work**
- Connectivity to BGS and creation of VRO Ruby microservice to make BGS calls (#1289)
- Retry logic in case initial call to BGS fails (#1343)
- Error handling in case BGS service is unavailable or fails (#1342)

### **Notes about work**
[...]

PR #1357, opened 2023-03-28
DB: add uploadedAt to EvidenceSummaryDocumentEntity

yoomlam:
## What was the problem?

To add a BGS note:
> "Hypertension automated review summary document uploaded MM/DD/YYYY"

we need the date but it's not currently stored in the DB.

- #1341

## How does this fix it?

Add `uploadedAt` to `EvidenceSummaryDocumentEntity`

(I didn't see an existing unit test that updates and checks the `folderId` so I can add a check for `uploadedAt`.)


hdaws (approved):
Does what it says on the tin
Beware though any exposure of this date outside vro is really in the format they want. We definitely have whacky inputs we have to custom parse for other integration.

hdaws:
Cc @alan-amida in case database documents need updated

PR #1371, opened 2023-03-29
Mcp 2579 not presumptive

hayleymdawson:
## What was the problem?
Not presumptive claims returned 200 and a message, but did not slack. 
Code also passed around a lot of exchange properties, tried to reduce those by leveraging the offramp reason in the claim payload.

- [MCP-2579](https://amida.atlassian.net/browse/MCP-2579)

## How does this fix it?
Slacks the message by going through the mas completion route to offramp and slack

## How to test this PR
Run e2e Tests


service/provider/src/main/java/gov/va/vro/service/provider/mas/MasProcessingObject.java
```diff
@@ -72,4 +72,8 @@ public String getDisplayName() {
   public String getClaimSubmissionDateTime() {
     return claimPayload.getClaimDetail().getClaimSubmissionDateTime();
   }
+
+  public String getOffRampReason() {
+    return claimPayload.getOffRampReason();
```
au2, line 77: I think we will need to make this an array since there might be multiple of these under rare circumstances. When and error and offramp adds up or error from both mas and lighhouse. Not this PR.

service/provider/src/main/java/gov/va/vro/service/provider/mas/service/MasProcessingService.java
```diff
@@ -77,12 +77,10 @@ public String processIncomingClaimPresumptiveOffRampClaimCheck(MasAutomatedClaim
     var offRampReasonOptional = getOffRampReasonPresumptiveCheck(payload);
     if (offRampReasonOptional.isPresent()) {
       var offRampReason = offRampReasonOptional.get();
-      Claim claim = toClaim(payload);
       payload.setOffRampReason(offRampReason);
-      claim.setOffRampReason(offRampReason);
-      saveToDbService.setOffRampReason(claim);
-      offRampClaim(payload, offRampReason);
-      return offRampReason;
+      MasProcessingObject mpo = new MasProcessingObject(payload, MasCamelStage.START_COMPLETE);
```
au2, line 81: We should not need this MasCamelStage.START_COMPLETE anymore. The whole reason was to identify the controller offramps.

au2 (approved):
I put two comments fro future.

PR #1393, opened 2023-04-03
MCP-2652 and MCP-2653 Lighthouse 500 and Timeouts

hayleymdawson:
## What was the problem?
See tickets, LH error cases were not working as intended/expected 

- [MCP-2652](https://amida.atlassian.net/browse/MCP-2652)
- [MCP-2653](https://amida.atlassian.net/browse/MCP-2653)

## How does this fix it?
SVC lighthouse MUST give an error message for the VRO app to catch it. This is now ensured even if exception has no message when SVC catches it.

VRO app code has altered to accommodate both paths with appropriate retry.

## How to test this PR
Run end to tend tests

Of note:
1. This makes E2E testing slightly over 4 minutes longer to test retry timeout twice in a row. You will have to wait.
2. The app logs will give a warning about early replies, or replies with no correlation ID. this happens because mock-lighthouse replies *after* the timeout, thus causing camel to discard the message with a warning. These messages are quite large, but apparently just part of camel


au2 (approved):
Test are passing so approving. Will add other tests later.

Issue #1373, opened 2023-03-29
De-duplicate BP readings between LH and HDR


amylai-nava:
### User Story
As a VA stakeholder, I want VRO to de-duplicate blood pressure (BP) readings between Lighthouse and HDR so that we are accurately assessing whether there is sufficient data to fast-track the claim.

### Context
With the most recent MAS release, the HDR data is now expected to include all the same BP readings as the LH data, with the implication that the VRO logic that is dependent on # of BP readings found in the data will return false positives based on duplicate readings. To address this risk we want to prefer the BP readings from the HDR document; however, we still want to query the LH data for two reasons:

1. We still need LH data for condition data
2. De-duping is an interim step that allows us to validate that all relevant BP readings are in fact represented in the HDR data. We believe this to be true, but the HDR is a new feature and including de-duping in VRO will allow us to validate with production data for a few months before we decide whether or not to take the step of removing our query for LH patient health data for Hypertension claims.

### Acceptance Criteria
- BP readings between Lighthouse and HDR are considered duplicate if they have: 
  - the same systolic reading
  - the same diastolic reading
  - the same observation date
- If a duplicate BP reading is identified, use the HDR result as the single reading. It should be used to:
  - assess sufficiency on a claim
  - be displayed in the evidence summary PDF
- No de-duplication is needed among BP readings within HDR or within Lighthouse data (example: if there are 2 BP readings within HDR that have the same systolic reading, diastolic reading, and observation date, count it as 2 BP readings)

### Additional Notes
- BP readings from HDR are received from MAS as an annotation and have a docTypeDescription of `VAMC Other Output Reports`

---

**Open questions**
- Do we need to check for duplicate BP readings against non-HDR documents? For example, against VA records that were uploaded by the Veteran? How is MAS handling this? **3/30 update: MAS is not de-duplicating between HDR and VA records uploaded by the Veteran**
- Potential edge case: If there's 2 BP readings in HDR and 1 BP reading in Lighthouse that have the same readings and observation date, do we consider it 1 reading or 2? **Going by my assumption that we should not de-duplicate within HDR or within Lighthouse data, that would leave us with 2 readings.**

**Future enhancements**
- Matching on the time the observation was taken - currently, the observation date's we receive from MAS do not include the time of the observation, just the date.
- Matching on organization - I'm assuming this information is available in the HDR JSON but would need to verify. 

**Related slack threads**
[...]

PR #1415, opened 2023-04-04
MCP-2318: Remove deduplicate BP readings

jbregstone:
## What was the problem?

HDR and LH may have duplicate BP readings.

- [MCP-2318](https://amida.atlassian.net/browse/MCP-2318)

## How does this fix it?[^1]

Add logic in #1373 
## How to test this PR
- Step 1
- Step 2[^secrel]


service-python/tests/assessclaim/dc7101/bp_sufficiency_test.py
```diff
@@ -814,7 +1002,7 @@
                                               'value': 200}}],
                  'twoYearsBpCount': 2}
         ),
-        (
+        (  # remove LH duplicate but keep MAS data intact
```
jbregstone, line 1005: test case for deduplication

service-python/assessclaimdc7101/src/lib/bp_calculator.py
greene858, line 1: If a reading is **not** `reading["dataSource"] == "LH" and bp_comp["organization"] == "VAMC Other Output Reports"` does it need to go through the inner for loop? Also, if it is a duplicate, can we remove the duplicate from the list of readings to be checked - we already know it's a duplicate. This may be more performance optimization than is called for now :-), but could matter if we get duplicates at the end of very long lists.

service-python/tests/assessclaim/dc7101/bp_sufficiency_test.py
greene858: Nice to have: more test cases to validate that only "same diastolic
     value, systolic value and date" are removed as duplicate

service-python/assessclaimdc7101/src/lib/bp_calculator.py
jbregstone: I can move `reading["dataSource"] == "LH"` to the outer for loop, but I don't think I can do the same for `bp_comp["organization"] == "VAMC Other Output Reports"` because it references the items from the inner for loop.

service-python/assessclaimdc7101/src/lib/bp_calculator.py
```diff
@@ -21,6 +21,30 @@ def sort_bp(bp_readings):
...
+    """
+    Return bp readings with any lighthouse duplicates removed. A duplicate is identified by having the same diastolic
+    value, systolic value and date. HDR data has organization set to VAMC Other Output Reports.
+    :param bp_readings: full list of BP readings
+    :return: deduplicated list
+    """
+    deduplicated_readings = []
+    for reading in bp_readings:
+        duplicate = False
+        if reading["dataSource"] == "LH":
```
alan-amida, line 34: We talked about the possibility of using a `set` to de-duplicate this. I might still like to see this implemented to mitigate away the loop-within-a-loop later, but that can be a performance tweak for later.

alan-amida (approved):
Changes look good to me. 👍

service-python/assessclaimdc7101/src/lib/bp_calculator.py
au2: i agree, this needs to use set or dictionary. Although the data sizes are small enough that performance gain will not be noticeable.  I would do it anyway, just better coding.

service-python/assessclaimdc7101/src/lib/bp_calculator.py
```diff
@@ -21,6 +21,30 @@ def sort_bp(bp_readings):
...
+            for bp_comp in bp_readings:
+                if reading["diastolic"]["value"] == bp_comp["diastolic"]["value"] \
+                    and reading["systolic"]["value"] == bp_comp["systolic"]["value"] \
+                        and reading["date"] == bp_comp["date"] \
+                        and bp_comp["organization"] == "VAMC Other Output Reports":
+                    duplicate = True
+                    bp_readings.remove(reading)
```
au2, line 41: removing from bp_readings while inside a for loop for bp_readings (and double on that) is a bad practice that hurt me once or twice. I do not know what happens in python actually. I suggest you reorganize.

service-python/assessclaimdc7101/src/lib/bp_calculator.py
```diff
@@ -73,6 +98,6 @@ def bp_reader(request_body):
               "twoYearsBpCount": len(bp_readings_in_past_two_years),
               "oneYearBpCount": len(bp_reading_in_past_year),
               "twoYearsElevatedBpCount": len(elevated_bp_in_past_two_years),
-              "totalBpCount": len(request_body["evidence"]["bp_readings"])}
+              "totalBpCount": len(deduplicated_bp_readings)}
 
     return result
```
au2, line 103: I would add some info on what is deduplicated in evidence summary so we can see it in logs and database.

au2 (dismissed):
Where exactly is the unit test for this?

service-python/assessclaimdc7101/src/lib/bp_calculator.py
jbregstone: It is not a set though. In the requirements, HDR data should not be removed even if there are duplicates within it.

jbregstone: From requirements:

Potential edge case: If there's 2 BP readings in HDR and 1 BP reading in Lighthouse that have the same readings and observation date, do we consider it 1 reading or 2? Going by my assumption that we should not de-duplicate within HDR or within Lighthouse data, that would leave us with 2 readings.

service-python/assessclaimdc7101/src/lib/bp_calculator.py
jbregstone: I added "lighthouseDuplicateBpCount" to evidence summary

service-python/assessclaimdc7101/src/lib/bp_calculator.py
jbregstone: > Where exactly is the unit test for this?

I added some unit tests in most recent commits

jbregstone:
> Where exactly is the unit test for this?

I added some unit tests in most recent commits

service-python/assessclaimdc7101/src/lib/bp_calculator.py
jbregstone: I will removed this line

service-python/assessclaimdc7101/src/lib/bp_calculator.py
anthonyvelazquez: Based on the requirements I say this could work. Loop through one and anything not `LH` or `VAMC Other Output Reports` is unique so all thats left is just comparing the (diastolic + systolic + date) in a set for the remaining uniqueness.
```
deduplicated_readings = []
unique_readings = set()
for reading in bp_readings:
    if reading["dataSource"] != "LH":
        deduplicated_readings.append(reading)
        continue
    if reading["organization"] != "VAMC Other Output Reports":
        deduplicated_readings.append(reading)
        continue
    reading_key = (reading["diastolic"]["value"], reading["systolic"]["value"], reading["date"])
    if reading_key not in unique_readings:
        unique_readings.add(reading_key)
        deduplicated_readings.append(reading)
return deduplicated_readings
```

jbregstone: This is clever but I do not think it will work because all MAS data will be caught by the first if statement and so it will not be added to `unique readings`. I do not think it would compare MAS data to LH data. It's possible this implementation could work with some more complicated conditions, I am not sure.

anthonyvelazquez: Is HDR the same as MAS or a 3rd data source?

jbregstone: We are getting HDR data from the MAS API. It is identified by having "VAMC Other Output Reports" as the organization.

alan-amida (approved):

anthonyvelazquez (approved):

PR #1424, opened 2023-04-05
Add message/reason to some offramp reason messages.

wlin2019:
Address the message inconsistency in two off-ramp slack messages.

## What was the problem?
1. The slack messages for off-ramp are inconsistent.
2. Slack message not having claim ID.
3. Inconsistency in the messages for other events.

MCP-2679, MCP-2700, MCP-2734

## How does this fix it?[^1]
Modified the messages to make them consistent, and add claim ID to the slack message.

## How to test this PR
- Step 1
- Step 2[^secrel]


service/provider/src/main/java/gov/va/vro/service/provider/camel/MasIntegrationRoutes.java
```diff
@@ -112,10 +112,12 @@ public class MasIntegrationRoutes extends RouteBuilder {
...
   // Possible OffRamp Reasons
-  public static final String SUFFICIENCY_UNDETERMINED = "Sufficiency cannot be determined.";
+  public static final String SUFFICIENCY_UNDETERMINED =
+      "Sufficiency cannot be determined. Reason: assessorError.";
   public static final String PDF_UPLOAD_FAILED_AFTER_ORDER_EXAM = "docUploadFailed";
   public static final String EXAM_ORDER_FAILED = "examOrderFailed";
-  public static final String NEW_NOT_PRESUMPTIVE = "newClaimMissingFlash266";
+  public static final String NEW_NOT_PRESUMPTIVE =
+      "New claim cannot be determined to be presumptive. Reason: newClaimMissingFlash266.";
   public static final String ANNOTATIONS_FAILED = "annotationDataRequestFailed";
```
au2, line 121: Please leave these as keys and change SUFFICIENCY_UNDETERMINED to be a key to. I believe others are using these to make decisions. At the place where the actual slack messages are generated, add the narrative based on key.

au2 (changes requested):
We need to address all the slack messages similarly.

service/provider/src/main/java/gov/va/vro/service/provider/camel/MasIntegrationRoutes.java
wlin2019: Changed the code to use an object EventReason for reason code and narrative. See comments in MCP-2679.

helm/charts/domain-rrd/templates/deployment.yaml
```diff
@@ -41,23 +41,3 @@ spec:
             {{- include "vro.redisClient.envVars" . | nindent 12 }}
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
```
hayleymdawson, line 43: Why are helm files changing in this PR or is this an unresolved conflict?

wlin2019: I think that change is come from develop branch. I just merge develop into my branch again this morning.

ianwolson (approved):

hayleymdawson (approved):
Current changes look correct.

au2 (changes requested):
We also send slacks if we hit errors during BIP updates. Please put that in here as well. Take a look MasIntegrationProcessors.completionProcessor.

au2 (commented):
@hayleymdawson @wlin2019 I looked at the code and PDF_UPLOAD_FAILED_AFTER_ORDER_EXAM is only being used when we fail in RFD case. So that error narrative appears to be wrong. I do not think PDF_UPLOAD_FAILED_AFTER_ORDER_EXAM is being used when pdf load indeed fails in exam route. That has the slack message hardcoded it looks like. I know that is not an offramp but it would be good to be consistent,

au2:
Right. The message and the constant name for that matter is wrong since it is only used when pdf generation fails in RFD. For Pdf generation failures from exam routes we do not use even a code I do not think. We should to be consistent. @wlin2019 @hayleymdawson

au2:
End 2 end tests only check if a slack is sent currently;  after this pr we can check for codes as well. That was my intention originally.

wlin2019:
> Right. The message and the constant name for that matter is wrong since it is only used when pdf generation fails in RFD. For Pdf generation failures from exam routes we do not use even a code I do not think. We should to be consistent. @wlin2019 @hayleymdawson

I added two cases (events) in EventReason, PDF_UPLOAD_FAILED_AFTER_RFD and BIP_UPDATE_FAILED.

service/provider/src/main/java/gov/va/vro/service/provider/camel/MasIntegrationProcessors.java
```diff
@@ -228,8 +234,14 @@ public static Processor slackEventPropertyProcessor(
 
   public static String getSlackMessage(MasProcessingObject mpo, String originalMessage) {
     String msg = originalMessage;
+    EventReason reason = EventReason.getEventReason(originalMessage.trim());
```
yoomlam, line 237: This seems hacky, i.e., using an _arbitrary_ message string to look up an enum. Shouldn't the reason (`sufficientForFastTracking`, `claimPayload.offRampReason`) be available as part of the `mpo` object, so you can look up the enum based on the contents of the `mpo` object?

service/provider/src/main/java/gov/va/vro/service/provider/bgs/service/BgsClaimNotes.java
```diff
@@ -1,13 +1,13 @@
...
-import static gov.va.vro.service.provider.camel.MasIntegrationRoutes.PDF_UPLOAD_FAILED_AFTER_ORDER_EXAM;
-import static gov.va.vro.service.provider.camel.MasIntegrationRoutes.SUFFICIENCY_UNDETERMINED;
-
 import java.util.Map;
 
 public class BgsClaimNotes {
 
+  public static final String SUFFICIENCY_UNDETERMINED = "Sufficiency cannot be determined.";
+  public static final String PDF_UPLOAD_FAILED_AFTER_ORDER_EXAM = "docUploadFailed";
+  public static final String NEW_NOT_PRESUMPTIVE = "newClaimMissingFlash266";
```
yoomlam, line 9: Should these be part of the `EventReason` enum? The concepts represented by these String constants seem to overlap with those in `EventReason`.

yoomlam (changes requested):

shared/domain/src/main/java/gov/va/vro/model/event/EventReason.java
```diff
@@ -0,0 +1,47 @@
...
+  PDF_UPLOAD_FAILED_AFTER_ORDER_EXAM(
+      "docUploadFailed", "PDF upload failed after exam order requested."),
+  PDF_UPLOAD_FAILED_AFTER_RFD("docUploadFailedRfd", "Failed to upload PDF file."),
+  EXAM_ORDER_FAILED("examOrderFailed", "Failed to order exam."),
+  ANNOTATIONS_FAILED("annotationDataRequestFailed", "Failed to get annotation data."),
+  BIP_UPDATE_FAILED("bipUpdateFailed", "BIP update failed.");
+
+  private static int MAX_CODE_LENGTH = 50;
+
+  private String code;
```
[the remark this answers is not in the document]
wlin2019: This PR is a work in progress. The original Jira ticket only mentioned one message. The discussions in the PR led to this solution.

As the magic strings (codes) were used in the code for messages, I guess, @au2 suggested to add a narrative and kept the code. It would avoid too much impact on the existing code while we make the event message handling a bit more consistent.

Since the code values are extracted from the current codes, I add MAX_CODE_LENGTH to avoid long string search.

If we know all the events, we can redefine the enum here and just use a simple enum without code and narrative later. At present, using code and narrative will reduce the possibility to break other parts in the existing code.

shared/domain/src/main/java/gov/va/vro/model/event/EventReason.java
```diff
@@ -0,0 +1,47 @@
...
+  NEW_NOT_PRESUMPTIVE(
+      "newClaimMissingFlash266", "New claim cannot be determined to be presumptive."),
+  PDF_UPLOAD_FAILED_AFTER_ORDER_EXAM(
+      "docUploadFailed", "PDF upload failed after exam order requested."),
+  PDF_UPLOAD_FAILED_AFTER_RFD("docUploadFailedRfd", "Failed to upload PDF file."),
+  EXAM_ORDER_FAILED("examOrderFailed", "Failed to order exam."),
+  ANNOTATIONS_FAILED("annotationDataRequestFailed", "Failed to get annotation data."),
+  BIP_UPDATE_FAILED("bipUpdateFailed", "BIP update failed.");
+
+  private static int MAX_CODE_LENGTH = 50;
```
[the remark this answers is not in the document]
wlin2019: Good catch! See my comment above. We can remove it later when we can have better code defined.

service/provider/src/main/java/gov/va/vro/service/provider/camel/MasIntegrationProcessors.java
wlin2019: Yes. It is not architected to have a consistent way of handling events and messages. At this stage, this is a solution. The MAX_CODE_LENGTH = 50 is for this to avoid long string comparison.

service/provider/src/main/java/gov/va/vro/service/provider/bgs/service/BgsClaimNotes.java
wlin2019: Yes. I tried not to remove the code in this file before asking you to review it. I'll use EventReason to replace these constants.

wlin2019: I replaced those constants with enum items.

wlin2019: @yoomlam Can we merge this BgsClaimNotes with EventReason? One issue is those "key" (code in EventReason) used in BgsClaimNotes have different messages (narratives in EventReason).

yoomlam (approved):

au2 (approved):

PR #1434, opened 2023-04-10
Make swagger example for automatedClaim endpoint work

hayleymdawson:
## What was the problem?
Example code no longer works

- [MCP-2551] (https://amida.atlassian.net/browse/MCP-2551)

## How does this fix it?
Fields borrowed from Test 376 which will absolutely run successfully to expose in swagger.
Only hitch- I use a mock ICN because that will go through. If we dislike that, I could change the ICN to not look like mock, or I could use another test . 375 is also successful but I should change the veteran name to something more ok (the last name may not fly publicly)

## How to test this PR
Pull up swagger for the automated claim endpoint


jbregstone (approved):
Works for me when I have the mock services up. I looked at the PDF and it has lots of data.

hayleymdawson:
Oh that's a good question, should this document use the mocks at all? if not i'll swap to 375. @au2 as i realized in production the mocks wont exist.

ianwolson (approved):

au2:
@hayleymdawson  ideally it should be something that is runnable on development. Collection 350 and some LH Sandbox record are good in development but we need to find a claim id in BIP UAT that has anchors for this to be useful in development; i do not think that is happening anytime soon. This will never work in prodtest since claims are refresh there daily. So Mocks is the only choice for now.

PR #1446, opened 2023-04-11
MCP-2661: Exam order info enhancement

jbregstone:
## What was the problem?

We need to be able too see which claims have do not have an exam ordered.

- [MCP-2661](https://amida.atlassian.net/browse/MCP-2661)

## How does this fix it?[^1]

Adds a parameter for returning only exam info objects without an exam ordered. 

## How to test this PR
- Step 1
- Step 2[^secrel]


ianwolson (approved):

service/db/src/main/java/gov/va/vro/service/db/ClaimMetricsServiceImpl.java
```diff
@@ -126,10 +126,23 @@ private Page<ExamOrderEntity> findAllExamOrderInfoPage(ExamOrderInfoQueryParams
...
   @Override
-  public ExamOrdersInfo findAllExamOrderInfo(ExamOrderInfoQueryParams params) {
+  public ExamOrdersInfo findExamOrderInfo(ExamOrderInfoQueryParams params) {
     Page<ExamOrderEntity> examOrders = findAllExamOrderInfoPage(params);
-    List<ExamOrderInfoResponse> examOrdersInfo =
+    List<ExamOrderInfoResponse> allExamOrdersInfo =
         examOrderInfoResponseMapper.toExamOrderInfoResponses(examOrders);
-    return new ExamOrdersInfo(examOrdersInfo, examOrders.getTotalElements());
+    List<ExamOrderInfoResponse> notOrderedExamOrdersInfo = new ArrayList<>();
+    List<ExamOrderInfoResponse> response;
```
au2, line 134: Use JPA not manual filtering. Paging will not work right the way you are doing either. findByOrderAtIsNull should work I believe.

au2 (changes requested):
See inline. We should also have some JPA based unit tests with H2. Mockito will get you only so far for testing functionality.

jbregstone:
> See inline. We should also have some JPA based unit tests with H2. Mockito will get you only so far for testing functionality.

Could the JPA based unit tests with H2 be added in a separate PR?

I put in the JPA querying, not completely sure about the Paging issue. I wanted to keep the existing functionality the same (which uses a Page). I changed the new version use a List.

jbregstone:
End2end test failures seem to be unrelated, also happening on remote develop branch,
```
- gov.va.vro.end2end.VroV1Tests::positive7101Case01()
- gov.va.vro.end2end.VroV1Tests::positive6602Case01()
- gov.va.vro.end2end.VroV2Tests::testAutomatedClaimFullPositiveDuplicateBloodPressures()
```

au2 (approved):
Read the requirement again. We need to add another query parameter for dates I think. But I will add another ticker.

PR #1459, opened 2023-04-13
Add BIP CLAIM ISS to HELM

hayleymdawson:
## What was the problem?
<strike>Bip claim ISS was not available through helm, and was relying on application.yml defaults</strike>
BIP_CLAIM_ISS should not be an environment variable as it never really needs to change.
The same is true of BIP_EVIDENCE_ISS
Other cleanup was found while removing the first.
https://amida.atlassian.net/browse/MCP-2366
- [MCP-2366](https://amida.atlassian.net/browse/MCP-2366)

## How does this fix it?
Unused BIP_APPLICATION_ISS is removed
Unused JTI in BIP Api Props is removed
<strike>For Non local deployments- Helm now has BIP_CLAIM_ISS that should flow through the yml files.</strike>
<strike>For local deployments- setenv sets the variable, and it goes to the yml files that way. </strike>
BIP_CLAIM_ISS and BIP_EVIDENCE_ISS now just run from the yml with no environment variables present. 

## How to test this PR
Locally i can verify that everything still stands up and runs. <strike>@au2 if there is something else I can do to test helm, let me know. Or if this is not the approach you wanted for local, I saw the note in setenv that you tried to move them to the yml but that seems to be the opposite of what's desired here?</strike> these variables now do not change.


yoomlam:
These are kind-of all the same question:
What are the possible values of `BIP_CLAIM_ISS`? Will it ever be different than `virtual_regional_office`? Why is it an env variable and not a setting in `application-*.yml` files? Will we ever need to manually change it in the deployment envs?

hayleymdawson:
@yoomlam apparently we use VRO as an alternative value. but i dont know if we have to. @au2 for that.
We *do* use it in two containers- but not in production. it's used by mockBIP and vro main app. 
The Ticket referenced since I'm not sure you can see MCP tickets : "At this point BIP_CLAIM_ISS is a meaningless environment variable for deployments since it is not integrated with HELM. It only works because the right value for dev/prodtest etc. is given in application.yml as default. This is unexpected and already caused one prodtest issue. We need to do the integration to HELM charts."

yoomlam:
> We do use it in two containers- but not in production. it's used by mockBIP and vro main app.

Helm configurations are only for containers deployed to LHDI. Since we don't deploy mocks to LHDI, I don't think it needs to be an environment variable.

> At this point BIP_CLAIM_ISS is a meaningless environment variable for deployments since it is not integrated with HELM. It only works because the right value for dev/prodtest etc. is given in application.yml as default. This is unexpected and already caused one prodtest issue. We need to do the integration to HELM charts.

I don't quite understand the problem. How does turning a setting (used by a single deployed container) into an env var solve the problem.

au2:
@hayleymdawson @yoomlam I think things changed a lot in the helm land since I wrote this ticket a while back. I tend to agree with @yoom that we do not need the env vars and we can just use application.yaml files. I think separate value in local and default yaml files add some additional testing (that we are actually using the setting in the yaml file). But I do not feel strongly so if you want to remove override in application-local.yml for claim_issuer that is fine as well. You just need to update the mock service for the same.

hayleymdawson:
@yoomlam @au2 this has been updated to be just a cleanup ticket getting out unused items and moving the ISS variables to application.yml only.

mocks/mock-bip-ce-api/src/main/resources/application.yml
```diff
@@ -53,6 +53,6 @@ mock-bip-ce-api:
   jwt:
     userId: "${BIP_EVIDENCE_USERID:VRO_USER}"
     secret: "${BIP_EVIDENCE_SECRET:theSecret}"
-    issuer: "${BIP_EVIDENCE_ISS:VRO}"
+    issuer: "VRO"
```
yoomlam, line 56:
```suggestion
    # Should match the value of `evidenceIssuer` in app/src/main/resources/application.yml
    issuer: "VRO"
```

mocks/mock-bip-claims-api/src/main/resources/application.yml
```diff
@@ -53,7 +53,7 @@ mock-bip-claims-api:
   jwt:
     userId: "${BIP_CLAIM_USERID:VRO_USER}"
     secret: "${BIP_CLAIM_SECRET:theSecret}"
-    issuer: "${BIP_CLAIM_ISS:VRO}"
+    issuer: "virtual_regional_office"
```
yoomlam, line 56:
```suggestion
    # Should match the value of `claimIssuer ` in app/src/main/resources/application.yml
    issuer: "virtual_regional_office"
```

yoomlam (approved):
Suggested adding comments to associate the setting values.

au2 (approved):
Approving but you can also remove BIP_STATION_ID while at it maybe.

hayleymdawson:
@au2 i started to remove it but it touches so many files i'd rather do it in another PR vs hold this one. So merging now but we can put in a ticket and i'll work it quickly.

PR #1482, opened 2023-04-19
Move BGS config from helm back to svc-bgs-api

nanotone:
## What changed?
- Previously, the BGS microservice got its base URL config from the `$BGS_BASE_URL` environment variable set by helm.
- Now base URLs are set in the microservice's `config/settings.yml`, so that BGS-specific config isn't all over helm. It can still be overridden by `$BGS_BASE_URL`.

### Associated tickets or Slack threads:
- [Slack thread](https://dsva.slack.com/archives/C04QLHM9LR0/p1681401302286209)


svc-bgs-api/src/config/setup.rb
```diff
@@ -31,13 +31,13 @@ def puts(args)
 lib_path = File.expand_path '../../lib', __FILE__
 $LOAD_PATH.unshift lib_path
 
-ENVIRONMENT = ENV['ENVIRONMENT'] || 'development'
+ENVIRONMENT = ENV['ENV'] || 'dev'
```
yoomlam, line 34: Thinking aloud: Should the default be `local` or [end2end-test](https://github.com/department-of-veterans-affairs/abd-vro/blob/42d2dc260cbcf4b8fa075b8b7b067b3bcb3057ac/.github/workflows/end2end-test.yml#L48), where `bgs.base_url: http://mock-bgs-api:9020`. Since `docker-compose.yaml` sets the env var to override the URL: https://github.com/department-of-veterans-affairs/abd-vro/blob/0507eb7bae911570bf5e3db97535d653754c918a/app/src/docker/docker-compose.yml#L324
we can eliminate the need for `end2end-test`-specific settings. 

That leaves `local`. We can set `BGS_BASE_URL` in `setenv.sh` but I think that will incur more maintenance. I propose using `local` as the default and adding a `local` section to `settings.yml`, where `bgs.base_url: http://mock-bgs-api:9020`. This will allow us to remove `BGS_BASE_URL: http://mock-bgs-api:9020` from `docker-compose.yaml` too! Thoughts?

yoomlam (approved):
Thanks! Left one thought for consideration.

svc-bgs-api/src/config/setup.rb
yoomlam, line 34:
```suggestion
ENVIRONMENT = ENV['ENV'] || 'local'
```

PR #1513, opened 2023-04-27
Handle unexpected assessment exceptions gracefully

hayleymdawson:
## What was the problem?
In the event something goes wrong in the python processing, the java code could previously throw an error that was not handled gracefully and left the claim in a stuck state.
The errors that revealed this have since been fixed upstream, but this closes a potential hole in the code. 

- [MCP-2805](https://amida.atlassian.net/browse/MCP-2805)

## How does this fix it?
If an error happens after evidence is collected instead of throwing the exception, we offramp the claim with a new event reason. This is an unlikely scenario but, could happen. 

## How to test this PR
Testing this PR is something of a challenge as all known ways to cause the problem are currently fixed in the code. 
The JIRA ticket linked contains a file that can be replaced in svc-lighthouse to forcibly cause the error for testAutomatedClaimFullPositiveIncompleteBloodPressures in the VROV2EndToEndTests - which logs from the app module can then be used to verify.


hayleymdawson:
ah will get e2e test fixed shortly.

yoomlam (approved):

ianwolson (approved):

PR #1519, opened 2023-04-28
Mcp 2546 - Filter Lighthouse FHIR resource using _lastUpdated search parameter

wlin2019:
Provided additional FHIR resource filtering capability by using _lastUpdated.

## What was the problem?
We are looking for possible additional Lighthouse FHIR data filtering to reduce the unneeded data.

MCP-2546

## How does this fix it?[^1]
Add the capability to use _lastUpdated search parameter to filter old data.

## How to test this PR
- Step 1. Build and launch the app. Use Full Health Assessment endpoint. Look into svc-lighthouse-api log, it should show something like
   Get FHIR data from https://sandbox-api.va.gov/services/fhir/v0/r4/MedicationRequest?patient=9000682&_count=100

- Step 2. Add lh.filterLastUpdatedDate=2020-12-21 in application.properties in svc-lighthouse-api main resource directory, and repeat Step 1. In svc-lighthouse-api, it should show something like,
  Get FHIR data from https://sandbox-api.va.gov/services/fhir/v0/r4/Condition?patient=9000682&_lastUpdated=ge2020-12-21&category=encounter-diagnosis&_count=100


jbregstone (approved):

au2 (approved):

svc-lighthouse-api/src/main/java/gov/va/vro/abddataaccess/service/FhirClient.java
```diff
@@ -57,7 +57,7 @@ public class FhirClient {
 
   @Autowired private LighthouseApiService lighthouseApiService;
 
-  private String lastUpdated = null;
+  private String lastUpdated = "";
```
yoomlam: Initialize with empty String instead of null to simplify logic in `getLastUpdated()` and return empty String instead of null from that method.

svc-lighthouse-api/src/main/java/gov/va/vro/abddataaccess/service/FhirClient.java
```diff
@@ -319,14 +319,14 @@ private List<BundleEntryComponent> getRecords(
...
-      lastUpdated = "";
+    if (lastUpdated.isEmpty()) {
       String lastUpdatedDate = properties.getFilterLastUpdatedDate();
       if ((lastUpdatedDate != null) && !lastUpdatedDate.isEmpty()) {
         log.info("Update lastUpdated to {}", lastUpdatedDate);
         SimpleDateFormat dft = new SimpleDateFormat(LAST_UPDATED_DATE_FORMAT);
         try {
-          Date date = dft.parse(lastUpdatedDate);
+          // Verify that it is in the expected date format
+          dft.parse(lastUpdatedDate);
```
yoomlam: The `date` variable wasn't being used so I assume the `dft.parse(lastUpdatedDate)` was simply to check that `lastUpdatedDate` is in the expected date format. Is my assumption correct?

svc-lighthouse-api/src/main/java/gov/va/vro/abddataaccess/service/FhirClient.java
wlin2019: The purpose of setting it to NULL is to make it to be set ONLY once as a Singleton. getLastUpdated() is expected to return either an empty string or a valid FHIR accepted date string. It should never return a NULL.

PR #1535, opened 2023-05-03
Update port numbers to use new convention

yoomlam:
## What was the problem?

Port numbers (for mock services) were being assigned arbitrarily with no obvious pattern.

## How does this fix it?

Update port numbers to use new port numbering convention.
Now it's obvious when a mock service is being used.

Added https://github.com/department-of-veterans-affairs/abd-vro/wiki/Software-Conventions#port-numbers

## How to test this PR
- Run end2end test


buildSrc/src/main/groovy/local.java.container-spring-conventions.gradle
```diff
@@ -14,7 +14,11 @@ docker {
         new File(project.rootDir, 'buildSrc/docker/entrypoint-wrapper.sh'),
         new File(project.rootDir, 'buildSrc/docker/set-env-secrets.src')
 
-    buildArgs([JAR_FILE: bootJar.archiveFileName.get()])
+    def healthcheckPort = project.hasProperty('healthcheck_port') ? project.property("healthcheck_port") : 8080
+    buildArgs([
+        JAR_FILE: bootJar.archiveFileName.get(),
+        HEALTHCHECK_PORT_ARG: healthcheckPort
+    ])
```
yoomlam, line 21: If a `healthcheck_port` property is set (via `gradle.properties`), then use it, otherwise use 8080 default port that spring-actuator uses.

PR #1570, opened 2023-05-16
Exclude RRD from Helm configurations and VRO deployments

tcraghu:
[Exclude RRD from Helm configurations and VRO deployments]

## What was the problem?
App domain-rrd is remove from helm configuration by default 

ttps://github.com/department-of-veterans-affairs/abd-vro/issues/1510

## How does this fix it?[^1]
This fix will exclude app installation in helm by default. 

## How to test this PR
- Make sure that domain-rrd pod is not deployed after the deployment


helm/values.yaml
yoomlam, line 1: Only this change is needed. Changes to the other files are not needed.

tcraghu: I see there are environment specific values.yaml.  I did not see a code that loads values.yaml by default.  and also environment specific values.yaml. can you please point me to place where both values. yaml and values-<env>.yaml both are loaded?

yoomlam: Helm loads `values.yaml`: https://helm.sh/docs/chart_template_guide/values_files/
If you search for `values-for` in the code, you'll find https://github.com/department-of-veterans-affairs/abd-vro/blob/c290f59b51f7bd4c45f1992ae5f8dbaa28850d9b/.github/workflows/update-deployment.yml#L223

yoomlam (approved):

msnwatson (approved):

Issue #1574, opened 2023-05-18
Define VRO release process for partner product teams to prepare for CC team's launch


dianagriffin:
**User Story**

Follows #1498 

As a partner product team engineer, I want to know how to release my product on VRO, so that I have autonomy over my product's releases.

As a VRO engineer, I want to pair with my teammates to define the release creation process for partner product teams, so that we can provide guidance and support to the CC team for their June launch as well as future product teams.

**Acceptance Criteria**

1. Schedule a working session with VRO engineers & Diana to discuss & define the process of creating a release.
2. Document the process
3. This release process supports the CC team's launch plan.

dianagriffin:
Assumption: they need to deploy to staging/UAT and prod. @dianagriffin will confirm.

lukey-luke:
Mason is currently working to get images available for use in staging. Flagging here, that we need our docker images signed prior to deployment to prod environment in LHDI.

See:
 https://animated-carnival-57b3e7f5.pages.github.io/container-platform/signed-images/#what-happens-when-an-unsigned-or-unofficial-image-is-used
and: 
https://github.com/department-of-veterans-affairs/abd-vro/wiki/Deploying-VRO#ghcr-images

yoomlam:
Will be completed by end of sprint.

PR #1636, opened 2023-06-02
Yoom/gradle update version numbers

yoomlam:
## What was the problem?

A few things:
- Some TODOs about setting Helm Chart.yaml files' `appVersion`
- Deployment of different containers need to be somewhat independent of each other to allow domain containers to be deployed independently from VRO platform containers. We'll need to start versioning them differently.

- #1574

## How does this fix it?

*Starts* to allow independent deployment of containers by assigning `appVersion`s to Helm charts.

Subsequent work in a separate PR will associate a Helm chart version to a specific container image version.

## How to test this PR
Don't do this unless you're familiar with git.
```
./gradlew  release
```
Examine the new commit has updated `appVersion`s.
Delete the created tag and undo the commit. Don't push the changes.


openapi.json
yoomlam, line 1: We don't need this file in the repo.

msnwatson (approved):

PR #1690, opened 2023-06-15
Gradle: move buildSrc to gradle-plugins

yoomlam:
## What was the problem?

Came across https://proandroiddev.com/stop-using-gradle-buildsrc-use-composite-builds-instead-3c38ac7a2ab3, which promotes using a custom composite build over the built-in `buildSrc`.

Major benefit: faster builds -- details in article

Sidenote: [Composite builds](https://docs.gradle.org/current/userguide/composite_builds.html) will help partition the codebase, so that we'll have more control over what is built. For example, the `mocks` folder doesn't need to be built every time, resulting in even faster builds. Maybe more importantly, it can help diagnose and pinpoint build problems.

## How does this fix it?

Use a custom composite build for `gradle-plugins` folder.

- `mv buildSrc gradle-plugins`
- Replace all occurrences of `buildSrc` with `gradle-plugins`
- Add `includeBuild 'gradle-plugins'` to `settings.gradle`.

## How to test this PR
Run gradle all tasks.


msnwatson (approved):

dfitchett (approved):

engineer-plus-plus (commented):
lgtm

Issue #1668, opened 2023-06-08
Implement v1 of Python microservice library


nanotone:
**User Story**

As a Python developer, I want to import a library to handle all the boilerplate setup and ensure adherence to VRO microservice conventions, so that I can focus on the microservice functionality that I actually care about.

**Acceptance Criteria**

1. `hoppy` Python library is added to the `shared/` directory, and can be `pip install`'d directly with a Github URL (not via PyPI)
2. `hoppy` handles RabbitMQ connection and communication, including parsing config values from env vars, setting up RMQ exchange/consumers, and decoding/encoding JSON payloads.
3. Publish and document an example microservice using `hoppy`.

**Not included in this work**
Further work on shared microservice functionality that should be added to `hoppy` in the future:
- Connecting and talking to Redis
- Connecting and talking to other shared infrastructure (Postgres?)
- Framework for executing long-running jobs

**Notes about work**
- Software convention issue originally discussed in #900 
- [Slack huddle/thread](https://dsva.slack.com/archives/C04QLHM9LR0/p1678208466292349) where VRO engineers first talked about `hoppy` (including naming)

PR #1691, opened 2023-06-15
Add hoppy to shared directory as a pip-installable Python lib

nanotone:
This PR adds hoppy to `shared/lib-hoppy` as a Python package that can be installed with a one-line pip invocation.

- Partially fulfills #1668

## How to test this PR

Boilerplate for creating a throwaway virtualenv
```bash
$ python3 -m venv scratchpad
$ cd scratchpad
$ source bin/activate
```

And then,
```bash
$ pip install git+https://github.com/department-of-veterans-affairs/abd-vro.git@nanotone/lib-hoppy#subdirectory=shared/lib-hoppy
```

## Notes
I opted to create a PR just for adding hoppy, separate from an upcoming PR that will add a sample Python microservice that uses hoppy, for a couple reasons:
- Keeps PRs smaller
- As demonstrated in the Pip command, the install URL can specify not just the path within the repo, but also the branch (or any other commit identifier) to check out. I wanted the sample microservice to be able to work with `@nanotone/lib-hoppy` while this PR is in review, and then with `@develop` (or however we want to do versioning/releases of hoppy) after this PR is merged. This wouldn't be possible with a single PR.


msnwatson (approved):

shared/lib-hoppy/hoppy/mq.py
```diff
@@ -0,0 +1,117 @@
...
+    # There are 2 retry levels for the consumer. The first being in _create_connection() which is based on retry_limit
+    # The other being when the app crashes which is based on timeout
+    # If it fails, it will try to delete any existing connection and the reference to the instantiated class
+    # since the retry loop will recreate it
+    def run(self):
+        retry_time = None
+        while True:
+            connection = self._connect()
+            if not connection:
+                continue
```
dfitchett, line 47: > connection = self._connect()
            if not connection:
                        continue
If for some reason, self._connect() never returns a connection after hitting the retry limit, the loop will continue and try to create the connections again. Should it just raise the exception if it returns None?

yoomlam (approved):

shared/lib-hoppy/hoppy/mq.py
nanotone: I think it's fine for the outer loop to continue trying to create/recreate a connection -- the idea is that the microservice container itself will stick around forever, making attempts in case it was a transient network issue getting in the way. To me, that's preferable to letting an exception bubble up and trashing the container.

I'm happy to get into nitty-gritty distributed system strategies (and we probably should in a future PR) but this code was ported over to hoppy from an RRD microservice that worked well enough, so I'm inclined to leave it as-is.

yoomlam: I tend to agree with @nanotone. If this microservice can't connect (after it initially connected), the problem is likely with RabbitMQ or the deployment network, so the microservice should continue to retry. 

If the microservice throws an exception and causes the container to fail, then that may distract from our diagnostics. A container failure will cause K8s to continually and needlessly restart the container, consuming resources in the process. If all our microservices followed this convention and RabbitMQ is the problem, they'll all be restarting.

Perhaps this is a good scenario where a Docker health check can be implemented?
@dfitchett Let's capture this concern in a separate ticket for further team discussion.

dfitchett (approved):

PR #1701, opened 2023-06-17
Add new Python microservice with sample usage of hoppy

nanotone:
This PR adds a `svc-hoppy-usage` microservice written in Python which demonstrates how to use the `hoppy` library.

- Fulfills the remainder of #1668 
- Part one: #1691

## How to test this PR
- Bring up the RabbitMQ container and a `svc-hoppy-usage` container
- Send RabbitMQ requests to this service. Since I already had a Python environment with `pika` installed, it's just
```python
>>> import pika
>>> creds = pika.PlainCredentials("guest", "guest")
>>> params = pika.ConnectionParameters(host="localhost", port=5672, credentials=creds)
>>> props = pika.BasicProperties(reply_to="reply-queue")
>>> conn = pika.BlockingConnection(params)
>>> chan = conn.channel()
>>> chan.basic_publish("sample-exchange", "queue-one", '{"hello": "world"}', properties=props)
```

Things to try out, and the corresponding stdout from the service container:
- Sending anything to `queue-two`
```
WARNING:root:Error response generated: {'header': {'statusCode': 503, 'statusMessage': "Unhandled ServiceUnavailableError: Custom exception with parametrized argument 'no handler for this queue'"}}
```
- Sending `{"value": "0"}`, `{"valu": 0}`, and `{"value": 0}` to `queue-reciprocal`
```
WARNING:root:Error response generated: {'header': {'statusCode': 500, 'statusMessage': "Unhandled TypeError: unsupported operand type(s) for /: 'int' and 'str'"}}
WARNING:root:Error response generated: {'header': {'statusCode': 500, 'statusMessage': "Unhandled KeyError: 'value'"}}
WARNING:root:Error response generated: {'header': {'statusCode': 500, 'statusMessage': 'Unhandled ZeroDivisionError: division by zero'}}
```

## Notes
1. I was undecided whether this code should be part of domain-xample (as a microservice `svc-xample-p`?) or separate, but the latter felt like the path of lesser resistance. If we go with the former, I could use some help navigating adding a Python microservice to domain-xample's existing build process.
2. When #1691 gets merged, this Dockerfile's `pip install` command can leave out the branch name which should default to `develop`. We can also discuss tagging hoppy versions, or pegging them to VRO versioning, etc etc.


svc-hoppy-usage/src/docker/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+
+WORKDIR /app
+ARG ENTRYPOINT_FILE=entrypoint.sh
+# Workaround to copy ENTRYPOINT_FILE that may not exist
+# https://stackoverflow.com/a/46801962
+# Copy script that runs entrypoint.sh if it exists
+COPY set-env-secrets.src entrypoint-wrapper.sh ${ENTRYPOINT_FILE}* ./
+
+
+RUN pip install -v --no-cache-dir git+https://github.com/department-of-veterans-affairs/abd-vro.git@nanotone/lib-hoppy#subdirectory=shared/lib-hoppy
```
yoomlam, line 16: Reminder to update this to use the latest version once #1691 merges.

svc-hoppy-usage/src/docker/Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM python:3.9-alpine
```
yoomlam, line 1: To reduce SecRel maintenance, have it use the same base image as https://github.com/department-of-veterans-affairs/abd-vro/blob/1d804fe37a035fb607919e891acb5841e7677550/gradle-plugins/src/main/resources/docker/Dockerfile-python#L5.
```suggestion
FROM bitnami/python:3.10
```

svc-hoppy-usage/build.gradle
yoomlam, line 1: This is probably the wrong folder to use since this container is a usage example, rather than a VRO Platform microservice like the other `svc-*` folders. For reviewers, ignore the location of this container for now. I think I'll move it under the `domain-xample` folder after this PR is merged.

svc-hoppy-usage/src/docker/Dockerfile
nanotone: `requirements.txt` works like a charm. Removed the Dockerfile, and updated the pip URL to omit branch name.

svc-hoppy-usage/src/docker/Dockerfile
nanotone: Dockerfile was removed so this is moot. Tested the build with bitnami base image and it works great.

yoomlam (approved):

dfitchett (approved):

Issue #1675, opened 2023-06-12
Implement a mock general (non-BIE) Kafka service for testing and dev


dianagriffin:
**User Story**

As a VRO client dev, I want connect to Kafka events though standard VRO micro services, so that I can develop against them within the VRO platform without learning Kafka.

**Context**
- Use a Kafka container image that can support authentication (https://hub.docker.com/r/bitnami/kafka/) with minimal modifications
- Save the Dockerfile in mocks/mock-bie-kafka/src/docker/

**Acceptance Criteria**

1. A mock kafka service has been implemented and tested in a local environment
2. A client (which is not part of this ticket) could subscribe to the mock kafka service and be notified when an event is published to the topic
3. How to test the mock service is documented

**Notes about work**
- This ticket is part of epic #1642

dianagriffin:
Could include exploration of other options for mocking Kafka

PR #1710, opened 2023-06-20
Yoom/mock bie kafka

yoomlam:
## What was the problem?

Need a Mock Kafka service to implement a Kafka client.

- #1675

## How does this fix it?

Adds a Mock Kafka service

## How to test this PR
```
./gradlew docker
./gradlew  :dockerComposeUp
COMPOSE_PROFILES="kafka" ./gradlew -p mocks :dockerComposeUp
```

To test: read https://docs.confluent.io/platform/current/clients/kafkacat-usage.html, install `kcat`, then test it out:
```
   kcat -b 127.0.0.1:9092 -P -t topic1
   kcat -b 127.0.0.1:9092 -C -t topic1
```


mocks/mock-bie-kafka/src/docker/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+# Start container: docker run --rm -it -p 9092:9092 va/vro_mocks-mock-bie-kafka
+# To test: https://docs.confluent.io/platform/current/clients/kafkacat-usage.html
+#   kcat -b 127.0.0.1:9092 -P -t topic1
+#   kcat -b 127.0.0.1:9092 -C -t topic1
+# Note: use 127.0.0.1 (instead of `localhost`) in case /etc/hosts causes `localhost` to resolve to `::1`
+#   - https://github.com/edenhill/kcat/issues/16#issuecomment-66664574
+#   - https://www.confluent.io/blog/kafka-client-cannot-connect-to-broker-on-aws-on-docker-etc/
+
+# To run in Docker Compose setup: COMPOSE_PROFILES="kafka" ./gradlew -p mocks :dockerComposeUp
+# Start the kafka client in the Docker Compose network and connect to `mock-bie-kafka:9092`
```
yoomlam, line 17: Keep these instructions for easy reference until they're no longer needed or at the end of the Kafka epic.

mocks/docker-compose.yml
```diff
@@ -55,6 +55,20 @@ services:
     networks:
       - vro_intranet
 
+  mock-bie-kafka:
+    profiles: ["all","kafka"]
```
dfitchett, line 59: Would it be worthwhile to also have a "bie" profile in addition to "kafka"? We could use kafka for any future kafka service, and "bie" would be specific to that use case.

dfitchett (approved):

mocks/docker-compose.yml
yoomlam: I considered it but wanted to keep it short. Didn't want to use just `bie` since BIE may have other services. I could be convinced otherwise, plus we can change it later.

dfitchett: No need to change it now! We can always deal with that later! Not a hard change!

CorySohrakoffUSDS (approved):

PR #1762, opened 2023-06-29
Docker: use alpine for Python base image

greene858:
## What was the problem?
Bitnami base image has a SecRel issue at the OS level and did not have a clear timeline when it would be fixed
- #?

## How does this fix it?[^1]
Python images will build with python:3.10-alpine, as was previously used.


yoomlam (approved):

yoomlam:
Manually ran [3. Continuous Integration (run tests and checks)](https://github.com/department-of-veterans-affairs/abd-vro/actions/runs/5413823291)

msnwatson (approved):

dfitchett (approved):

PR #1766, opened 2023-06-29
Gradle: apply spotless plugin to Java library projects

yoomlam:
## What was the problem?

Spotless plugin wasn't applied to Java library projects

## How does this fix it?

Apply spotless plugin to Java library projects

## How to test this PR
All tests and CI checks pass


gradle-plugins/src/main/groovy/local.std.java.library-conventions.gradle
```diff
@@ -8,6 +8,7 @@ plugins {
 
     id 'starter.java.deps-open-tracing-common-conventions'
     id 'starter.java.publish-jar-conventions'
+    id 'starter.java.lint-spotless-java-conventions'
```
yoomlam, line 11: Crux of this PR. Other files were automatically updated due to `./gradlew  spotlessApply`

yoomlam:
[3. Continuous Integration (run tests and checks)](https://github.com/department-of-veterans-affairs/abd-vro/actions/runs/5413794923) works

PR #1799, opened 2023-07-12
story 1594 and 1777 Create bip api service.

engineer-plus-plus:
This is the third iteration of pull request solving story 1594

## What was the problem?
bip api doesn't exist

1594

## How does this fix it?[^1]
it provides the code for it.

## How to test this PR
- stanup rabbitmq
- run the integration test


app/docker-compose.yml
```diff
@@ -84,20 +84,22 @@ services:
     networks:
       - vro_intranet
```
engineer-plus-plus, line 85: Good job, Ilya.

engineer-plus-plus (commented):
ok

domain-rrd/rrd-workflows/src/main/java/gov/va/vro/service/provider/BipApiProps.java
greene858, line 1: If this class is part of the svc, please create it as part of the svc in that namespace; if not please leave domain-rrd out of the PR

greene858 (commented):
Please leave comment blocks, unless they are no longer accurate. If possible, re-use the xampleErrorHandler from domain-sample.

domain-rrd/rrd-workflows/src/main/java/gov/va/vro/service/provider/BipApiProps.java
engineer-plus-plus: I reverted the changes to this file.  It is not referenced from svc.  I just cleaned it up by mistake because it's a duplicate.

engineer-plus-plus: Here is the ref for the fix.  if this satisfies your recommendation, please click "Resolve Conversation" 
e32be15582fd7b3ed0453627fd50aed12fc8a89a

engineer-plus-plus:
> Please leave comment blocks, unless they are no longer accurate. If possible, re-use the xampleErrorHandler from domain-sample.

I've made the following changes to satisfy these recommendations.  If that's good enough and there are no additional change request, please press Approve Review.

a8f40671731fd1ea9371aa92d0ab2a0707fddb6e
a8f40671731fd1ea9371aa92d0ab2a0707fddb6e

svc-bip-api/src/main/java/gov/va/vro/bip/config/ConfigForSetClaimToRfdStatus.java
```diff
@@ -0,0 +1,46 @@
...
+    return new RabbitListenerErrorHandler() {
+      @Override
+      public Object handleError(
+          Message amqpMessage,
+          org.springframework.messaging.Message<?> message,
+          ListenerExecutionFailedException exception)
+          throws Exception {
+
+        return RMQConfig.respondToClientDueToUncaughtExcdeption(
+            amqpMessage, message, exception, new BipUpdateClaimResp());
```
dfitchett, line 42: Should this be a different response? 
```
            amqpMessage, message, exception, new BipUpdateClaimResp());
```

svc-bip-api/src/main/java/gov/va/vro/bip/service/RMQController.java
```diff
@@ -0,0 +1,38 @@
+package gov.va.vro.bip.service;
+
+import gov.va.vro.bip.model.*;
+import org.springframework.amqp.rabbit.annotation.RabbitListener;
+import org.springframework.beans.factory.annotation.Autowired;
+
+public class RMQController {
+  @Autowired BipApiService service;
+
+  @RabbitListener(queues = "getClaimDetailsQueue", errorHandler = "xampleErrorHandler")
```
dfitchett, line 10: I think all the error handlers on these `@RabbitListener` methods need to be updated to the corresponding error handler bean. For example this one should be updated to `@RabbitListener(queues = "getClaimDetailsQueue", errorHandler = "errorHandlerForGetClaimDetails"`

dfitchett (changes requested):
ErrorHandlers for `@RabbitListener`s need to be updated.

svc-bip-api/src/main/java/gov/va/vro/bip/service/RMQController.java
engineer-plus-plus: It was initially `errorHandlerForGetClaimDetails` but I changed it to `xampleErrorHandler` because of this comment https://shorturl.at/fgD59.   Please talk to Tom about which way you all want it to be.

svc-bip-api/src/main/java/gov/va/vro/bip/service/IBipApiService.java
greene858, line 1: Still not clear why the comment blocks were removed. Please restore, or correct them if they are not accurate.

svc-bip-api/src/main/java/gov/va/vro/bip/service/RMQController.java
dfitchett: That's fine with me, but that xampleErrorHandler is pointing directly to the bean by that name. If that works for all of these listeners then there isnt an issue and there is no point in declaring different error handlers for each queue.

engineer-plus-plus: depends on what is meant by "works,"  I guess.

engineer-plus-plus: I initially decide to have multiple handlers because methods in the service have different return values.  And it appeared that error handler has to return the same type as the `@RabbitListener` that caused the exception.  I was getting serialization error otherwise.

svc-bip-api/src/main/java/gov/va/vro/bip/config/ConfigForSetClaimToRfdStatus.java
engineer-plus-plus: what should it be?

svc-bip-api/src/main/java/gov/va/vro/bip/service/IBipApiService.java
engineer-plus-plus: I generally remove comments that are useless to the reader of the code.  For example if they just say the authors name.  or if they just duplicate what's already expressed by method signature without adding any value.  But as per your previous comment about this, [I already brought them back](https://github.com/department-of-veterans-affairs/abd-vro/pull/1799/commits/e623be7c78876d0d18874347073548e82f52099e).  Also yesterday.

dfitchett (approved):

greene858 (approved):
LGTM

PR #1800, opened 2023-07-12
Updating latest snappy java version using override file

### LINKED ISSUE #1798: SecRel: Package upgrade for snappy-java in dev_vro-svc-bie-kafka image
**User Story**

Resolve Snappy Java package upgrade sec issue.

https://github.com/department-of-veterans-affairs/abd-vro-internal/actions/runs/5534927706

High | snappy-java is a fast compressor/decompressor for Java. Due to unchecked multiplications, an integer overflow may occur in versions prior to 1.1.10.1, causing a fatal error. The function shuffle(int[] input) in the file BitShuffle.java receives an array of integers and applies a bit shuffle on it. It does so by multiplying the length by 4 and passing it to the natively compiled shuffle function. Since the length is not tested, the multiplication by four can cause an integer overflow and become a smaller value than the true size, or even zero or negative. In the case of a negative value, a java.lang.NegativeArraySizeException exception will raise, which can crash the program. In a case of a value that is zero or too small, the code that afterwards references the shuffled array will assume a bigger size of the array, which might cause exceptions such as java.lang.ArrayIndexOutOfBoundsException. The same issue exists also when using the shuffle functions that receive a double, float, long and short, each using a different multiplier that may cause the same issue. Version 1.1.10.1 contains a patch for this vulnerability. | Upgrade package snappy-java to version 1.1.10.1 or above. | 1.1.10.1 | CVE-2023-34453
-- | -- | -- | -- | --
High | snappy-java is a fast compressor/decompressor for Java. Due to unchecked multiplications, an integer overflow may occur in versions prior to 1.1.10.1, causing an unrecoverable fatal error. The function compress(char[] input) in the file Snappy.java receives an array of characters and compresses it. It does so by multiplying the length by 2 and passing it to the rawCompressfunction.  Since the length is not tested, the multiplication by two can cause an integer overflow and become negative. The rawCompress function then uses the received length and passes it to the natively compiled maxCompressedLength function, using the returned value to allocate a byte array.  Since the maxCompressedLength function treats the length as an unsigned integer, it doesn’t care that it is negative, and it returns a valid value, which is casted to a signed integer by the Java engine. If the result is negative, ajava.lang.NegativeArraySizeExceptionexception will be raised while trying to allocate the arraybuf. On the other side, if the result is positive, the bufarray will successfully be allocated, but its size might be too small to use for the compression, causing a fatal Access Violation error.  The same issue exists also when using thecompress` functions that receive double, float, int, long and short, each using a different multiplier that may cause the same issue. The issue most likely won’t occur when using a byte array, since creating a byte array of size 0x80000000 (or any other negative value) is impossible in the first place. Version 1.1.10.1 contains a patch for this issue. | Upgrade package snappy-java to version 1.1.10.1 or above. | 1.1.10.1 | CVE-2023-34454
High | snappy-java is a fast compressor/decompressor for Java. Due to use of an unchecked chunk length, an unrecoverable fatal error can occur in versions prior to 1.1.10.1. The code in the function hasNextChunk in the fileSnappyInputStream.java checks if a given stream has more chunks to read. It does that by attempting to read 4 bytes. If it wasn’t possible to read the 4 bytes, the function returns false. Otherwise, if 4 bytes were available, the code treats them as the length of the next chunk. In the case that the compressedvariable is null, a byte array is allocated with the size given by the input data. Since the code doesn’t test the legality of the chunkSize variable, it is possible to pass a negative number (such as 0xFFFFFFFF which is -1), which will cause the code to raise a java.lang.NegativeArraySizeException exception. A worse case would happen when passing a huge positive value (such as 0x7FFFFFFF), which would raise the fatal java.lang.OutOfMemoryError error. Version 1.1.10.1 contains a patch for this issue. | Upgrade package snappy-java to version 1.1.10.1 or above. | 1.1.10.1 | CVE-2023-34455

High	snappy-java is a fast compressor/decompressor for Java. Due to unchecked multiplications, an integer overflow may occur in versions prior to 1.1.10.1, causing a fatal error. The function shuffle(int[] input) in the file BitShuffle.java receives an array of integers and applies a bit shuffle on it. It does so by multiplying the length by 4 and passing it to the natively compiled shuffle function. Since the length is not tested, the multiplication by four can cause an integer overflow and become a smaller value than the true size, or even zero or negative. In the case of a negative value, a java.lang.NegativeArraySizeException exception will raise, which can crash the program. In a case of a value that is zero or too small, the code that afterwards references the shuffled array will assume a bigger size of the array, which might cause exceptions such as java.lang.ArrayIndexOutOfBoundsException. The same issue exists also when using the shuffle functions that receive a double, float, long and short, each using a different multiplier that may cause the same issue. Version 1.1.10.1 contains a patch for this vulnerability.	Upgrade package snappy-java to version 1.1.10.1 or above.	1.1.10.1	https://github.com/advisories/GHSA-pqr6-cmr2-h8hf
High	snappy-java is a fast compressor/decompressor for Java. Due to unchecked multiplications, an integer overflow may occur in versions prior to 1.1.10.1, causing an unrecoverable fatal error. The function compress(char[] input) in the file Snappy.java receives an array of characters and compresses it. It does so by multiplying the length by 2 and passing it to the rawCompressfunction.  Since the length is not tested, the multiplication by two can cause an integer overflow and become negative. The rawCompress function then uses the received length and passes it to the natively compiled maxCompressedLength function, using the returned value to allocate a byte array.  Since the maxCompressedLength function treats the length as an unsigned integer, it doesn’t care that it is negative, and it returns a valid value, which is casted to a signed integer by the Java engine. If the result is negative, ajava.lang.NegativeArraySizeExceptionexception will be raised while trying to allocate the arraybuf. On the other side, if the result is positive, the bufarray will successfully be allocated, but its size might be too small to use for the compression, causing a fatal Access Violation error.  The same issue exists also when using thecompress` functions that receive double, float, int, long and short, each using a different multiplier that may cause the same issue. The issue most likely won’t occur when using a byte array, since creating a byte array of size 0x80000000 (or any other negative value) is impossible in the first place. Version 1.1.10.1 contains a patch for this issue.	Upgrade package snappy-java to version 1.1.10.1 or above.	1.1.10.1	https://github.com/advisories/GHSA-fjpj-2g6w-x25r
High	snappy-java is a fast compressor/decompressor for Java. Due to use of an unchecked chunk length, an unrecoverable fatal error can occur in versions prior to 1.1.10.1. The code in the function hasNextChunk in the fileSnappyInputStream.java checks if a given stream has more chunks to read. It does that by attempting to read 4 bytes. If it wasn’t possible to read the 4 bytes, the function returns false. Otherwise, if 4 bytes were available, the code treats them as the length of the next chunk. In the case that the compressed variable is null, a byte array is allocated with the size given by the input data. Since the code doesn’t test the legality of the chunkSize variable, it is possible to pass a negative number (such as 0xFFFFFFFF which is -1), which will cause the code to raise a java.lang.NegativeArraySizeException exception. A worse case would happen when passing a huge positive value (such as 0x7FFFFFFF), which would raise the fatal java.lang.OutOfMemoryError error. Version 1.1.10.1 contains a patch for this issue.	Upgrade package snappy-java to version 1.1.10.1 or above.	1.1.10.1	https://github.com/advisories/GHSA-qcwq-55hx-v3vh

**Acceptance Criteria**
Package is upgraded and SecRel issues resolved

tejans24:
## What was the problem?
Older version of Snappy Java has security issues

- #1798 

## How does this fix it?[^1]
Fixes #1798 - updating a dependent package of Kafka lib

## How to test this PR
- Makes sure that SecRel is passing after un-suppressing snappy java issues.

### DIFF
--- a/gradle-plugins/src/main/groovy/shared.java.vro-dep-constraints.gradle
+++ b/gradle-plugins/src/main/groovy/shared.java.vro-dep-constraints.gradle
@@ -46,6 +46,11 @@ dependencies {
             because("security vulnerability in prior versions")
         }
 
+        // For Kafka svc-bie-kafka service override
+        implementation('org.xerial.snappy:snappy-java:1.1.10.1') {
+            because("security vulnerability in prior versions")
+        }
+
         implementation('com.google.guava:guava:32.0.1-jre') {
             because("security vulnerability in prior versions")
         }
@@ -53,5 +58,6 @@ dependencies {
         // These are used indirectly but we need to specify versions, otherwise build error
         annotationProcessor "org.springframework.boot:spring-boot-autoconfigure-processor:${spring_boot_version}"
         annotationProcessor "org.mapstruct:mapstruct-processor:${mapstruct_version}"
+
     }
 }


gradle-plugins/src/main/groovy/shared.java.vro-dep-constraints.gradle
```diff
@@ -46,12 +46,18 @@ dependencies {
             because("security vulnerability in prior versions")
         }
 
+        // For Kafka svc-bie-kafka service override
+        implementation('org.xerial.snappy:snappy-java:1.1.10.1') {
+            because("security vulnerability in prior versions")
```
yoomlam, line 51: The `because()` isn't particularly useful since all the constraints in this block resulted from the same reason: SecRel alerts. I recommend leaving it out or referencing the vulnerability ID or CVE ID provided by Aqua -- feel free to bring it up for a team discussion.

yoomlam (approved):

gradle-plugins/src/main/groovy/shared.java.vro-dep-constraints.gradle
msnwatson: hmm my thoughts were to include this so it's clear that we cannot go backwards and why it is important that there are no prior versions included in the project. not super attached to it though

yoomlam: We can add that as a code comment at the beginning of the constraints block so we don't repeat the same message for each dependency.

tejans24:
https://github.com/department-of-veterans-affairs/abd-vro-internal/actions/runs/5575981577 passing secrel aqua on Snappy Java.

Issue #1821, opened 2023-07-19
Store maximum ratings in VRO


amylai-nava:
As a VA product owner, I want to store the maximum ratings for each disability in VRO to enable the max CFI project and potentially other claims-fast tracking efforts in the future.

### Acceptance Criteria
* Create a new API handler in VRO for max CFI
* Implement an API endpoint that takes a list of disabilities and returns their max ratings if known.
  * For MVP, the only disability with a "known" max rating is tinnitus (diagnostic code 6260), which has a max rating of 10%

### Not included in this work
* Including max ratings for disabilities other than tinnitus
* Figuring out how we'll ensure that the max rating stays up-to-date with VA regulations

### Relevant link
* related VRO/EE kick off slack thread: https://dsva.slack.com/archives/C01Q7979Z7D/p1689338645726979
* VRO software conventions: https://github.com/department-of-veterans-affairs/abd-vro/wiki/Software-Conventions#vro-software-conventions
* VRO intro and setup documentation: https://github.com/department-of-veterans-affairs/abd-vro/wiki#welcome-application-developers
* https://github.com/department-of-veterans-affairs/abd-vro/issues/1715#issuecomment-1615227087

amylai-nava:
relatedly: we need to first decide on the language and tech stack

amylai-nava:
@dfitchett @nanotone, here's the `phase 2a` ticket that needs to be broken down further. 

I had made a note earlier that we needed to decide on the language, but it sounds like python is the answer ✅

amylai-nava:
takeaways from our chat today:
- we can split this task into two
  - one for implementing the API endpoint (this ticket
  - one for integrating it with vets-api (new ticket)
- for our pilot, we'll keep the API as simple as possible, leveraging what team CC's API

PR #1851, opened 2023-07-26
1821: Created Max CFI API application. 

dfitchett:
Added MAX CFI API to API Gateway. Added integration test for API Gateway. Added configurations to secrel/config.yml, delete-published-dev-images.yml, and delete-published-images.yml

## What was the problem?
Team EE needs an API to retrieve the max rating for a list of diagnostic codes. 

- #1821

## How does this fix it?[^1]

EE Max CFI API application created with a single post endpoint. 

## How to test this PR
- refer to README in the `domain-ee/README.md` folder for testing instructions
- successful run: [continuous-integration](https://github.com/department-of-veterans-affairs/abd-vro/actions/runs/5672041601/job/15370423901)


domain-ee/max-cfi-app/src/python_src/api.py
```diff
@@ -0,0 +1,62 @@
...
+            }
+            ratings.append(rating)
+        else:
+            raise HTTPException(status_code=404, detail=f"Could not find max rating for diagnostic_code={dc}")
+
+    response = {
+        "ratings": ratings
+    }
+
+    logging.info(f"event=getMaxRating response={response}")
```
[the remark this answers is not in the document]
dfitchett: I'm not sure this is an issue here since we do validation on the diagnostic codes array and each entry individually.

.github/workflows/delete-published-images.yml
```diff
@@ -53,6 +53,7 @@ jobs:
         - vro-svc-bie-kafka
         - vro-xample-workflows
         - vro-cc-app
+        - vro-max-cfi-app
```
tejans24, line 56: nit: should this have the domain within the name? such as `vro-ee-max-cfi-app` for each domain recognition?

tejans24 (approved):
LGTM, it looks like VRO conventions are being followed and structure seems to make sense. I haven't looked at the domain logic/python code in depth since it's outside of the VRO scope.

We'll have one other VRO engineer review this PR as well since it's the initial start of EE onboarding!

nanotone (changes requested):
This is great! I have two change requests to make of the API itself:
- Ratings are always in whole-percentage amounts, and expressed in VBA business logic as integer percentage-points. I'd stick to that convention here, as it also gives us peace of mind to never worry about subtle floating representation and serialization issues.
- It looks like the current behavior is to return 404 if _any_ diagnostic code isn't found in the VRO-side table. I appreciate the nod to HTTP semantics but think the API would be more useful if it just returned 200 with the known max ratings. From the perspective of the client, it can deduce any missing ratings as unknown, but it can't do anything with a 404 besides retry with a subset of diagnostic codes.

That's it for the important stuff. I have a smattering of Python-stylistic suggestions to make in a later pass, but this looks really solid already.

.github/workflows/delete-published-images.yml
yoomlam: Agree. Perhaps the `domain-ee` folder should be `domain-max-cfi`?

domain-ee/.pre-commit-config.yaml
yoomlam, line 1: Is this file actually used? I know there is a `domain-cc/.pre-commit-config.yaml` but I question if it's being used. I'm only aware of `/.pre-commit-config.yaml` being used as documented at https://github.com/department-of-veterans-affairs/abd-vro/wiki/Development-process#pre-commit-hook

dfitchett: Yeah, it is still used for python as well within pycharm if pre-commit is installed.

domain-ee/ee-max-cfi-app/src/python_src/api.py
```diff
@@ -0,0 +1,70 @@
...
+    format="[%(asctime)s] %(levelname)-8s %(message)s",
+    level=logging.INFO,
+    datefmt="%Y-%m-%d %H:%M:%S",
+    stream=sys.stdout,
+)
+
+
+@app.get("/health")
+def get_health_status():
+    if not len(MAX_RATINGS_BY_CODE):
```
nanotone, line 36: Empty lists are falsey, so this can just be `if not MAX_RATINGS_BY_CODE:`.

nanotone (commented):
Just a few Python style nits, but you're good to go once those are fixed!

nanotone (approved):
LGTM!

Issue #1863, opened 2023-07-28
Create an interface between kafka consumer and rabbitmq with proper data schema(s)


tejans24:
**User Story**

As an engineer, I want send BIE data from Kafka consumer to Rabbitmq -> xample with all the fields and shared interface

**Dependent on** 
- PR #1859 

**Context**
You will need to:
- Create mock pojo classes out of Avro files (until schema registry is figured out)
- Create a common interface for Kafka consumer to send messages to RabbitMQ service based on the topics
- Add fields to BieMessagePayload to extract certain critical fields, store the rest as generic map in `eventDetails` until there is a use for other fields.

**Acceptance Criteria**

1. Kafka Consumer can send BIE schema based messages to Rabbitmq
2. RabbitMQ can properly associate those messages to `BieMessagePayload` and send the messages to consumers

**Notes about work**
- This ticket is part of epic #1642

Issue #1680, opened 2023-06-12
Update example domain code to save BIE-specific event details to the DB


dianagriffin:
**User Story**

As VRO product owner, I want to track BIE contention events, so that I have a proof of concept that VRO can provide visibility into claims data.

**Dependencies**
- #1678 
- #1677 

**Context**
You will need to:
- Subscribe to BIE topics
- Update integration tests

**Acceptance Criteria**

1. Example domain code saves BIE-specific event details to the DB. 

Ideally this would include:

- The event type, e.g. “Contention created”, “Contention updated”, etc.
- The event timestamp
- Benefit Claim ID
- Contention ID
- Diagnostic Code(s)
- Classification (if available)

See #1642 for more details

**Notes about work**
- This ticket is part of epic #1642

dianagriffin:
Event Bus team has looked at these BIE event streams and has some helpful detail written up here: https://github.com/department-of-veterans-affairs/VES/blob/master/research/Event%20Bus/Use%20Cases/Contention%20Events%20Use%20Case%20One%20Pager.md

yoomlam:
@tejans24 There's DB schema info here: https://github.com/department-of-veterans-affairs/abd-vro/wiki/BIE-Kafka-Client#xample-workflow

dianagriffin:
Blocked pending the working schema

PR #1869, opened 2023-07-31
Handle BIE kafka events, pass through the queue and store them in db

tejans24:
## What was the problem?

- #1863 
- #1680 

## How does this fix it?[^1]
- Deserialize AVRO data and pass it into Rabbitmq
- Handles string json data for local purposes
  - This will need to be changed to Avro once we have Schema Registry Mocks
- Stores all of the data from BIE kafka messages in db

## How to test this PR
- Merge into develop, deploy it to dev and try out changes


shared/lib-bie-kafka/src/main/java/gov/va/vro/model/biekafka/ContentionEvent.java
```diff
@@ -0,0 +1,19 @@
...
+  CONTENTION_DELETED("CONTENTION_BIE_CONTENTION_DELETED_V02");
+
+  private final String topicName;
+
+  ContentionEvent(String topicName) {
+    this.topicName = topicName;
+  }
+
+  public String getTopicName() {
+    return "TST_" + topicName;
```
tejans24, line 17: prefix needs to be updated based on the application.yaml

chengjie8 (approved):
Looks good overall. Hopefully, once we figure out the mock schema registry, we would no longer need to use `StringDeserializer` for local

shared/lib-bie-kafka/src/main/java/gov/va/vro/model/biekafka/test/BieMessagePayloadFactory.java
yoomlam, line 1: Move this to `src/test`

yoomlam (changes requested):

db-init/src/main/resources/database/migrations/V1.2__BIE_Contention_Event_Recreate_Table.sql
```diff
@@ -0,0 +1,19 @@
+-- Drop the old table
+DROP TABLE IF EXISTS "bie_contention_event";
+
+-- Create the new table
+CREATE TABLE IF NOT EXISTS "bie_contention_event"
```
yoomlam, line 5: Since you're using the same table name, why not remove the 2 undesired columns (`event` and `event_details`) and add the new columns?

domain-xample/xample-workflows/src/main/java/gov/va/vro/routes/xample/DbHelper.java
```diff
@@ -50,12 +52,19 @@ private VeteranEntity createVeteran(final String veteranIcn, final String vetera
 
   public ContentionEventEntity saveContentionEvent(final BieMessagePayload bieMessagePayload) {
     final ContentionEventEntity contentionEventEntity = new ContentionEventEntity();
-    contentionEventEntity.setEventType(bieMessagePayload.getEvent());
+    contentionEventEntity.setClaimId(bieMessagePayload.getClaimId());
+    contentionEventEntity.setContentionId(bieMessagePayload.getContentionId());
+    contentionEventEntity.setEventType(bieMessagePayload.getEventType().toString());
+    contentionEventEntity.setEventDetails(bieMessagePayload.getEventDetails());
```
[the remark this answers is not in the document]
tejans24: removed it in a commit

db-init/src/main/resources/database/migrations/V1.2__BIE_Contention_Event_Recreate_Table.sql
tejans24: It's much cleaner, especially with data already existing that will be inconsistent to reset the table. There is not much cost for it.

shared/lib-bie-kafka/src/main/java/gov/va/vro/model/biekafka/test/BieMessagePayloadFactory.java
yoomlam: I see; it's being used by `domain-xample` and `svc-bie-kafka`. It's acceptable for now. 👍 

For future reference, some options if this is a common problem:
- create a `lib-bie-kafka-test` Gradle subproject, which is overkill for 1 shared class.
- copy the class to each project; must update all occurrences when the class changes.
- use symlinks to the class in each project; incompatible with Windows without more setup.

yoomlam:
Running [CI: BIE Kafka End-2-End Test](https://github.com/department-of-veterans-affairs/abd-vro/actions/runs/5895187894)

build.gradle
```diff
@@ -173,3 +173,27 @@ tasks.named("dockerComposeDown").configure {
...
+  dependsOn ':dockerComposeDown', ':app:dockerComposeDown', ':domain-xample:dockerComposeDown'
+  doLast {
+    // Setting an environment variable for the exec task
+    exec {
+      environment 'COMPOSE_PROFILES', 'kafka'
+      commandLine './gradlew', '-p', 'mocks', ':dockerComposeDown'
+    }
+    println 'Started Kafka related containers!'
+  }
+}
```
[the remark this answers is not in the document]
yoomlam: Why is this resolved?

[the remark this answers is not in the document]
tejans24: I removed it altogether, I'll add it back in a separate PR.

db-init/src/main/resources/database/migrations/V1.2__BIE_Contention_Event_Recreate_Table.sql
```diff
@@ -0,0 +1,38 @@
...
+    claim_id BIGINT,
+
+    -- Reference to the specific contention
+    contention_id BIGINT,
+
+    -- Code representing the type of diagnostic related to the contention
+    diagnostic_type_code VARCHAR(255),
+
+    -- Code classifying the contention
+    contention_classification_code VARCHAR(255),
```
yoomlam, line 29: nit: How was 255 chosen?

shared/lib-bie-kafka/build.gradle
```diff
@@ -8,4 +8,6 @@ ext {
 
 dependencies {
   implementation 'org.springframework.boot:spring-boot-starter-amqp'
+  implementation 'net.datafaker:datafaker:2.0.1'
```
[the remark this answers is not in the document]
yoomlam: How is this resolved?

svc-bie-kafka/build.gradle
```diff
@@ -24,7 +24,6 @@ jacocoTestReport {
 
 dependencies {
   implementation project(":shared:lib-bie-kafka")
-
```
yoomlam, line 27: nit: I think this empty line separates the group of Avro dependencies nicely

svc-bie-kafka/src/test/java/gov/va/vro/services/bie/config/KafkaConsumerTest.java
yoomlam, line 1: What's being tested by this class?

.github/workflows/bie-kafka-end2end-test.yml
```diff
@@ -136,7 +137,7 @@ jobs:
             FROM
               claims.bie_contention_event
             WHERE
-              event_details='Lorem ipsum'
+              event_type='${{env.EVENT_TYPE}}'
```
yoomlam, line 140: [CI: BIE Kafka End-2-End Test](https://github.com/department-of-veterans-affairs/abd-vro/actions/runs/5895187894) failed

svc-bie-kafka/src/test/java/gov/va/vro/services/bie/config/KafkaConsumerTest.java
tejans24: this hasn't changed much from develop. @chengjie8 will add code back to this one schema registry mock is figured out. Without it, it's hard to send messages in Avro to this topic.

.github/workflows/bie-kafka-end2end-test.yml
tejans24: What's the event_type being used for this integration test? It's a bit abstract in which how the query is selected. I would have preferred all to be in a single test file.

yoomlam: You added `EVENT_TYPE` and `event_type`. They're not used elsewhere.

See https://github.com/department-of-veterans-affairs/abd-vro/blob/c51fbb917e0009fbde55d7318a227d5cfddb60b2/svc-bie-kafka/src/integrationTest/java/gov/va/vro/services/bie/BieKafkaApplicationTest.java#L68-L72
and https://github.com/department-of-veterans-affairs/abd-vro/pull/1869/files#diff-59cc2c7b74dd0f68f656e8a2b3914053117bf9b2a6ae60fa10816a77e7a90350L61

yoomlam: > In the previous iteration of this test - we were using kcat to send the data to kafka ...

Yes. Since I added SSL, kcat would need to authenticate and I couldn't get that to work. Also, `BieKafkaApplicationTest` uses the SSL configuration (https://github.com/department-of-veterans-affairs/abd-vro/blob/c51fbb917e0009fbde55d7318a227d5cfddb60b2/svc-bie-kafka/src/main/resources/application-integration-test.yaml#L15-L21), similar to our Kafka microservice, so the test also ensures the Spring-based SSL config works.

> Does the Java integration test that's added here actually send the data to the db through the xample workflow?

It sends it to our Kafka microservice -- check the microservice code to see what gets saved to the DB.

yoomlam: Here are some options:
   1. Move the `Check for saved DB entry in Postgres` logic (https://github.com/department-of-veterans-affairs/abd-vro/blob/c51fbb917e0009fbde55d7318a227d5cfddb60b2/.github/workflows/bie-kafka-end2end-test.yml#L124) into `BieKafkaApplicationTest`.
   2. Read an env var like `EVENT_BODY` (that is set in `bie-kafka-end2end-test.yml`) and set `BieKafkaApplicationTest.kafkaEventBody` to the var value. Then update `Check for saved DB entry in Postgres` to look for the expected DB entry.

tejans24:
https://github.com/department-of-veterans-affairs/abd-vro/actions/runs/5931443455/job/16083319517

.github/workflows/bie-kafka-end2end-test.yml
tejans24: https://github.com/department-of-veterans-affairs/abd-vro/pull/1869/commits/ac45d2b42246e0fad4a4cbc1f71daac8d2fe3dc9

shared/lib-bie-kafka/src/main/java/gov/va/vro/model/biekafka/ContentionEvent.java
yoomlam, line 1: The list of Kafka topics are listed in the `application.yaml` as well as in this new enum class. This information should be in 1 location. I recommend using the `BieProperties` class instead.

Our Kafka microservice treats these topics as a collection (map) of strings (to MQ exchange names), which is why I think it belongs in the yml file as configuration settings. The microservice doesn't care what the Kafka topics are semantically, i.e., the code logic doesn't change for different topics. If this microservice treated topics like constants and had different behavior for specific a topic, then an enum would be warranted.

svc-bie-kafka/src/main/java/gov/va/vro/services/bie/service/kafka/MessageHelper.java
yoomlam, line 1: I suggest moving these methods to the `ContentionEvent` enum class. https://stackoverflow.com/questions/3340032/are-utility-classes-evil
https://medium.com/@kamilmasyhur/you-should-avoid-the-use-of-util-class-d777ac664fd4

yoomlam (changes requested):
Looks good! Some minor suggestions.

shared/lib-bie-kafka/src/main/java/gov/va/vro/model/biekafka/BieMessagePayload.java
```diff
@@ -1,20 +1,23 @@
...
+@EqualsAndHashCode
 public class BieMessagePayload {
   private Integer status;
   private String statusMessage;
-  private String event;
-  private String eventDetails;
-  private String notifiedAt;
+  private ContentionEvent eventType;
+  private long claimId;
+  private long contentionId;
```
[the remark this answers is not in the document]
tejans24: good catch!

yoomlam:
Reminder to add a link to a passing end-to-end test.

.github/workflows/bie-kafka-end2end-test.yml
```diff
@@ -9,8 +9,8 @@ on:
 
 env:
   COMPOSE_PROFILES: 'kafka'
+  # Maps to the value of the BieKafkaApplicationTest run by ./gradlew :svc-bie-kafka:integrationTest
```
yoomlam, line 12:
```suggestion
```

svc-bie-kafka/src/main/java/gov/va/vro/services/bie/service/kafka/MessageHelper.java
```diff
@@ -0,0 +1,22 @@
+package gov.va.vro.services.bie.service.kafka;
+
+import gov.va.vro.model.biekafka.ContentionEvent;
+
+import java.util.Arrays;
+
+public class MessageHelper {
+  public static ContentionEvent mapTopicToEvent(String topic) {
+    // remove first word prefix from topic seperated by _
+    String noPrefixTopic = topic.substring(topic.indexOf("_") + 1);
```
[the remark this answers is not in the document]
tejans24: True, now that this is moved into the `ContentionEvent` which is in a shared library, it doesn't have access to the bie properties. Passing in as a variable seems a bit ugly overall, since java doesn't support default params. Let me create another issue to get this resolved.

tejans24:
Passing integration tests - https://github.com/department-of-veterans-affairs/abd-vro/actions/runs/5967202745

shared/lib-bie-kafka/build.gradle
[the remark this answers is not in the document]
tejans24: this can't be converted to use testImplementation because it's being used in shared factory which is in src

svc-bie-kafka/src/main/java/gov/va/vro/services/bie/config/MessageExchangeConfig.java
```diff
@@ -28,11 +30,14 @@ public MessageConverter messageConverter() {
   @Bean
   Declarables topicBindings(final BieProperties bieProperties) {
     final List<AbstractDeclarable> list =
-        bieProperties.getTopicToExchangeMap().values().stream()
+        Arrays.stream(bieProperties.topicNames())
```
yoomlam, line 33: Since Kafka topics are listed in ContentionEvent only (and not in application.yaml), this should iterate through the `ContentionEvent` enum values. Then, line 38 can be `contentionEvent.rabbitMqExchangeName()`.

shared/lib-bie-kafka/src/main/java/gov/va/vro/model/biekafka/ContentionEvent.java
```diff
@@ -0,0 +1,36 @@
...
+    this.topicName = topicName;
+  }
+
+  public String getTopicName() {
+    return topicName;
+  }
+
+  public static ContentionEvent mapTopicToEvent(String topic) {
+    // remove first word prefix from topic seperated by _
+    String noPrefixTopic = topic.substring(topic.indexOf("_") + 1);
```
yoomlam, line 24: Create issue for https://github.com/department-of-veterans-affairs/abd-vro/pull/1869#discussion_r1304700625

shared/lib-bie-kafka/src/main/java/gov/va/vro/model/biekafka/ContentionEvent.java
```diff
@@ -0,0 +1,36 @@
...
+
+    return Arrays.stream(ContentionEvent.values())
+        .filter(event -> event.getTopicName().equals(noPrefixTopic))
+        .findFirst()
+        .orElseThrow(() -> new IllegalArgumentException("Unrecognized topic: " + noPrefixTopic));
+  }
+
+  public static String generateRabbitMQChannelName(String topic) {
```
yoomlam, line 32:
```suggestion
  public static String generateRabbitMqExchangeName(String topic) {
```

After addressing my comment for MessageExchangeConfig.java, this will be non-static:
```suggestion
  public static String rabbitMqExchangeName() {
```

yoomlam (changes requested):
Almost there!

svc-bie-kafka/src/main/java/gov/va/vro/services/bie/config/MessageExchangeConfig.java
tejans24: `topicNames` already gets values from `ContentionEvent`

yoomlam: But it's using `bieProperties` which is not where the topics are.

yoomlam: This code should be iterating over the Kafka topics and MQ exchange names provided by `ContentionEvent`. `bieProperties` shouldn't be used anywhere except in `ContentionEvent`.

chengjie8 (approved):

tejans24:
created an issue here: https://github.com/department-of-veterans-affairs/abd-vro/issues/1986

db-init/src/main/resources/database/migrations/V1.2__BIE_Contention_Event_Recreate_Table.sql
```diff
@@ -0,0 +1,38 @@
...
+    contention_type_code VARCHAR(255),
+
+    -- Timestamp when the record was created in the database
+    created_at TIMESTAMP NOT NULL,
+
+    -- Timestamp when the record was last updated in the database
+    updated_at TIMESTAMP NOT NULL,
+
+    PRIMARY KEY(id)
+);
```
msnwatson, line 38: nit: newline

msnwatson (approved):
one nit

Issue #1682, opened 2023-06-12
Update Helm configs and Vault secrets with BIE credentials, and deploy to LHDI and test in each relevant env


dianagriffin:
**User Story**

As , I want , so that .

**Depends on**
- #1680 

**Context**
- Only the platform BIE Kafka client microservice needs these secrets (in order to authenticate)
- Deploy to LHDI and test in each relevant env

**Acceptance Criteria**

1. Example domain code is tested and deployable across VRO environments. 

**Notes about work**
- This ticket is part of epic #1642

chengjie8:
we need 

- truststore string (decoded from base64 string from vault)
- truststore password
- keystore string 
- keystore password

yoomlam:
I recommend you add the following keys to Vault under a new `VRO_SECRETS_BIE_KAFKA` folder:
```
BIE_KAFKA_TRUSTSTORE_BASE64
BIE_KAFKA_TRUSTSTORE_PASSWORD
BIE_KAFKA_KEYSTORE_BASE64
BIE_KAFKA_KEYSTORE_PASSWORD
```

Refer to `VRO_SECRETS_BIP` Vault folder for similar examples.
Search for `_BASE64` in https://github.com/department-of-veterans-affairs/abd-vro/wiki/Secrets-Vault and read its usage.

chengjie8:
I am waiting for BIE team to provide answers to our kafka consumer group permission. Currently, our consumer group is blocked 

<img width="730" alt="image" src="https://github.com/department-of-veterans-affairs/abd-vro/assets/89034346/a0683d28-6b27-422d-bfb2-13f1f0627950">

PR #1879, opened 2023-08-02
Update Kafka Consumer code after integrating schema registry KafkaAvroDeserializer

chengjie8:
## What was the problem?
Currently we can only see byte array binary data in kafka topics. Adding confluent schema registry deserializer library along with ssl configurations, now the messages are decoded and become readable, which provide us the opportunity to update/modify the data prior to sending to rabbitmq queues

- #1682 

## How does this fix it?[^1]
- for non-local, ssl configurations for both kafka broker and schema registry are added in application-<env>.yaml files

## How to test this PR
since the connection to BIE kafka and schema registry requires VA network access, it cannot be verified in local. As soon as the svc-bie-kafka image is published and pod deployed to non-local environments, verify it in `vro-xample-workflows` log or datadog log

**_This PR depends on correct keystore/truststore base64 strings and passwords updated in Vault for each environment._**

### DIFF
--- a/svc-bie-kafka/src/main/java/gov/va/vro/services/bie/service/kafka/KafkaConsumer.java
+++ b/svc-bie-kafka/src/main/java/gov/va/vro/services/bie/service/kafka/KafkaConsumer.java
@@ -4,6 +4,7 @@ import gov.va.vro.services.bie.config.BieProperties;
 import gov.va.vro.services.bie.service.AmqpMessageSender;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
+import org.apache.avro.generic.GenericRecord;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.springframework.kafka.annotation.KafkaListener;
 import org.springframework.stereotype.Component;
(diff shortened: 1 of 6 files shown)


tejans24 (approved):
LGTM!

tejans24:
FYI: It looks like the PR has some code coverage related failures on certain files

svc-bie-kafka/build.gradle
```diff
@@ -4,7 +4,19 @@ plugins {
...
+    'gov/va/vro/services/bie/service/kafka/**'
+  ]
+}
+
+jacocoTestReport {
+  afterEvaluate {
+    classDirectories.setFrom(files(classDirectories.files.collect {
+      fileTree(dir: it, exclude: jacocoExclude)
+    }))
+  }
```
yoomlam, line 19: Why is this needed? I don't see it anywhere else in the code.

yoomlam (approved):
One question

svc-bie-kafka/build.gradle
chengjie8: This is to bypass the unit test on the KafkaConsumer class, since there's no good way to test this yet, this is just a temporary fix.

PR #1957, opened 2023-08-22
Topic name prefixes in non dev environment are incorrect

chengjie8:
## What was the problem?
Topic name prefixes in non dev environment are incorrect, need to be corrected based on [our wiki page](https://github.com/department-of-veterans-affairs/abd-vro/wiki/BIE-Kafka-Client#kafka-topic-prefixes)

## How does this fix it?[^1]
update non dev yaml files

## How to test this PR
- deploy to any non dev environment, and check `vro-xample-workflows` log to see if kafka events go through

### DIFF
--- a/svc-bie-kafka/src/main/resources/application-prod-test.yaml
+++ b/svc-bie-kafka/src/main/resources/application-prod-test.yaml
@@ -33,3 +33,13 @@ spring:
 kafka:
   topic:
     prefix: PRE
+## Specify bie properties
+bie:
+  kafka-topic-to-amqp-exchange-map:
+    # Map of entries where the keys are the kafka topics to which this app will subscribe. The value is the corresponding
+    # RabbitMQ exchange upon which the payload will be put. These values are separated by a colon ":" character.
+    PRE_CONTENTION_BIE_CONTENTION_ASSOCIATED_TO_CLAIM_V02: bie-events-contention-associated-to-claim
+    PRE_CONTENTION_BIE_CONTENTION_UPDATED_V02: bie-events-contention-updated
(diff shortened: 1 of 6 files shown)


svc-bie-kafka/src/main/resources/application-prod-test.yaml
```diff
@@ -33,3 +33,13 @@ spring:
 kafka:
   topic:
     prefix: PRE
```
yoomlam, line 35: It seems like this can be used to set the prefix, rather than copy-and-pasting the tweaked `kafka-topic-to-amqp-exchange-map` configuration to all yaml files.

chengjie8: This one is used in KafkaConsumer class. We can auto generate this based on env variable, but it's probably a good idea to address this in a separate PR, don't want to delay the prod verification.

yoomlam: How long will it take to do it correctly? We shouldn't add this tech debt if it can be implemented correctly in a short time. 

Based on this https://github.com/department-of-veterans-affairs/abd-vro/blob/8f9ce8d0cd2effa85796cc8e9a8e8ed21a071b02/svc-bie-kafka/src/main/java/gov/va/vro/services/bie/service/kafka/KafkaConsumer.java#L37-L44

It seems like this is what this prefix configuration was designed for.

yoomlam (changes requested):
Use `kafka.topic.prefix` in combination with `BieProperties` instead of copy-and-pasting to all yaml files. If you need help, I can implement it after the "Workshop" meeting today.

PR #2044, opened 2023-09-22
Created async hoppy client for request/response pattern

dfitchett:
## What was the problem?

Current implemention of the hoppy library did not include a way to make requests to a service connected via rabbitMQ.

- This is an effort to increase capability of hoppy library for VRO partners in pursuit of this task:
  - department-of-veterans-affairs/abd-vro/issues/2028
  - department-of-veterans-affairs/abd-vro/issues/2063

## How does this fix it?[^1]
This adds the ability to create async publishers, consumers, and an async client uses for blocking request/response. 

## How to test this PR
Pull the PR and install the hoppy library:
1. First create a python virtual environment
2. within `abd-vro` repo, run `cd shared/lib-hoppy`
3. run install command to install locally by running: `pip install .`
4. run `pytest`

<details>
<summary>Optionally, create a program like this to exercise the publisher and consumer:</summary>

```
 import asyncio
import functools
import logging
import sys

from async_consumer import AsyncConsumer
from async_publisher import AsyncPublisher
from hoppy_properties import ExchangeProperties, QueueProperties

logging.basicConfig(
    format="[%(asctime)s] %(levelname)-8s %(message)s",
    level=logging.INFO,
    datefmt="%Y-%m-%d %H:%M:%S",
    stream=sys.stdout,
)


def _on_reply(_channel, properties, delivery_tag, body):
    logging.info(f"Message Rx msg={body}")
    _channel.basic_ack(delivery_tag)


def publish_messages(publisher: AsyncPublisher, msg, count, delay):
    for i in range(count):
        loop.call_later(delay * (i + 1), functools.partial(publish, publisher.publish_message, i, msg))


def publish(publish_function, i, msg):
    publish_function(msg)
    logging.info(f"Message Tx #{i} msg={msg}")


async def close(clients: [AsyncPublisher | AsyncConsumer], delay):
    await asyncio.sleep(delay)
    for client in clients:
        client.stop()


if __name__ == "__main__":
    loop = asyncio.new_event_loop()

    exchange_props = ExchangeProperties("hey", passive_declare=False, auto_delete=False)
    queue_props = QueueProperties("hey", passive_declare=False, auto_delete=False)

    consumer = AsyncConsumer(exchange_properties=exchange_props, queue_properties=queue_props, routing_key="1",
                             reply_callback=_on_reply)
    logging.info(consumer.__dict__)
    consumer.connect(loop)

    publisher = AsyncPublisher(exchange_properties=exchange_props, queue_properties=queue_props, routing_key="1")
    publisher.connect(loop)
    publish_messages(publisher, "yo", 10, 1)

    # task = loop.create_task(close([consumer], 300))
    # task = loop.create_task(close([publisher], 12))
    task = loop.create_task(close([publisher, consumer], 12))
    loop.run_until_complete(task)
```
</details>

### DIFF
--- a/shared/lib-hoppy/hoppy/async_consumer.py
+++ b/shared/lib-hoppy/hoppy/async_consumer.py
@@ -6,13 +6,35 @@ from hoppy_properties import ExchangeProperties, QueueProperties
 
 
 class AsyncConsumer(BaseQueueClient):
+    """Creates an asynchronous consumer that can be used to consume messages from a queue and execute a callback upon
+    message received"""
 
     def __init__(self, config: [dict | None] = None,
                  exchange_properties: ExchangeProperties = ExchangeProperties(),
                  queue_properties: QueueProperties = QueueProperties(),
                  routing_key: str = '',
-                 reply_callback: Callable = None,
-                 prefetch_count: int = 1):
+                 prefetch_count: int = 1,
+                 reply_callback: Callable = None):
+        """
+        Creates this class
+
+        :param config: dict | None = None
+            collection of key value pairs used to create the RabbitMQ connection parameters (see pika.ConnectionParameters)
+            this config is merged with the default RABBITMQ_CONFIG
+        :param exchange_properties: ExchangeProperties
+            properties dictating how the exchange is declared
+        :param queue_properties: QueueProperties
+            properties dictating how the queue is declared
+        :param routing_key: str = ''
+            the routing key used to route messages to the queue
+        :param reply_callback: int = 1
+            number of messages to pull from server at a time, experiment with higher prefetch_count for higher consumer
+            throughput
+        :param prefetch_count: Callable = None
+            if present, this callback is called with the following parameters:
+                reply_callback(_channel, properties, basic_deliver.delivery_tag, body)
+        """
+
         super().__init__(Type.CONSUMER, config, exchange_properties, queue_properties, routing_key)
 
         self._consuming = False
@@ -28,8 +50,9 @@ class AsyncConsumer(BaseQueueClient):
         self._consuming = False
 
     def _ready(self):
-        """Executed when the exchange and queue are ready and this class can start the process of consuming.
+        """Called when the exchange and queue are ready and this class can start the process of consuming.
         Overrides super class abstract method."""
+
         logging.debug(f'event=specifyingQualityOfService '
                       f'client_type={self._client_type} '
                       f'prefetch_count={self._prefetch_count}')
@@ -38,6 +61,7 @@ class AsyncConsumer(BaseQueueClient):
     def _shut_down(self):
         """Called when the client is requested to stop.
         Overrides super class abstract method"""
+
         if self._consuming:
             self._stop_consuming()
 
@@ -94,6 +118,8 @@ class AsyncConsumer(BaseQueueClient):
                               f'err={e}')
 
     def acknowledge_message(self, properties, delivery_tag):
+        """Notifies the server that the received message is acknowledged"""
+
         logging.debug(f'event=ackedMessage '
                       f'client_type={self._client_type} '
                       f'delivery_tag={delivery_tag} '
@@ -101,6 +127,8 @@ class AsyncConsumer(BaseQueueClient):
         self._channel.basic_ack(delivery_tag)
 
     def reject_message(self, properties, delivery_tag, requeue=True):
+        """Notifies the server that the received message is rejected"""
+
         logging.debug(f'event=rejectedMessage '
                       f'client_type={self._client_type} '
                       f'delivery_tag={delivery_tag} '
--- a/shared/lib-hoppy/hoppy/async_consumer.py
+++ b/shared/lib-hoppy/hoppy/async_consumer.py
@@ -1,4 +1,3 @@
-import logging
 from typing import Callable
 
 from hoppy.base_queue_client import BaseQueueClient, ClientType
@@ -53,9 +52,8 @@ class AsyncConsumer(BaseQueueClient):
         """Called when the exchange and queue are ready and this class can start the process of consuming.
         Overrides super class abstract method."""
 
-        logging.debug(f'event=specifyingQualityOfService '
-                      f'client_type={self._client_type} '
-                      f'prefetch_count={self._prefetch_count}')
+        self._debug('specifyingQualityOfService',
+                    prefetch_count=self._prefetch_count)
         self._channel.basic_qos(prefetch_count=self._prefetch_count, callback=self._on_basic_qos_ok)
 
     def _shut_down(self):
@@ -66,45 +64,39 @@ class AsyncConsumer(BaseQueueClient):
             self._stop_consuming()
 
     def _on_basic_qos_ok(self, _unused_frame):
-        logging.debug(f'event=specifiedQualityOfService '
-                      f'client_type={self._client_type} '
-                      f'prefetch_count={self._prefetch_count}')
+        self._debug('specifiedQualityOfService',
+                    prefetch_count=self._prefetch_count)
         self._start_consuming()
 
     def _start_consuming(self):
         self._channel.add_on_cancel_callback(self._on_consumer_cancelled)
         self._consumer_tag = self._channel.basic_consume(self.queue_name, self._on_message)
         self._consuming = True
-        logging.debug(f'event=startConsuming '
-                      f'client_type={self._client_type} '
-                      f'consumer_tag={self._consumer_tag}')
+        self._debug('startConsuming',
+                    consumer_tag=self._consumer_tag)
 
     def _on_consumer_cancelled(self, method_frame):
-        logging.debug(f'event=serverCancelledConsumer '
-                      f'client_type={self._client_type} '
-                      f'method_frame={method_frame}')
+        self._debug('serverCancelledConsumer',
+                    method_frame=method_frame)
         super()._close_channel()
 
     def _stop_consuming(self):
         if self._channel:
-            logging.debug(f'event=stoppingConsuming '
-                          f'client_type={self._client_type} '
-                          f'consumer_tag={self._consumer_tag}')
+            self._debug('stoppingConsuming',
+                        consumer_tag=self._consumer_tag)
             self._channel.basic_cancel(self._consumer_tag, self._on_cancel_ok)
 
     def _on_cancel_ok(self, _unused_frame):
         self._consuming = False
-        logging.debug(f'event=stoppedConsuming '
-                      f'client_type={self._client_type} '
-                      f'consumer_tag={self._consumer_tag}')
+        self._debug('stoppedConsuming',
+                    consumer_tag=self._consumer_tag)
         self._close_channel()
 
     def _on_message(self, _unused_channel, basic_deliver, properties, body):
-        logging.debug(f'event=receivedMessage '
-                      f'client_type={self._client_type} '
-                      f'app_id={properties.app_id} '
-                      f'delivery_tag={basic_deliver.delivery_tag} '
-                      f'correlation_id={properties.correlation_id}')
+        self._debug('receivedMessage',
+                    app_id=properties.app_id,
+                    delivery_tag=basic_deliver.delivery_tag,
+                    correlation_id=properties.correlation_id)
         if self.reply_callback is not None:
             try:
                 self.reply_callback(self._channel,
@@ -112,26 +104,23 @@ class AsyncConsumer(BaseQueueClient):
                                     basic_deliver.delivery_tag,
                                     body)
             except Exception as e:
-                logging.error(f'event=couldNotCallConsumeCallback '
-                              f'client_type={self._client_type} '
-                              f'callback={self.reply_callback} '
-                              f'err={e}')
+                self._error('couldNotCallConsumeCallback',
+                            err=e,
+                            callback=self.reply_callback)
 
     def acknowledge_message(self, properties, delivery_tag):
         """Notifies the server that the received message is acknowledged"""
 
-        logging.debug(f'event=ackedMessage '
-                      f'client_type={self._client_type} '
-                      f'delivery_tag={delivery_tag} '
-                      f'correlation_id={properties.correlation_id}')
+        self._debug('ackedMessage',
+                    delivery_tag=delivery_tag,
+                    correlation_id=properties.correlation_id)
         self._channel.basic_ack(delivery_tag)
 
     def reject_message(self, properties, delivery_tag, requeue=True):
         """Notifies the server that the received message is rejected"""
 
-        logging.debug(f'event=rejectedMessage '
-                      f'client_type={self._client_type} '
-                      f'delivery_tag={delivery_tag} '
-                      f'correlation_id={properties.correlation_id} '
-                      f'requeue={requeue}')
+        self._debug('rejectedMessage',
+                    delivery_tag=delivery_tag,
+                    correlation_id=properties.correlation_id,
+                    requeue=requeue)
         self._channel.basic_reject(delivery_tag, requeue)
--- a/shared/lib-hoppy/hoppy/async_consumer.py
+++ b/shared/lib-hoppy/hoppy/async_consumer.py
@@ -1,0 +1,211 @@
+import asyncio
+import functools
+import logging
+import time
+
+import pika
+from pika.adapters.asyncio_connection import AsyncioConnection
+
+
+class AsyncConsumer(object):
+    def __init__(self, exchange, exchange_type, queue, routing_key,
+                 connection_parameters: pika.ConnectionParameters,
+                 reply_callback):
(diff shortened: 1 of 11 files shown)


shared/lib-hoppy/hoppy/async_consumer.py
```diff
@@ -0,0 +1,195 @@
...
+
+    def on_message(self, _unused_channel, basic_deliver, properties, body):
+        logging.debug(f'Consumer - Received message # {basic_deliver.delivery_tag} from {properties.app_id}: {body}')
+        self.reply_callback(self._channel,
+                            properties,
+                            basic_deliver.delivery_tag,
+                            body)
+
+    def acknowledge_message(self, delivery_tag):
+        logging.debug('Consumer - Acknowledging message %s', delivery_tag)
```
chengjie8, line 166: what is the reason for using .debug, should we use .info instead? Just curious

dfitchett: I set it to default because these classes can generate tons of logs, so i figure we can change the log level if there's an issue that is noticed.

shared/lib-hoppy/hoppy/async_consumer.py
```diff
@@ -0,0 +1,210 @@
...
+
+    def on_connection_closed(self, _unused_connection, reason):
+        self._channel = None
+        if self._stopping:
+            self._connection.ioloop.stop()
+        else:
+            logging.warning(f'Consumer - Connection closed, reason: {reason}')
+            self.reconnect()
+
+    def reconnect(self):
```
tejans24, line 97: would this cause a loop if the connection keeps failing?

tejans24: Does this method need exception handler?

shared/lib-hoppy/hoppy/async_consumer.py
```diff
@@ -0,0 +1,210 @@
...
+        self._channel.add_on_cancel_callback(self.on_consumer_cancelled)
+        self._consumer_tag = self._channel.basic_consume(self.queue_name, self.on_message)
+        self._consuming = True
+
+    def on_consumer_cancelled(self, method_frame):
+        logging.debug(f'Consumer - Consumer was cancelled remotely, shutting down: {method_frame}')
+        if self._channel:
+            self._channel.close()
+
+    def on_message(self, _unused_channel, basic_deliver, properties, body):
```
tejans24, line 173: what happens when reply_callback method fails? Do we need another error handler for this?

shared/lib-hoppy/hoppy/async_publisher.py
```diff
@@ -0,0 +1,217 @@
...
+        else:
+            logging.warning(f'Publisher - Connection closed, reason: {reason}')
+            self.reconnect()
+
+    def reconnect(self):
+        self.stop()
+        reconnect_delay = self._get_reconnect_delay()
+        logging.warning('Reconnecting after %d seconds', reconnect_delay)
+        time.sleep(reconnect_delay)
+        self.connect(self._loop)
```
tejans24, line 95: can some of the repeated code be shared across the two classes?

shared/lib-hoppy/hoppy/async_consumer.py
```diff
@@ -0,0 +1,210 @@
...
+
+        self._loop = None
+        self._connection = None
+        self._channel = None
+        self._consumer_tag = None
+        self._max_reconnect_delay = self.config.get('max_reconnect_delay', 30)
+
+        # In production, experiment with higher prefetch values
+        # for higher consumer throughput
+        self._prefetch_count = 1
```
tejans24, line 46: Magic number might be better as constant to provide better description and context

tejans24 (approved):
Overall LGTM! Adding a few comments.

shared/lib-hoppy/hoppy/async_consumer.py
dfitchett: 1: yes it will cause an infinite loop of retrying to connect, but this is desired behavior. 
2: This shouldn't need an exception handler because if the connection cannot be made, it will trigger the `_on_connection_open_error`.

shared/lib-hoppy/hoppy/async_consumer.py
dfitchett: This would kill the connection prompting a reconnect. I'll add in a try:except to catch it and log an error instead of dumping the connection.

tejans24 (approved):
LGTM! Thanks for making the updates.

shared/lib-hoppy/hoppy/async_consumer.py
```diff
@@ -0,0 +1,137 @@
...
+                      f'client_type={self._client_type} '
+                      f'consumer_tag={self._consumer_tag}')
+        self._close_channel()
+
+    def _on_message(self, _unused_channel, basic_deliver, properties, body):
+        logging.debug(f'event=receivedMessage '
+                      f'client_type={self._client_type} '
+                      f'app_id={properties.app_id} '
+                      f'delivery_tag={basic_deliver.delivery_tag} '
+                      f'correlation_id={properties.correlation_id}')
```
nanotone, line 107: These logging calls can get pretty verbose (just randomly picking on this one here).

I'll leave this up to your judgment, but one possible improvement might be to define a method that takes `**kwargs` (or something like that) and unpacks them into the string. So it could at least be a little prettier to read, like

```python
self.debug(event='receivedMessage',
           client_type=self._client_type,
           app_id=properties.app_id,
...
```

And if client_type almost always gets logged, that could just get rolled into `self.debug()`? Lotta ways this could go.

shared/lib-hoppy/hoppy/async_consumer.py
```diff
@@ -0,0 +1,137 @@
...
+            try:
+                self.reply_callback(self._channel,
+                                    properties,
+                                    basic_deliver.delivery_tag,
+                                    body)
+            except Exception as e:
+                logging.error(f'event=couldNotCallConsumeCallback '
+                              f'client_type={self._client_type} '
+                              f'callback={self.reply_callback} '
+                              f'err={e}')
```
nanotone, line 118: When converting Python exceptions to strings, you generally want to use `repr(e)` or `{e!r}`, instead of `str(e)` or `{e}`, because the repr is a more faithful representation of the argument. For example,

```python
try: {}[123]
except Exception as e: print(str(e), repr(e))
```

will print `123` as the str and `KeyError(123)` as the repr.

shared/lib-hoppy/hoppy/async_hoppy_client.py
```diff
@@ -0,0 +1,204 @@
...
+                    await asyncio.sleep(0)
+                elif time.time() - wait_for_response_time >= self.max_latency:
+                    logging.warning(f"event=requestError "
+                                    f"client={self.name} "
+                                    f"id={request_id} "
+                                    f"queue={self.request_queue_properties.name} "
+                                    f"correlation_id={correlation_id} "
+                                    f"error='{TIMED_OUT}'")
+                    self._terminate_correlation_id(correlation_id)
+                    raise ResponseException(message=TIMED_OUT)
```
nanotone, line 120: This method is pretty long! What about breaking it into a publish part, a handle-response part, and a wait-for-response part?

shared/lib-hoppy/hoppy/async_hoppy_client.py
```diff
@@ -0,0 +1,204 @@
...
+                    logging.info(f"event=requestCompleted "
+                                 f"client={self.name} "
+                                 f"id={request_id} "
+                                 f"queue={self.request_queue_properties.name} "
+                                 f"correlation_id={correlation_id}")
+                    return response
+            else:
+                if wait_for_response_time is None:
+                    wait_for_response_time = time.time()
+                    await asyncio.sleep(0)
```
nanotone, line 111: I'm worried there might be a scary coroutine bug here -- `await asyncio.sleep(0)` only gets called the first time through the loop. In the off-chance that no response arrives before awaking, won't the method just keep spinning hard without yielding, until timeout?

shared/lib-hoppy/hoppy/async_hoppy_client.py
```diff
@@ -0,0 +1,204 @@
...
+        self.async_consumer.stop()
+
+    async def make_request(self, request_id, body):
+        correlation_id = str(uuid.uuid4())
+        logging.info(
+            f"event=requestStarted id={request_id} "
+            f"client={self.name} "
+            f"queue={self.request_queue_properties.name} "
+            f"correlation_id={correlation_id}")
+        self.responses[correlation_id] = None
```
nanotone, line 71: It took me a while to realize that this dictionary is actually just `AsyncHoppyClient.responses`. Is there a reason to make it a class-level dictionary, rather than initiate an instance-level in `__init__`?

And my second question is, what is the benefit of inserting a None at the beginning of the request here, only to replace it with the response (or a JSON error) later?

shared/lib-hoppy/hoppy/base_queue_client.py
```diff
@@ -0,0 +1,264 @@
...
+from pika import ConnectionParameters, PlainCredentials
+from pika.adapters.asyncio_connection import AsyncioConnection
+
+
+class ClientType(str, Enum):
+    CONSUMER = "CONSUMER"
+    PUBLISHER = "PUBLISHER"
+
+    def __str__(self):
+        return self.value
```
nanotone, line 17: Usually enums in Python are defined with a one-liner:

```python
ClientType = Enum('ClientType', ["CONSUMER", "PRODUCER"])

# and then you use it like
ct = ClientType.CONSUMER
print(ct) # prints ClientType.CONSUMER
print(ct.name) # prints CONSUMER
```

shared/lib-hoppy/hoppy/async_hoppy_client.py
```diff
@@ -0,0 +1,204 @@
...
+                                    f"id={request_id} "
+                                    f"queue={self.request_queue_properties.name} "
+                                    f"correlation_id={correlation_id} "
+                                    f"error='{TIMED_OUT}'")
+                    self._terminate_correlation_id(correlation_id)
+                    raise ResponseException(message=TIMED_OUT)
+
+    def _terminate_correlation_id(self, correlation_id):
+        if correlation_id in self.responses.keys():
+            del self.responses[correlation_id]
```
nanotone, line 124: There's an idiomatic way to do these two lines, a.k.a. removing a key safely, in one go:

```python
self.responses.pop(correlation_id, None)
```

shared/lib-hoppy/hoppy/async_hoppy_client.py
```diff
@@ -0,0 +1,204 @@
...
+                self.responses[cor_id] = response
+                self.async_consumer.acknowledge_message(properties, delivery_tag)
+
+            except json.JSONDecodeError as e:
+                self.responses[cor_id] = e
+                self.async_consumer.reject_message(properties, delivery_tag, requeue=False)
+
+            return
+
+        if not cor_id:
```
nanotone, line 144: This seems like a more basic short-circuit than the one above, so maybe do this one first?

shared/lib-hoppy/hoppy/async_hoppy_client.py
```diff
@@ -0,0 +1,204 @@
...
+    def _terminate_correlation_id(self, correlation_id):
+        if correlation_id in self.responses.keys():
+            del self.responses[correlation_id]
+        if correlation_id in self.rejected.keys():
+            del self.rejected[correlation_id]
+
+    def _on_reply(self, _channel, properties, delivery_tag, body):
+        cor_id = properties.correlation_id
+
+        if cor_id and cor_id in self.responses.keys():
```
nanotone, line 131: A dictionary is iterable and membership-checkable as its keyset, so Python prefers just `if cor_id in self.responses:`

shared/lib-hoppy/hoppy/async_hoppy_client.py
```diff
@@ -0,0 +1,204 @@
...
+    async def make_request(self, request_id, body):
+        attempt = 0
+        while attempt < self.max_retries:
+            try:
+                response = await super().make_request(request_id, body)
+                if response is not None:
+                    return response
+            except ResponseException:
+                attempt += 1
+                continue
```
nanotone, line 198: This `continue` is extraneous 😉

shared/lib-hoppy/hoppy/async_publisher.py
```diff
@@ -0,0 +1,111 @@
...
+            self._acked += 1
+        elif confirmation_type == 'nack':
+            self._nacked += 1
+        elif confirmation_type == 'reject':
+            self._rejected += 1
+
+        del self._deliveries[delivery_tag]
+
+        if ack_multiple:
+            for tmp_tag in list(self._deliveries.keys()):
```
nanotone, line 79: Similar comment as before; you can just do `list(self._deliveries)`

nanotone (changes requested):
This is monumental! 👏 

I wrote down just a couple concerns, and then a slew of tiny Python style nits.

shared/lib-hoppy/hoppy/async_hoppy_client.py
dfitchett: Ooo good catch. 🎯

shared/lib-hoppy/hoppy/async_hoppy_client.py
dfitchett: Q1: No reason! Makes sense to make it instance level. 
Q2: I need to insert a value of None for the correlation_id in the responses at the beginning of the request because the client needs to keep track of the correlation_id's for which it is responsible. If we had two clients both listening to the same queue for replies, we don't want it processing correlation_ids for requests it didnt send out.

shared/lib-hoppy/hoppy/async_hoppy_client.py
```diff
@@ -0,0 +1,211 @@
...
+
+        self.publish(body, correlation_id, request_id)
+
+        wait_for_response_time = time.time()
+        while True:
+            response = self.responses.get(correlation_id)
+            if response is not None:
+                return self.handle_response(request_id, correlation_id, response)
+            else:
+                self.wait_for_response(correlation_id,
```
nanotone, line 82: This method is way more readable now! I would just suggest renaming this call to something other than `wait_for_response`, since it's really just checking whether the timeout has been reached, and all the actual waiting happens in the await call below.

shared/lib-hoppy/hoppy/base_queue_client.py
```diff
@@ -0,0 +1,273 @@
...
+                    exchange=self.exchange_name,
+                    routing_key=self.routing_key)
+        self._ready()
+
+    @staticmethod
+    def __kwarg_str(**kwargs):
+        msg = ''
+        for k, v in kwargs.items():
+            msg += f'{k}={v} '
+        return msg.strip()
```
nanotone, line 257: Nice, nice. Let me offer you some more sugar:

```python
return ' '.join(f'{k}={v}' for k, v in kwargs.items())
```

shared/lib-hoppy/hoppy/base_queue_client.py
```diff
@@ -0,0 +1,273 @@
...
+    def _info(self, event, **kwargs):
+        msg = f'event={event} client_type={self._client_type.name} {self.__kwarg_str(**kwargs)}'
+        logging.info(msg)
+
+    def _warning(self, event, **kwargs):
+        msg = f'event={event} client_type={self._client_type.name} {self.__kwarg_str(**kwargs)}'
+        logging.warning(msg)
+
+    def _error(self, event, error=None, **kwargs):
+        msg = f'event={event} client_type={self._client_type.name} err={repr(error)} {self.__kwarg_str(**kwargs)}'
```
nanotone, line 272: Fstrings have a shorthand for `err={repr(error)}` -- `err={error!r}`

shared/lib-hoppy/hoppy/async_hoppy_client.py
```diff
@@ -0,0 +1,211 @@
...
+                                                                 content_type="application/json",
+                                                                 reply_to=self.reply_queue_properties.name,
+                                                                 correlation_id=correlation_id))
+        except Exception as e:
+            logging.warning(f"event=requestError "
+                            f"client={self.name} "
+                            f"id={request_id} "
+                            f"queue={self.request_queue_properties.name} "
+                            f"correlation_id={correlation_id} "
+                            f"error='{UNEXPECTED_PUBLISH_ERROR}: {e}'")
```
nanotone, line 100: Other than a suggestion to replace `{e}` with `{e!r}`, I'll just leave you with a nagging feeling that there might still be a better way to make these logging statements nicer to read, which would probably involve a bigger overhaul than this PR should have to deal with 😛

nanotone (approved):
Looking really close to being done! Giving this an approval, since my remaining suggestions are pretty straightforward changes.

PR #2211, opened 2023-11-09
fix svc-bip-api service unit and integration test

chengjie8:
## What was the problem?
`svc-bip-api` service integration test github check was failing constantly on all PRs

- https://dsva.slack.com/archives/C04QLHM9LR0/p1699890846994409


.github/workflows/svc-bip-api-integration-test.yml
```diff
@@ -44,14 +44,10 @@ jobs:
           export -p | sed 's/declare -x //'
 
           ./gradlew :dockerComposeUp
-
-          ./gradlew -p mocks docker
-
+          ./gradlew -p mocks :build
```
dfitchett, line 47: I think you could also have `:mock-bip-claims-api:build` here instead!

svc-bip-api/build.gradle
```diff
@@ -6,8 +6,10 @@ plugins {
 
 ext {
   jacocoExclude = ['**/*model*']
+  jacoco_minimum_coverage = 0.5
```
dfitchett, line 9: Alternatively, we can just ignore the entire config package: `jacocoExclude = ['**/*model*', '**/config']`. No need to test the spring config files.

svc-bip-api/src/main/resources/application-test.yml
```diff
@@ -35,15 +35,14 @@ lhAPIProvider:
   validateToken: "NO"
 
 bip:
-  claimBaseUrl: localhost
+  claimBaseUrl: ${BIP_CLAIM_URL}
```
dfitchett, line 38: I think this still needs to be `localhost:20300` because locally, when `source scripts/setenv.sh` sets the `BIP_CLAIM_URL` to `mock-bip-claims-api:2030` which is great if your running the integration test within `vro_intranet`. But the integration test isnt, so its unable to resolve that hostname.

dfitchett (changes requested):

svc-bip-api/src/main/resources/application-test.yml
dfitchett: Also, if this isn't the case, I wonder whats different between our setups? 🤔

chengjie8: I think we should update and set  `BIP_CLAIM_URL` value to `localhost:20300` in our local secret folder

.github/workflows/svc-bip-api-integration-test.yml
chengjie8: added

.github/workflows/svc-bip-api-integration-test.yml
```diff
@@ -44,14 +44,13 @@ jobs:
           export -p | sed 's/declare -x //'
 
           ./gradlew :dockerComposeUp
+          ./gradlew -p mocks :build
```
dfitchett, line 47: Looks like this guy is still hanging around

dfitchett (changes requested):

svc-bip-api/build.gradle
chengjie8: Sounds good! I will exclude the config folder and remove the minimum coverage %

.github/workflows/svc-bip-api-integration-test.yml
chengjie8: addressed

.github/workflows/svc-bip-api-integration-test.yml
```diff
@@ -49,9 +49,7 @@ jobs:
 
           ./gradlew -p mocks :dockerComposeUp
 
-          ./gradlew :domain-xample:dockerComposeUp
           ./gradlew :app:dockerComposeUp
```
dfitchett, line 52: `./gradlew :app:dockerComposeUp` this brings up the `svc-bip-api` and i believe that it might be stealing the responses from the integration tests calls or preventing the calls from being made to the mock service at all. When i run through this locally, it fails all the tests. But if i remove this line, the integration tests pass.

dfitchett (approved):
As long as its passing! 😅

tejans24 (approved):

Issue #2346, opened 2023-12-12
resolve svc-bie-kafka service ClmntTxt field exception


chengjie8:
With the latest code changeand deployment to dev. `svc-bie-kafka` service is showing this exception in the k8s pod log
```
g.v.v.s.bie.service.kafka.KafkaConsumer  : Topic name: TST_CONTENTION_BIE_CONTENTION_ASSOCIATED_TO_CLAIM_V02
g.v.v.s.bie.service.kafka.KafkaConsumer  : Exception occurred while processing message: Not a valid schema field: ClmntTxt
```

AC:
- Confirm with the BIE team regarding the `ClmntTxt` field in the `ContentionAssociatedToClaim` topic schema.
- Ensure no more exceptions associated with the `ClmntTxt` field show up in the development log.
- Verify that Kafka data is saved correctly in the RDS.

PR #2368, opened 2023-12-16
Introduce annotations to handle the fields exclusive to only certain kafka topic names

chengjie8:
## What was the problem?
Currently, we try to fetch all field values in `BieMessagePayload` class from `GenericRecord`. Due to some fields that are not commonly shared from all kafka topics we listen to, it causes exceptions when the `GenericRecord` does not have the field value available.

- #2346 

## How does this fix it?[^1]
Introduced two new annotations that support the definition of exclusive fields for Kafka topic event names. This approach will reduce the amount of repetitive and tedious code required when transferring data from GenericRecord to our own `BieMessagePayload` objects. Most of the heavy lifting is now handled by the `BieMessageUtils` class.
- `@Ignore` - mark fields that should be ignored when fetching from a GenericRecord.
- `@TargetEvents` - is used to define a field that is exclusive to certain contention classification events. Valid values are derived from the topic names in the ContentionEvent enum.

## How to test this PR
- `source scripts/kafka-service.sh`
- `./gradlew svc-bie-kafka:test --rerun `
- `./gradlew svc-bie-kafka:integrationTest --rerun`

### DIFF
--- a/svc-bie-kafka/src/main/java/gov/va/vro/services/bie/utils/BieMessageUtils.java
+++ b/svc-bie-kafka/src/main/java/gov/va/vro/services/bie/utils/BieMessageUtils.java
@@ -11,13 +11,13 @@ import org.springframework.util.StringUtils;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.util.Arrays;
-import java.util.HashSet;
 import java.util.Set;
 import java.util.stream.Collectors;
 
 @Slf4j
 public class BieMessageUtils {
-  public static BieMessagePayload processBieMessagePayloadFields(ContentionEvent contentionEvent, GenericRecord genericRecord) {
+  public static BieMessagePayload processBieMessagePayloadFields(
+      ContentionEvent contentionEvent, GenericRecord genericRecord) {
     BieMessagePayload payload = BieMessagePayload.builder().status(200).build();
 
     for (Field field : BieMessagePayload.class.getDeclaredFields()) {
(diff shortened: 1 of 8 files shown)


svc-bie-kafka/src/main/java/gov/va/vro/services/bie/utils/BieMessageUtils.java
```diff
@@ -1,65 +1,88 @@
...
-          log.warn(
-              "Type mismatch for field '{}'. Expected type: '{}', Actual value: '{}'",
-              fieldName,
-              field.getType().getSimpleName(),
-              value);
+        boolean matchedTopicName =
+            Arrays.stream(annotationValues).anyMatch(v -> v.equals(contentionEvent.getTopicName()));
+
+        if (matchedTopicName) {
+          invokeSetterMethod(payload, field, genericRecord);
```
agile-josiah, line 43: Is the if statement still needed? It seems like the if else is calling the same method and everything else is inconsequential/for logging.

agile-josiah (changes requested):
Other than the if statement LGTM

svc-bie-kafka/src/main/java/gov/va/vro/services/bie/utils/BieMessageUtils.java
agile-josiah: Ah that makes sense. LGTM then

agile-josiah (approved):

tejans24 (approved):
LGTM

Issue #2398, opened 2023-12-28
Implement K8s liveness and readiness checks on `svc-bip-api` and `svc-lighthouse-api`


Ponnia-M:
**Background**
While we have Docker health checks setup for all of our Docker containers, these commands are not automatically leveraged by K8s in our deployments to LHDI remote environments.

**User Story**

As an administrator of the VRO platform, I want to ensure that we maintain the highest level of visibility into the health and readiness of platform and partner team services to ensure operational resiliency and reliability.

**Acceptance Criteria**

1. Deployment specs through Helm have been confirmed to contain liveness and readiness checks included in the spec, including implementation if they are missing.
2. New ticket is created to verify liveness/readiness checks exist for ALL deployments (w/ exception of partner teams).

**Notes about work**

- https://github.com/department-of-veterans-affairs/abd-vro/blob/develop/helm/api-gateway/templates/deployment.yaml#L35 for a code example of health and readiness checks in Helm deployments
- https://github.com/department-of-veterans-affairs/abd-vro/blob/develop/.github/workflows/bie-kafka-end2end-test.yml for code example of how to update/create readiness URLs.

Ponnia-M:
K8s liveness and readiness have been implemented to both service containers. Integration test for bip-api service was successful. In addition, I checked the readiness URL for bip-api and its up. Used Lens to confirm implementation of liveness and readiness for svc-lighthouse-api.

PR #2402, opened 2023-12-28
 Add actuator readiness URL for svc-bip-api 

Ponnia-M:
## What was the problem?
Add actuator health check to use the latest readiness urls

- #2398 

## How does this fix it?[^1]
Creates an endpoint for readiness for svc-bip-api

## How to test this PR
- Run integration test

### DIFF
--- a/.github/workflows/svc-bip-api-integration-test.yml
+++ b/.github/workflows/svc-bip-api-integration-test.yml
@@ -66,6 +66,16 @@ jobs:
           timeout: 2000
           # Quit after 60 seconds
           retries: 30
+          
+      - name: 'Wait for svc-bip-api to be ready'
+        uses: nev7n/wait_for_response@v1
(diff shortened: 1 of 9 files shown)


msnwatson:
I'm almost wanting to hold off on implementing this for this service given the complexities still ongoing in getting the service functional. That's my bad for not calling that out earlier.  If it seems to be working, then we can move forward but if you run into any issues, I'd say we keep the PR open until the service gets into a good state. cc'ing the folks most involved @tejans24  @chengjie8  @nelsestu  @agile-josiah

msnwatson (commented):
see above, feel free to re-request if necessary

agile-josiah (approved):
LGTM

msnwatson (approved):
discussed in standup and confirmed with folks working on the BIP effort

PR #2404, opened 2023-12-29
Feature - endpoints for multi contention + vbms claim_id logging

lukey-luke:
## What was the problem?

 - Contention Classification `/classifier` endpoint only accepts single-issue claims.
 - Also, VA.gov claim id is collected from `/classifier`, but VBMS id for the claim is not saved or linked to the va.gov claim presently.

- https://github.com/department-of-veterans-affairs/vagov-claim-classification/issues/152

## How does this fix it?[^1]

 - added `/classifier/v2` with different input format, including `"contentions"` array
 - added second endpoint for linking claim ids `/claim-linker`
 
## Additional junk
 - moved some existing code to improve readability, I think?
 - do_get_classification() now handles logic for checking the lookup table, instead of the endpoints themselves
 - added `deprecated=True` to `/classifier` which greys it out in the documentation
 - renamed previous pydantic model from `Claim` to `FlattenedSingleIssueClaim` in order to be more explicit
 - added new pydantic models for inputs/outputs of `/classifier/v2`

## How to test this PR
- Tests added to `domain-cc/cc-app/tests/test_classifier_v2.py`
- `pytest`
- additionally, running the code locally, API can be tested manually via POST request

### DIFF
--- a/domain-cc/cc-app/src/python_src/api.py
+++ b/domain-cc/cc-app/src/python_src/api.py
@@ -14,6 +14,7 @@ from .pydantic_models import (
 from .util.brd_classification_codes import get_classification_name
 from .util.logging_dropdown_selections import build_logging_table
 from .util.lookup_table import ConditionDropdownLookupTable, DiagnosticCodeLookupTable
+from .util.sanitize import sanitize
 
 dc_lookup_table = DiagnosticCodeLookupTable()
 dropdown_lookup_table = ConditionDropdownLookupTable()
@@ -94,7 +95,6 @@ def do_get_classification(
 
 
 def generate_flattened_claim(contention_text, diagnostic_code, multi_contention_claim):
-    print(f"got dc: {diagnostic_code}")
     return FlattenedSingleIssueClaim(
         claim_id=multi_contention_claim.va_gov_claim_id,
         form526_submission_id=multi_contention_claim.va_gov_form526_submission_id,
@@ -109,7 +109,9 @@ def classifier_v1(
     claim: FlattenedSingleIssueClaim,
 ) -> Optional[PredictedClassification]:
     logging.info(
-        f"claim_id: {claim.va_gov_claim_id}, form526_submission_id: {claim.va_gov_form526_submission_id}"
+        sanitize(
+            f"claim_id: {claim.va_gov_claim_id}, form526_submission_id: {claim.va_gov_form526_submission_id}"
+        )
     )
     return do_get_classification(claim)
 
@@ -144,7 +146,9 @@ def classifier_v2(
         classifications.append(classification)
 
     logging.info(
-        f"claim_id: {multi_contention_claim.claim_id}, form526_submission_id: {multi_contention_claim.form526_submission_id}"
+        sanitize(
+            f"claim_id: {multi_contention_claim.claim_id}, form526_submission_id: {multi_contention_claim.form526_submission_id}"
+        )
     )
     logging.info(f"contention_log_info: {contention_log_info}")
 
--- a/domain-cc/cc-app/src/python_src/api.py
+++ b/domain-cc/cc-app/src/python_src/api.py
@@ -53,9 +53,6 @@ def get_health_status():
 def do_get_classification(
     claim: FlattenedSingleIssueClaim,
 ) -> Optional[PredictedClassification]:
-    logging.info(
-        f"claim_id: {claim.claim_id}, form526_submission_id: {claim.form526_submission_id}"
-    )
     classification_code = None
     if claim.claim_type == "claim_for_increase":
         logging.info(f"diagnostic code: {claim.diagnostic_code}")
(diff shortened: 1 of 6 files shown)


domain-cc/cc-app/src/python_src/pydantic_models.py
```diff
@@ -16,14 +18,38 @@ def check_dc_for_cfi(cls, values):
...
 
+class Contention(BaseModel):
+    contention_text: str
+    diagnostic_code: Optional[int]  # only required for claim_type: "claim_for_increase"
+
+
+class VaGovClaim(BaseModel):
+    claim_id: int
+    form526_submission_id: int
+    claim_type: str  # "claim_for_increase" or "new"
```
raudabaugh, line 35: Should `claim_type` be an enum?

domain-cc/cc-app/src/python_src/pydantic_models.py
```diff
@@ -1,9 +1,11 @@
 from typing import Optional
 
+from fastapi import HTTPException
 from pydantic import BaseModel, root_validator
+from pydantic.types import conlist
 
 
-class Claim(BaseModel):
+class FlattenedSingleIssueClaim(BaseModel):
     claim_id: int
```
raudabaugh, line 9: Does this refer to the VA.gov claim ID or the VBMS claim ID? If the former, how will the VBMS claim ID be obtained?

domain-cc/cc-app/src/python_src/api.py
```diff
@@ -88,3 +91,75 @@ def get_classification(claim: Claim) -> Optional[PredictedClassification]:
...
+def classifier_v1(
+    claim: FlattenedSingleIssueClaim,
+) -> Optional[PredictedClassification]:
+    logging.info(
+        f"claim_id: {claim.claim_id}, form526_submission_id: {claim.form526_submission_id}"
+    )
+    return do_get_classification(claim)
+
+
+@app.post("/classifier/v2")
```
raudabaugh, line 117: We may want to create a separate `api.py` module for `v2`, and/or have the version number before the controller name here, especially if we anticipate interdependencies between routes.

domain-cc/cc-app/src/python_src/api.py
```diff
@@ -88,3 +91,75 @@
...
+                "contention_text": contention_text_no_pii,
+                "is_in_dropdown": is_in_dropdown,
+                "is_mapped": classification is not None,
+                **(classification or {}),
+            }
+        )
+        classifications.append(classification)
+
+    logging.info(
+        f"claim_id: {multi_contention_claim.claim_id}, form526_submission_id: {multi_contention_claim.form526_submission_id}"
```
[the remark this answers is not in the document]
raudabaugh: Suggest borrowing the fix from EE codebase: https://github.com/department-of-veterans-affairs/abd-vro/blob/develop/domain-ee/ee-ep-merge-app/src/python_src/util/sanitizer.py

domain-cc/cc-app/src/python_src/api.py
```diff
@@ -88,3 +91,75 @@ def get_classification(claim: Claim) -> Optional[PredictedClassification]:
...
+        flattened_single_issue_claim = generate_flattened_claim(
+            contention_text, diagnostic_code, multi_contention_claim
+        )
+        classification = do_get_classification(flattened_single_issue_claim)
+
+        is_in_dropdown = contention_text.strip().lower() in dropdown_values
+        contention_text_no_pii = (
+            contention_text if is_in_dropdown else "Not in dropdown"
+        )
+        contention_log_info.append(
```
tyler-spangler6, line 136: Should this logging only be for new claims rather than claims for increase? Should the same log statements that are generated in the do_get_classification function be removed to avoid duplicate logs?

domain-cc/cc-app/src/python_src/pydantic_models.py
lukey-luke: I think for database models, makes a lot of sense to use an enum for fixed set of fields. 

For pydantic model which is indicative of inputs to the API in this case, do you think it makes sense for client to include
"claim_type": 1 on the request?

I feel that's less intuitive than the string indicating claim type from the client?

domain-cc/cc-app/src/python_src/pydantic_models.py
lukey-luke: this contains the vagov_claim_id, I will rename accordingly

domain-cc/cc-app/src/python_src/api.py
lukey-luke: Yes, we'd ideally remove that, but won't be switching over to new API immediately, so I'll try to figure out a way to preserve old logging formatting before switching over

domain-cc/cc-app/src/python_src/api.py
lukey-luke: alternatively, would moving 
```
@app.post("/classifier", deprecated=True)
def classifier_v1(
```
to a deprecated folder achieve the same purpose?

domain-cc/cc-app/src/python_src/pydantic_models.py
raudabaugh: Maybe a [StrEnum ](https://docs.python.org/3/library/enum.html#enum.StrEnum)would work well here?

lukey-luke:
Closing in favor of https://github.com/department-of-veterans-affairs/abd-vro/pull/2432 and https://github.com/department-of-veterans-affairs/abd-vro/pull/2437

Issue #2449, opened 2024-01-16
Test EP Merge App in dev


dfitchett:
# User Story
As a user of the EP Merge App, I want to have EP Merge App deployed to the dev environment so that I can perform acceptance testing and verify connectivity between VRO apps and downstream services.

## Acceptance Criteria
Verify connectivity in dev to VRO apps and downstream:
- [x] Postgres / RDS
- [x] RabbitMQ
- [x] BIP API
- [x] ~BGS API~ Not possible to get to this point in the workflow for merge requests in dev. BGS will be tested separately.

## Related
Blocked by issues with dev deployments of VRO's RabbitMQ, BIP, BGS

dfitchett:
Blocked in dev due to issue with deployment of rabbitmq. Also, need to change EP Merge App start hoppy clients in the background of startup task.

meganhicks:
@amylai-nava - I think we should move this to next Sprint.

amylai-va:
@meganhicks thanks for flagging. @dfitchett, are you still blocked in dev due to rabbitmq issues? if so, should we move this ticket out to the next sprint?

amylai-va:
2/1: make a new ticket to track BGS testing

PR #2507, opened 2024-01-25
EP Merge: Use POSTGRES_URL env variable to generate db URI

dfitchett:
## What was the problem?

The app was composing the database uri from `POSTGRES_USER, POSTGRES_PASSWORD, POSTGRES_HOST, POSTGRES_PORT, POSTGRES_DB` environment variables. However, `POSTGRES_HOST` and `POSTGRES_PORT` are not available in the deployed environment. Instead `POSTGRES_URL` is available and has the form `postgresql://host:port/db`.

- [slack](https://dsva.slack.com/archives/C01Q7979Z7D/p1706039452149679?thread_ts=1706036851.050139&cid=C01Q7979Z7D)
Discovered while testing EP Merge App in dev:
- #2449

## How does this fix it?[^1]

Using string manipulation to replace the initial portion of the `POSTGRES_URL` with `postgresql://user:password@`


domain-ee/ee-ep-merge-app/src/python_src/config.py
```diff
@@ -81,4 +81,5 @@ class ClientName(str, Enum):
 POSTGRES_PORT = os.environ.get("POSTGRES_PORT") or "5432"
 POSTGRES_DB = os.environ.get("POSTGRES_DB") or "vro"
 POSTGRES_SCHEMA = os.environ.get("POSTGRES_SCHEMA") or "claims"
-SQLALCHEMY_DATABASE_URI = f"postgresql://{POSTGRES_USER}:{POSTGRES_PASSWORD}@{POSTGRES_HOST}:{POSTGRES_PORT}/{POSTGRES_DB}"
+POSTGRES_URL = os.environ.get("POSTGRES_URL") or f"postgresql://{POSTGRES_HOST}:{POSTGRES_PORT}/{POSTGRES_DB}"
+SQLALCHEMY_DATABASE_URI = POSTGRES_URL.replace('postgresql://', f"postgresql://{POSTGRES_USER}:{POSTGRES_PASSWORD}@")
```
raudabaugh: I suggest checking that `POSTGRES_USER` is set before doing this replacement. It's possible that the credentials could someday get moved into `POSTGRES_URL` which would cause an error here.

dfitchett: Ah. It should be set to `vro_user` by default if the env variable doesn't exist. Are you suggesting removing that default value and then make the check to see if its not None?

nanotone: I think Sam is saying, the env var `$POSTGRES_URL` itself could already contain the user and password, in which case this line would produce an unusable value for `SQLALCHEMY_DATABASE_URI`. Doesn't matter what the default value is.

Easiest way is to let Python's `urllib.parse.urlparse` do the dirty work of string splitting.

raudabaugh (approved):

PR #2508, opened 2024-01-26
Bgs liveness

ponny-m:
## What was the problem?
Implementing K8s Liveness and Readiness on BGS-api (Ruby application)

Associated tickets or Slack threads:
- #2417 

## How does this fix it?[^1]
This configures K8s liveness and readiness for svc-bgs-api to ensure highest level of visibility and reliability into the health and readiness of platform and partner team services.

## How to test this PR
- Step 1
    - Use Lens to check Pod Descriptions for Readiness Probe Output
    
- Step 2
    - Simulate Probe Failure to test if the probes are working correctly. 
    - **For Liveness:** Disrupt connection between RabbitMQ and BGS-api 
    - **For Readiness:** Temporarily disrupt the connection to RabbitMQ and also disrupt BGS service and make it unavailable 
       or unresponsive. 
    - Check if BOTH liveness and readiness probes correctly identifies these. Pods will restart if Liveness probe fails while 
       pods will be marked as 'Not Ready' if Readiness probe fails.

Issue #2417, opened 2024-01-03
Investigate how to implement K8s Liveness and Readiness on svc-bgs-api


Ponnia-M:
We want to implement K8s liveness and readiness for all services (not including partner-team applications) since these probes help Kubernetes understand when our application is ready to serve traffic (readiness) and when it is functioning properly (liveness). All of our other services (JAVA applications) except svc-bgs-api (Ruby application) have this configured and the goal of this ticket is to research how to implement K8s liveness and readiness on a Ruby application. 

**Questions we want to answer**
1. How to configure K8s liveness and readiness probes for a serverless Ruby application?
2. How to create endpoints for svc-bgs-api that Kubernetes can use to perform the liveness and readiness checks?
3. How to test this locally using our integration testing strategy? 

**Acceptance Criteria**
Share findings with team. Create tickets for additional work that needs to be done.

PR #2520, opened 2024-01-31
Add Ruby method to check BGS-api availability (2417)

ponny-m:
## What was the problem?
Ruby method for check_bgs_availability not configured in bgs_client.rb file. This is how we plan on checking bgs-api availability and this information is used by the k8s readiness probe. 

- #2417 

How to test this PR
Follow steps as described in this PR: **https://github.com/department-of-veterans-affairs/abd-vro/pull/2508**

### DIFF
--- a/svc-bgs-api/src/lib/bgs_client.rb
+++ b/svc-bgs-api/src/lib/bgs_client.rb
@@ -69,6 +69,17 @@ class BgsClient
     end
   end
 
+   # Method to check BGS service availability. It makes use of the vro_participant_id method to check if BGS service is 
+  # reachable and responding. This method should return a participant ID if the service is available.
+  def check_bgs_availability
+    participant_id = vro_participant_id
+    !participant_id.nil? && participant_id.is_a?(String) && !participant_id.empty?
+  rescue StandardError => e
+    puts "BGS-api availability check failed: #{e.message}"
+    false
+    end
+  end
+
   def create_claim_notes(claim_id:, notes:)
     note_hashes = notes.map do |note|
       { claim_id: claim_id, txt: note, user_id: vro_participant_id }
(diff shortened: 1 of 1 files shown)


msnwatson (approved):

svc-bgs-api/src/lib/bgs_client.rb
```diff
@@ -69,6 +69,16 @@ def vro_participant_id
     end
   end
 
+   # Method to check BGS service availability. It makes use of the vro_participant_id method to check if BGS service is 
+  # reachable and responding. This method should return a participant ID if the service is available.
+  def check_bgs_availability
+    participant_id = vro_participant_id
+    !participant_id.nil? && participant_id.is_a?(String) && !participant_id.empty?
```
nanotone, line 76: Ruby style tries to avoid negations as well as explicit type-checks. So I would suggest that this line could just be `participant_id.present?` (and you could fold it back into the previous line as `vro_participant_id.present?`).

svc-bgs-api/src/lib/bgs_client.rb
```diff
@@ -69,6 +69,16 @@ def vro_participant_id
     end
   end
 
+   # Method to check BGS service availability. It makes use of the vro_participant_id method to check if BGS service is 
+  # reachable and responding. This method should return a participant ID if the service is available.
+  def check_bgs_availability
```
nanotone, line 74: Another thing about Ruby style: methods that return a boolean should be named as an English-ish question, including question mark. So perhaps `def bgs_available?` could work better here. (And you'd have to update it in the readiness script.)

nanotone (changes requested):
Just the couple minor suggestions about Ruby style, this looks good!

PR #2575, opened 2024-02-08
Update logs to fix misaligned coverage stats in datadog

lukey-luke:
## What was the problem?

Logs that are for CfI will sometimes log "lookup table match", so we don't have an accurate message to determine NEW vs CfI claims in datadog.

- https://github.com/department-of-veterans-affairs/vagov-claim-classification/issues/361

## How does this fix it?[^1]

- this adds a new logging statement per classifier request in order to provide a single log message to draw datadog graphs from

## How to test this PR
- `pytest`
- running api locally


domain-cc/cc-app/src/python_src/api.py
```diff
@@ -110,6 +137,7 @@ def get_classification(claim: Claim) -> Optional[PredictedClassification]:
         classification = None
 
     log_as_json({"classification": classification})
+    log_claim_stats(claim, PredictedClassification(**classification) if classification else None)
```
lukey-luke: I think it might make sense to use new logging statements to build more accurate graphs, then when old datadog charts are phased out and we're not having old queries lingering around in dashboard(s), deleting some of the other logging statements to clean up the code a bit

domain-cc/cc-app/src/python_src/api.py
```diff
@@ -82,6 +82,31 @@ def log_as_json(log: dict):
...
+    log_as_json({
+        "claim_id": sanitize_log(claim.claim_id),
+        "claim_type": sanitize_log(claim.claim_type),
+        "classification_code": classification_code,
+        "classification_name": classification_name,
+        "contention_text": sanitized_contention_text,
+        "diagnostic_code": sanitize_log(claim.diagnostic_code),
+        "form526_submission_id": sanitize_log(claim.form526_submission_id),
+        "is_in_dropdown": is_in_dropdown,
+        "is_lookup_table_match": classification_code is not None,
```
tyler-spangler6: I think we should add the contention text if it was in a lookup table even if it is not in the current dropdown option. in our current logs we use the already_mapped_text field to do this.

domain-cc/cc-app/src/python_src/api.py
tyler-spangler6: I think I am confused on this part. Can you explain the benefit of keeping the old log statements in there? Are you saying that we should set a timeframe to get rid of the old charts and delete the log statements at that point?

lukey-luke: I'm saying that if I delete the old log statements in this PR, current datadog dashboards will not receive new logs and not be updated until I update datadog

Are you suggesting that this is a worse alternative to just accepting loss of data in charts until switching over?

tyler-spangler6: That makes sense; i think we might just have to update the charts measuring metrics for is_in_dropdown to remove duplicates if they count both logs. But the other charts should be good. Thanks for explaining it!

tyler-spangler6 (approved):
Approved

domain-cc/cc-app/src/python_src/api.py
```diff
@@ -92,13 +120,15 @@ def get_classification(claim: Claim) -> Optional[PredictedClassification]:
...
-        log_as_json({"diagnostic code": sanitize_log(claim.diagnostic_code)})
         classification_code = dc_lookup_table.get(claim.diagnostic_code, None)
 
     if claim.contention_text and not classification_code:
         classification_code = dropdown_lookup_table.get(claim.contention_text, None)
 
+    if claim.claim_type == "new":
         log_lookup_table_match(classification_code, claim.contention_text)
+    else:
+        log_as_json({"diagnostic code": sanitize_log(claim.diagnostic_code)})
```
raudabaugh, line 131: Should this be `diagnostic_code`?

domain-cc/cc-app/src/python_src/api.py
```diff
@@ -82,6 +82,34 @@ def log_as_json(log: dict):
...
+    is_mapped_text = dropdown_lookup_table.get(claim.contention_text, None) is not None
+    sanitized_contention_text = contention_text if is_mapped_text else "unmapped contention text"
+
+    log_as_json(
+        {
+            "claim_id": sanitize_log(claim.claim_id),
+            "claim_type": sanitize_log(claim.claim_type),
+            "classification_code": classification_code,
+            "classification_name": classification_name,
+            "contention_text": sanitized_contention_text,
```
raudabaugh: Is `contention_text` the same as `lookup_table_match` in `log_lookup_table_match`? Maybe keeping this name consistent would make for less work on the dashboard side?

raudabaugh (approved):

domain-cc/cc-app/src/python_src/api.py
lukey-luke: yeah, I think we wanted to use diagnostic code as signal for CfI and dropdown match for new, but ended up logging above message for both, so now shifting to single log message per claim

PR #2654, opened 2024-02-21
Upgrade Libxml2 package to version 2.11.7-r0 or above

ponny-m:
##What was the problem?
During a recent Secrel run, a vulnerability with high severity came up and the remediation was to upgrade Libxml2 package to version 2.11.7-r0 or above. 

##How to test this PR
- Validate the results of a Secrel run

### DIFF
--- a/app/Dockerfile
+++ b/app/Dockerfile
@@ -1,11 +1,8 @@
 FROM eclipse-temurin:17-jre-alpine
 
 # hadolint ignore=DL3018
-RUN apk update && \
-    apk --no-cache add curl redis sudo gcompat && \
-    apk upgrade --available && \
-    apk add libxml2=2.11.7-r0 && \
-    rm -rf /var/cache/apk/*
+RUN apk add --no-cache curl redis sudo gcompat libxml2=2.11.7-r0 && \
+    apk upgrade --available 
 
 WORKDIR /app
 COPY set-env-secrets.src entrypoint-wrapper.sh docker-entry*.sh ./
--- a/app/Dockerfile
+++ b/app/Dockerfile
@@ -1,7 +1,7 @@
 FROM eclipse-temurin:17-jre-alpine
 
 # hadolint ignore=DL3018
-RUN apk add --no-cache curl redis sudo gcompat libxml2=2.11.7-r0 && \
+RUN apk add --no-cache curl redis sudo gcompat libxml2 && \
     apk upgrade --available 
 
 WORKDIR /app
--- a/app/Dockerfile
+++ b/app/Dockerfile
@@ -4,6 +4,7 @@ FROM eclipse-temurin:17-jre-alpine
 RUN apk update && \
     apk --no-cache add curl redis sudo gcompat && \
     apk upgrade --available && \
+    apk add libxml2=2.11.7-r0 && \
     rm -rf /var/cache/apk/*
 
 WORKDIR /app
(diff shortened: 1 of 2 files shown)


agile-josiah (approved):
LGTM, just have a comment.

app/Dockerfile
```diff
@@ -1,10 +1,8 @@
 FROM eclipse-temurin:17-jre-alpine
 
 # hadolint ignore=DL3018
-RUN apk update && \
-    apk --no-cache add curl redis sudo gcompat && \
-    apk upgrade --available && \
-    rm -rf /var/cache/apk/*
```
agile-josiah, line 7: Why is the rm -rf line removed?

nelsestu: I see that the intent of this change is to add the libxml2 library and while that sounds like a reasonable solution, there are at least three different issues with how that addition has been applied.  I'll list out the issues and then follow that with more explaination about why.
1. The "apk update &&" that starts the command off has been removed and shouldn't be
2. The remove cache line that Josiah mentioned has been removed and shouldn't be
3. The addition of libxml2 explicitly references a version number, and it is best to exclude the version number when possible. 

So if you rolled back to the original command and just appended libxml2 to the end of the list on line 5, I anticipate that to be the only change we want.

To share the explaination then... It is considered a best practice to remove the package manager's cache directory. Not only do any files left in the cache increase the overall size of the image, but the cache can also include package files that are not needed in the final image, or even intermediate build states that present new vulnerability potential. Thus, removing the package manager cache results in a smaller image and a reduced attack surface.

And with this best practice in mind, it also explains why we need to keep the apk update line first. Without the update, your base image will have nothing in its cache, and therefore nothing to add or update.

Lastly, I would prefer that we exclude using the explicit version number when referencing the library.  There can be cases where we want to pin the version number, but pinning without good reason for doing so creates a maintenance concern that leads future devs to wonder why this library is pinned to this specific version, and what they are going to break if they need to update it. We'd rather continue to receive any fixes that are deployed beyond 2.11.7 until we find an actual reason to pin it.

nelsestu (changes requested):

app/Dockerfile
ponny-m: Here is why I made those changes: 

1. I do see your concern of removing `apk update` since this could potentially lead to issues as it's responsible for updating the package index. However, when using `apk add --no-cache`, the necessity for a separate apk update command is negated since --no-cache fetches the index during the install process without storing it locally. 
2. In regards to not specifying the package version, you are right since we don't want to hard code versions to our docker image. So I will change this to: `libxml2`
3. Finally, though removing the package manager cache with this command line `rm -rf /var/cache/apk/*` is best practice for reducing Docker images sizes, the reason for removing this line is again back to the `apk add --no-cache` command since the cache is not preserved  and manually removing the cache directory becomes redundant at this point.

agile-josiah: LGTM, I think this is a more elegant solution as I don't think we want to retain our cache for other containers either, being able to build each image cleanly is good practice. Thanks for sharing :) LGTM

nelsestu: I stand corrected!  Thanks for the explaination Ponnia, I didn't realize that the "--no-cache" flag was such a direct and efficient replacement for the "update and remove" pattern that is so prevalent.  It makes me wonder how many code samples on the internet make the same mistake i did by missing the true meaning of "--no-cache". Thanks again!

nelsestu (approved):

PR #2644, opened 2024-02-20
EP Merge: Fixed state query parameter for get merge jobs endpoint

dfitchett:
## What was the problem?

Previous change to using auto() for enums affected ability for strings to be passed as query parameters and used for getting jobs from database based on state. 

## How does this fix it?[^1]
This fixes that issue and adds some integration tests to test get endpoints with actual database.

## How to test this PR
- pull code
- Unit Test:
  - from /domain-ee/ee-ep-merge-app run pytest
- Integration Tests:
  - run [base platform](https://github.com/department-of-veterans-affairs/abd-vro/wiki/Docker-Compose#platform-base)
  - from /domain-ee/ee-ep-merge-app run pytest ./integration
- End to End tests:
  - run this [CI](https://github.com/department-of-veterans-affairs/abd-vro/actions/workflows/ee-ep-merge-end-to-end.yml) action to verify successful integration test
  - run end to end tests locally:
    - `export BIP_CLAIM_URL=mock-bip-claims-api:20300`
    - run bgs and bip services and mocks:
      - `COMPOSE_PROFILES="bip,bgs" ./gradlew :app:dockerComposeUp`
      - `COMPOSE_PROFILES="bip,bgs" ./gradlew -p mocks :dockerComposeUp`
    - `./gradlew :domain-ee:ee-ep-merge-app:endToEndTest`

_Note:_ the integration tests must execute before the end to end tests, or you will need to clear the database between runs.

[^1]: [Pull-Requests guidelines](https://github.com/department-of-veterans-affairs/abd-vro/wiki/Pull-Requests). If PR is significant, update [Current Software State](https://github.com/department-of-veterans-affairs/abd-vro/wiki/Current-Software-State) wiki page.
[^secrel]: To check if a PR will succeed in the SecRel workflow, [test PRs in the SecRel pipeline](https://github.com/department-of-veterans-affairs/abd-vro-internal/wiki/Secure-Release-process#to-test-prs-in-the-secrel-pipeline).

PR #2683, opened 2024-02-27
Simplify EP Merge integration test sorter

nanotone:
## What was the problem?
`pytest_collection_modifyitems` had an unnecessarily written-out algorithm that could be collapsed to Python's built-in sort.

- Quick follow-up to #2644

## How does this fix it?
Use Python's built-in sort with a sort key that's either 0, 1, or 2 depending on what module the item came from. The sort is stable, meaning items within each of the 3 groups maintain their relative ordering.


dfitchett (approved):
Thats so much cleaner! 👍

PR #2695, opened 2024-02-29
EP Merge: Fixed issue with created_at being incorrectly set.

dfitchett:
## What was the problem?

`create_at` time was not being set correctly resulting in inaccurate duration times.

## How does this fix it?[^1]

Made create_at time initialized on instance construction. 

## How to test this PR
- pull code
- Unit Test:
  - from /domain-ee/ee-ep-merge-app run pytest
- Integration Tests:
  - run [base platform](https://github.com/department-of-veterans-affairs/abd-vro/wiki/Docker-Compose#platform-base)
  - from /domain-ee/ee-ep-merge-app run pytest ./integration
- End to End tests:
  - run this [CI](https://github.com/department-of-veterans-affairs/abd-vro/actions/workflows/ee-ep-merge-end-to-end.yml) action to verify successful integration test
  - run end to end tests locally:
    - `export BIP_CLAIM_URL=mock-bip-claims-api:20300`
    - run bgs and bip services and mocks:
      - `COMPOSE_PROFILES="bip,bgs" ./gradlew :app:dockerComposeUp`
      - `COMPOSE_PROFILES="bip,bgs" ./gradlew -p mocks :dockerComposeUp`
    - `./gradlew :domain-ee:ee-ep-merge-app:endToEndTest`


domain-ee/ee-ep-merge-app/src/python_src/schema/merge_job.py
```diff
@@ -38,16 +37,14 @@ def __str__(self):
...
     job_id: UUID
     pending_claim_id: conint(strict=True)
     ep400_claim_id: conint(strict=True)
     state: JobState = JobState.PENDING
     error_state: JobState | None = None
     messages: list[Any] | None = None
-    created_at: datetime = _init_time
-    updated_at: datetime = _init_time
+    created_at: datetime
+    updated_at: datetime
```
nanotone: I don't know Pydantic super well, but I think we can salvage this rather than go through all that work with every instantiation. What if you import Field from pydantic and do something like

```python
    created_at: datetime = Field(default_factory=datetime.now)
    updated_at: datetime = Field(default_factory=datetime.now)
```

nanotone (changes requested):
Give that suggestion a spin and see how it goes. If it doesn't work, we can stick with the hard way.

domain-ee/ee-ep-merge-app/src/python_src/schema/merge_job.py
dfitchett: Oh nice. I'll play with it and see!

dfitchett: Turns out doing it that way makes those two time stamps different. So on job submission it will return different values for those implying that the job has been updated but it hasn't at that point.

nanotone: Oh, yeah I should've thought of that.

Is it still possible to hand-bake it into the model, perhaps by implementing an `__init__` that cleverly provides values for the two fields if they're None (and then calling the super-init)?

dfitchett: Figured out a way to set the updated_at to the created_at time on init.

domain-ee/ee-ep-merge-app/src/python_src/schema/merge_job.py
```diff
@@ -38,16 +37,20 @@ def __str__(self):
...
     error_state: JobState | None = None
     messages: list[Any] | None = None
-    created_at: datetime = _init_time
-    updated_at: datetime = _init_time
+    created_at: datetime = Field(default_factory=datetime.now)
+    updated_at: datetime = None
+
+    def __init__(self, *args, **kwargs) -> None:
+        super().__init__(*args, **kwargs)
+        if "updated_at" not in kwargs or not self.updated_at:
```
nanotone: I think you probably only need one of these? Depending on whether you want to allow calling with `updated_at=None` or to enforce non-null.

nanotone (approved):
Just the one suggestion but this looks good.

PR #2756, opened 2024-03-14
Update spring-boot-version 

agile-josiah:
## What was the problem?
tomcat ddos vulnerability

### DIFF
--- a/api-gateway/gradle.properties
+++ b/api-gateway/gradle.properties
@@ -2,4 +2,4 @@
 
 healthcheck_port=8061
 
-spring_boot_version=3.2.2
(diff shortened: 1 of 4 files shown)


nelsestu:
It is not anything that is required to address now, as I know we are in the middle of troubleshooting production rabbitmq issues and just trying to pass secrel.  But I wanted to put this comment out for further consideration. We really should only need to reference the spring_boot_version property in the root gradle.properties.  Typically we would only respecify that property in subdirectory properties files in cases that need to override it. If someone knows of another reason why we need the subdirectories to specify, please respond.

PR #2852, opened 2024-04-11
WIP - Offers extended deadline to the helm upgrade command used by our deployment scripts

nelsestu:
## What was the problem?

We see relatively frequent occurrence of 'context deadline exceeded' errors during deployments.  While the message that accompanies this problem sounds highly ambiguous, the error is not that ambiguous, and the resolution is to just try again in a moment. The github actions runner that is actively running the deployment could be experiencing resource utilization issues, for example: network or CPU utilization. Or it could be the LHDI cluster that is experiencing issues, or a burst of k8 commands etc. While there isn't much we can do to prevent the root cause from occurring, we can avoid timeout errors by increasing the timeout. [more info here](https://uptrace.dev/blog/context-deadline-exceeded.html)

- #?

## How does this fix it?[^1]
By setting the timeout parameter to 10 minutes, we'll alleviate most occurrences of the problem.

## How to test this PR
- deploying to dev will test the helm upgrade command


helm/deploy.sh
```diff
@@ -94,6 +94,7 @@ set -x
 # Exit with error code when command fails so that GH Action fails
 set -e
 helm upgrade "$RELEASE_NAME" "helm/$HELM_CHART" -n "${NAMESPACE}" \
+  --timeout 600 \
```
lisac, line 93: to double-check: I'm reading this as 600 seconds (10 minutes). The PR writeup suggests a timeout parameter of 5 minutes. Do we need to adjust this value?  potentially:
```suggestion
  --timeout 300 \
```

nelsestu: Sorry for the confusion, it is actually a default of 5 minutes so the comment should say 10 which is the intended value.  However, I also figured out that we are already setting the timeout in the HELM_ARGS variable that collects parameters prior to the `helm upgrade` command: https://github.com/department-of-veterans-affairs/abd-vro/blob/improve-deployment-reliability/helm/deploy.sh#L36, so this PR Is not going to help

Issue #2472, opened 2024-01-19
Tech Spec w/ prototype: Re-configure Secrel workflow to allow signing of individual images-Pair to complete in 1 sprint


msnwatson:
**User Story**

As a VRO engineer or customer of the VRO platform, I want to have the capability to obtain signatures for individual images independently from others that I do not immediately intend to deploy. This enables me to mitigate the risk of delays and lays the groundwork for increased deployment automation and autonomy for partner teams in the future. 

**Acceptance Criteria**

1. Create a technical specification outlining the solution for reconfiguring SecRel workflows to enable individual image signing. The design should be adaptable to support partner team deployment autonomy and automation in the future. Specific decisions supporting these initiatives must be explicitly identified, with recommendations provided.
2. Develop a transition plan within the tech spec to facilitate scaling additional changes or reverting them if necessary.
3. Develop a prototype that adequately tests the problem, providing sufficient evidence to assess the proposed solution's effectiveness. 
4. Ensure that dependencies, such as situations requiring unit deployment or potential incompatibilities between image versions and deployment environments, are considered in the tech spec. Coordinate with Mason to understand if there is any impact based off his recent versioning changes on this process.
5. Review the tech spec with the team, and create user stories for any necessary follow-up work to implement suggested changes.

meganhicks:
For consideration: In the follow up ticket to this we need to consider how we communicate to our partner teams

meganhicks:
@msnwatson How do we want to work with partner teams on this?

PR #2870, opened 2024-04-17
Secrel Decoupling

msnwatson:
After the changes in my PR, users of the SecRel workflow can choose to run existing images through SecRel, unrelated of when those images were published. For instance, a common issue that we used to face with running SecRel was that if there were no changes made from the last time SecRel ran, then SecRel would not run as there would be no new images published to be fed into the next step of the workflow. However, now images can be published in one run of the workflow, and then future runs of the workflow can perform SecRel scanning as many times as the user desires.

Essentially, the changes allow for publishing to be skipped and only SecRel performed whereas historically we only allowed for skipping secrel but always required image publishing as the images published were fed directly into the SecRel pipeline. Now, any published image can be fed into the SecRel pipeline. 

On PR merge to `develop`, image publishing and SecRel will continue to run in the same way it has historically: publishing all images and running SecRel on all of them simultaneously. 

With these changes, partner team engineers could select only their images for a SecRel run, decoupling their SecRel Aqua image scanning from the rest of the platform. This means that even if there are vulnerabilities found on VRO images, if the EE team’s images passes SecRel, deployment of EE team’s applications could continue to all environments without interruption. The same goes for any VRO images that we seek to deploy. Any image can be deployed with a passing Aqua status, even if other images in the platform do not pass Aqua. 

Snyk and SDE alerts must continue to pass or block deployment across the whole platform; however, Aqua is the most frequent offender in terms of deployment blockers and SecRel failures. 

In the short-term, VRO team will need to continue to manage SecRel on behalf of our partner teams, as there will definitely be an education component. However, these changes get us closer to partner teams being able to manage their own images’ SecRel statuses and to partner teams being able to perform their own deployments to LHDI environments that require SecRel-signed images.

The changes will be available for use as soon as the PR containing the changes is merged. The changes can be easily reverted through a quick PR reversion. This work does not seem to be impacted at all by the recent work on removing versioning, beyond simplifying the mental model of the process. 

- #2472 

## How to test this PR
- Tested through manual workflow dispatching
- Will have to followup with some testing after merge as certain triggers can only be performed once the workflow is in the `develop` branch

### DIFF
--- a/.github/workflows/secrel.yml
+++ b/.github/workflows/secrel.yml
@@ -210,11 +210,11 @@ jobs:
           fi
 
           # Throw an error if the IMAGE_LIST variable is an empty array
-          if [ ${{ IMAGE_LIST }} == "[]"]; then
+          if [ "${{ IMAGE_LIST }}" == "[]" ]; then
             exit 1
           fi
 
-          echo "image_list=IMAGE_LIST" >> "$GITHUB_OUTPUT"
+          echo "image_list=${IMAGE_LIST}" >> "$GITHUB_OUTPUT"
 
   secrel:
     name: SecRel Pipeline
--- a/.github/workflows/secrel.yml
+++ b/.github/workflows/secrel.yml
@@ -221,7 +221,7 @@ jobs:
 
           echo "${{needs.gate-check.outputs.run-secrel}}"
 
-          RUN_SECREL="${{needs.gate-check.outputs.run-secrel}}"
+          RUN_SECREL="true"
           {
             echo "image_list=${IMAGE_LIST}"
             echo "run_secrel=${RUN_SECREL}"
--- a/.github/workflows/secrel.yml
+++ b/.github/workflows/secrel.yml
@@ -6,6 +6,7 @@ on:
   push:
     branches: [ main, qa, develop, domain-* ]
 
+  # Run SecRel on Dependabot PRs, automatically
   pull_request:
     branches: [ develop ]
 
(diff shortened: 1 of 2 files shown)


nelsestu (approved):
This sounds and looks just like I hoped it would. Thanks for making this happen and for communicating the significance of this enhancement so clearly.  The only other question I had was whether we need to be handling the case where images have already been signed?  This might be adding scope to this ticket, and if so, we can answer with a new backlog item.  But what i was thinking is something like this:

```
# Perform signature verification
          if docker trust inspect --pretty --signer <SIGNER_NAME> ${{ github.repository }}:${{ github.sha }}; then
            echo "Image is signed by expected signer"
            echo "::set-output name=signature_verified::true"
          else
            echo "Image is not signed by expected signer"
            echo "::set-output name=signature_verified::false"
          fi
```

tejans24 (approved):
LGTM, nice work @msnwatson . Excited to try this out!

msnwatson:
> This sounds and looks just like I hoped it would. Thanks for making this happen and for communicating the significance of this enhancement so clearly. The only other question I had was whether we need to be handling the case where images have already been signed? This might be adding scope to this ticket, and if so, we can answer with a new backlog item. But what i was thinking is something like this:
> 
> ```
> # Perform signature verification
>           if docker trust inspect --pretty --signer <SIGNER_NAME> ${{ github.repository }}:${{ github.sha }}; then
>             echo "Image is signed by expected signer"
>             echo "::set-output name=signature_verified::true"
>           else
>             echo "Image is not signed by expected signer"
>             echo "::set-output name=signature_verified::false"
>           fi
> ```

So I actually would not want to skip signing a new image that has already been signed in the past because SecRel signatures do expire an I'm not sure I'd have access to all that information from this script. As far as I know, there are no downsides to signing images multiple times beyond just computation time in the SecRel pipeline which we haven't had a need to optimize at this point.

Issue #2886, opened 2024-04-22
Complete Alembic Setup for CC Team Use


nelsestu:
As a VRO platform developer, my goal is to empower partner teams in handling their data migrations independently. 
Originally the work for Alembic Credential setup was done in #2329, but there are parts missing and work required to roll it out to each environment.

Acceptance Criteria
1. The following criteria should hold for both local and LHDI environments:
2. A new Postgres schema is created for use by the CC team
3. A new DB user with superuser privileges only for the schema from 1 is created with credentials propagated stored in HashiCorp vault and propagated to the K8s cluster.
4. References for these credentials are communicated to the CC team, including context on the expectation that CC owns it from here. The CC Team has asked for a pairing session for this.
5. Documentation is created for this offering from a VRO perspective under partner team documentation in the wiki - https://github.com/department-of-veterans-affairs/abd-vro/wiki

Notes about work
To complete this work, an engineer will need to log into the RDS postgres instance in each of our LHDI environments and use psql to manipulate the instances to satisfy the AC. This can be done by creating a shell into one of the running pods in the cluster namespace, installing psql as a CLI tool if necessary, and issuing the proper psql commands.

**Task list- please check off as you complete them**

- [x] 1. ~Use dev-tools pod to login with psql and create a new DB user and postgres schema for CC team~ the original assignee had included this manual step in the task list they created, however this is not what we want and misses the purpose of flyway, our database migration tool.  Instead of leaving this to be a manual process performed during deployment by someone who both remembers to do it, and how to do it, we want to automate this step using a flyway migration script.  That way, when the script gets deployed, the requirements we've verified in this ticket will be applied to what ever environment was just deployed.
- [ ] 1. Create a flyway migration script that gets invoked automatically just like any other database migration process.  The script will use flyways placeholder feature to support passing the values in from vault.  This suggests that this subtask will have the following configuration elements:
  - [ ] a. pass the username, password and schema name of the domain-cc postgres user in as flyway placeholder environment variables into the db-init helm chart.  These config settings will need to be populated via Kubernetes secret.  Once we have argocd in place, these kubernetes secrets will populate automatically from vault.  In the meantime creating the secrets with the correct values will be handed as a bash script as described in the next bullet point.
  - [ ] b. pass the username, password and schema name of the domain-cc postgres user in as environment variables to the domain-cc service.  This time the intended recipient is the CC team, where they will then be able to run alembic commands using the values passed via these environment variables.
[...]

Issue #2841, opened 2024-04-09
Deploy Alembic to prod 


meganhicks:
As a VRO platform developer, my goal is to empower partner teams in handling their data migrations independently. This ensures they have the freedom to select the most convenient method tailored to their application's language, offering them flexibility and adaptability. By utilizing the Alembic tool, the CC team can efficiently manage their databases and maintain data organization with ease.

Acceptance Criteria
1. Complete E2E Testing 
2. Deploy to Prod 

Notes about work
To complete this work, an engineer will need to log into the RDS postgres instance in each of our LHDI environments and use psql to manipulate the instances to satisfy the AC. This can be done by creating a shell into one of the running pods in the cluster namespace, installing psql as a CLI tool if necessary, and issuing the proper psql commands.

**Task list- please check off as you complete them**

- [ ] 5.Test to confirm a successful connection from the CC team
- [ ] 
- [ ] 6. Set up pairing sessions with the CC team and walk them through this
- [ ]

meganhicks:
This is dependent on @yurenanava  telling me when she is done building her solution so we can test.

msnwatson:
@meganhicks @nelsestu I think this can be closed?

meganhicks:
I don't think we have done this right?

yurenanava:
We have not tested it yet indeed.

PR #2901, opened 2024-04-25
Flyway migration to Create Alembic with Scripts in Support

nelsestu:
Adding a flyway migration that creates the alembic user and schema for the CC team. Populating environment variables both locally in docker compose and in LHDI with helm updates. Introducing a create-secret.sh script that provides a generic mechanism for creating kubernetes secrets

## What was the problem?
While some of the work to enable alembic for CC was completed manually, this PR is intended to automate many of the steps that were previously being done manually.

- #2886
- https://dsva.slack.com/archives/C01Q7979Z7D/p1713990908740209?thread_ts=1711990256.657119&cid=C01Q7979Z7D

## How does this fix it?[^1]
Automates the manual steps.

## How to test this PR
- Deploy to dev, qa, and sandbox 
- Note: deployment to prod environments is being tracked in #2841


rabbitmq/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Use the Bitnami RabbitMQ image as the base
-FROM bitnami/rabbitmq:3.12
+FROM rabbitmq:3.13.1
```
lisac: not sure I'm reading this correctly, but if it's the case we're not using the Bitnami library, can we remove the comment on the previous line?

helm/_shared/named_templates/_postgres.tpl
```diff
@@ -60,3 +60,34 @@
       name: vro-db
       key: DB_CLIENTUSER_PASS
 {{- end }}
+
+{{- define "domainCc.alembic.envVars" -}}
```
lisac, line 64: something beyond my helm chart knowledge, not sure it's noteworthy: there's another partial template using this name: https://github.com/department-of-veterans-affairs/abd-vro/blob/7f3468fb67cd5e9bb7e56818f165a3852fa43928/helm/domain-cc/templates/_alembicdb.tpl#L4

~edit: the way I see the code, re-use of the name likely will not cause issues, considering where this other instance is defined. if anything, we could consider moving what's defined in `domain-cc` into `_shared` (but doesn't need to be part of this PR)~  i'm not confident about this. 

can we be sure that `domainCc.alembic.envVars` as defined in `_shared` won't negatively impact the helm chart for `domain-cc`?  https://github.com/department-of-veterans-affairs/abd-vro/blob/7f3468fb67cd5e9bb7e56818f165a3852fa43928/helm/domain-cc/templates/deployment.yaml#L33

lisac (approved):
nice work. 
two comments, although I'm not sure they should block the PR.

rabbitmq/Dockerfile
nelsestu: yes, good catch, this was a mistake and shouldn't have been included in the pr.  it was just something i was playing with after seeing the debian vulnerabilities pop back up for the nth time.

helm/_shared/named_templates/_postgres.tpl
nelsestu: you are right about this.  helm won't have a problem with it, because it is good at overlaying templates like this, but it does get confusing for the humans trying to follow what happens. What we really want is for the domainCc.alembic.envVars to contain the flyway variables for the db-init service and for domain-cc to contain the plain DOMAIN_CC variables.  The current setup accomplishes that but it isn't clear why and we could clean that up a bit.

nelsestu:
Unfortunately, after that RabbitMQ image change snuck in without me noticing, it looks like RabbitMQ isn't starting up on this PR.  Additional troubleshooting of that is going to be required in order to see the checks passing on this.

PR #2978, opened 2024-05-13
Fix exchange and queue incorrect naming

chengjie8:
## What was the problem?
One of the RabbitMQ fanout exchanges, bie-events-contention-associated, didn't bind with a queue correctly. This issue is causing contention associated Kafka messages not to be processed by Camel and stored in RDS correctly.

## How to test this PR
- Verify the local RabbitMQ setup and check if the exchange binds with a queue on the RabbitMQ management admin page.
- Verify that contention associated Kafka messages are stored properly in RDS.


docker-compose.yml
```diff
@@ -124,7 +124,6 @@ services:
       BIP_EVIDENCE_USERID: ${BIP_EVIDENCE_USERID}
       BIP_EVIDENCE_SECRET: ${BIP_EVIDENCE_SECRET}
       BIP_TRUSTSTORE: ${BIP_TRUSTSTORE}
-      BIP_PASSWORD: ${BIP_PASSWORD}
```
lisac: just to double-check: is this deletion intentional?

domain-xample/xample-workflows/src/main/java/gov/va/vro/routes/xample/DbHelper.java
```diff
@@ -81,7 +81,9 @@ public ContentionEventEntity saveContentionEvent(final BieMessagePayload bieMess
     contentionEventEntity.setActionResultName(bieMessagePayload.getActionResultName());
     contentionEventEntity.setDescription(bieMessagePayload.getDescription());
 
-    return contentionEventRepository.save(contentionEventEntity);
+    contentionEventRepository.save(contentionEventEntity);
+    log.info("Saved bieMessagePayload {}", bieMessagePayload);
```
lisac, line 85: I think this might not come up in practice due to this code being in the xample app, but just wanted to double-check: i see that [BieMessagePayload](https://github.com/department-of-veterans-affairs/abd-vro/blob/a4f75bd8a956dde965b121ff4caaa1fea4be06c5/shared/lib-bie-kafka/src/main/java/gov/va/vro/model/biekafka/BieMessagePayload.java#L51) has a `description` field - is that something that might contain PII? if so, any precautions we should take about logging the payload?

shared/lib-bie-kafka/src/main/java/gov/va/vro/model/biekafka/test/BieMessagePayloadFactory.java
```diff
@@ -45,12 +45,13 @@ private static void setDeletedPayloadValues(BieMessagePayload payload) {
 
   public static BieMessagePayload create() {
     // generate a random contention classification event type
-    ContentionEvent eventType = faker.options().option(ContentionEvent.values());
+    //    ContentionEvent eventType = faker.options().option(ContentionEvent.values());
+    ContentionEvent eventType = ContentionEvent.CONTENTION_ASSOCIATED;
```
lisac: this appears to be a slight adjustment to the test behavior, changing the event type from random selection to `CONTENTION_ASSOCIATED` - does that align with your intent? if so, would recommend deleting the comment and the commented-out `faker` code (L47-48).

lisac (changes requested):
the code looks clean to me. had a few questions, more to double-check details, but nothing major. thanks!

domain-xample/xample-workflows/src/main/java/gov/va/vro/routes/xample/DbHelper.java
nelsestu: @lisac you raise an excellent point here (two of them actually):
1. first, despite this code existing in the xample project, it is a relevant point because xample contains both example/test code and production code. We've talked about prioritizing the separation a long time ago, however I don't see anything in the backlog about it when looking just now.
2. second, I can see that some thought was applied around that concern based on the [NoLogging annotation](https://github.com/department-of-veterans-affairs/abd-vro/blob/a4f75bd8a956dde965b121ff4caaa1fea4be06c5/shared/lib-bie-kafka/src/main/java/gov/va/vro/model/biekafka/BieMessagePayload.java#L45). However, I am not seeing how that annotation would prevent the logging from working in this case since the entity is simply being formatted 

So all that to say that we need to confirm that this log.info call is properly obfuscated.

nelsestu (changes requested):
We just need to address the comment about the log.info line and we should be able to approve this after that peace of mind.

docker-compose.yml
chengjie8: hmmm interesting, this could be a mistake. I will revert this change. Thanks for catching that.

domain-xample/xample-workflows/src/main/java/gov/va/vro/routes/xample/DbHelper.java
chengjie8: @lisac @nelsestu, I introduced the `@NoLogging` annotation to be used on top of the description field. Then, I incorporated the logic to not print the value of any field that has this annotation in the toString() method.
The benefits of that approach are that it makes the code clean and self-explanatory without introducing additional if conditions.

shared/lib-bie-kafka/src/main/java/gov/va/vro/model/biekafka/test/BieMessagePayloadFactory.java
chengjie8: Oops, yes, I will revert this as well. It was intended to force the `eventType` to be of this type only, as opposed to being set in a randomized fashion.

domain-xample/xample-workflows/src/main/java/gov/va/vro/routes/xample/DbHelper.java
lisac: @chengjie8 : I hadn't noticed the `@NoLogging` annotation - yes, that looks like it addresses my question. It's a clean approach - nice!
Although... it brings me to a different request: would you be able to ensure that there's some sort of test that verifies the no logging behavior on the `description` field? I'm not certain about this, but possibly this would involve asserting that it's not present in the return value of the class's `toString()` method? Such a test might already be in-place, however I wasn't able to find it in a quick browse, including of [BieMessageUtilsTest.java](https://github.com/department-of-veterans-affairs/abd-vro/blob/develop/svc-bie-kafka/src/test/java/gov/va/vro/services/bie/utils/BieMessageUtilsTest.java). 

What I want to protect against is someone removing that annotation without realizing the consequences; and I suppose also protecting against someone accidentally applying that annotation to other fields and not being aware of how it impacts the `toString()`.

lisac (approved):
thanks for fielding that round of questions, @chengjie8 . i had [one follow-up request ](https://github.com/department-of-veterans-affairs/abd-vro/pull/2978#discussion_r1631286738)around the `@NoLogging` behavior, however i defer to you on whether that needs to be part of this PR.

Issue #2988, opened 2024-05-14
Add test coverage reports to EE projects in VRO.


dfitchett:
## Summary
Currently, EP Merge and Max Ratings API projects in VRO do not have test coverage reports. Test coverage of minimum of 80% is required for staging review.

## Associated Tickets
* #2892

PR #2989, opened 2024-05-14
Add support for test coverage reports via pytest-cov.

dfitchett:
## What was the problem?

EE projects did not include test coverage. 

- #2988 

## How does this fix it?[^1]

Adds [pytest-cov](https://pytest-cov.readthedocs.io/en/latest/readme.html) to domain-ee subprojects.

## How to test this PR
- Run via gradle: ```./gradlew :domain-ee:test```
- Run via `pytest` from individual project folder

### DIFF
--- a/domain-ee/ee-ep-merge-app/src/requirements.txt
+++ b/domain-ee/ee-ep-merge-app/src/requirements.txt
@@ -5,6 +5,7 @@ httpx==0.24.*
 psycopg2-binary==2.9.*
 pydantic==2.4.*
 pytest==7.3.1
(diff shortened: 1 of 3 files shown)


nanotone (approved):
Approving, with the understanding that this change doesn't incorporate test coverage reports into GitHub PR feedback.

dfitchett:
> Approving, with the understanding that this change doesn't incorporate test coverage reports into GitHub PR feedback.

Seemed like more work than it was worth. If there isn't 80% pytest coverage, then the code check will fail.

PR #3009, opened 2024-05-21
EE Code Quality - Add Ruff linter/formatter

dfitchett:
## Summary
Removed black and flake8 to replace with Ruff which covers the same formatting and lint requirements. This PR includes formatting changes only as a result of running the ruff format. Also modified the precommit hook to call the linter and formatter. 

## Associated tickets
- #3010


dfitchett:
BLUF: _most_ of the lines changed are results of Ruff formatting changing from double quotes to single quotes
---
Couple of things to note about these updates:
* ruff linter is a drop in replacement for pyflake8 and other linters, and defaults are set to match
* ruff formatter is a drop in replacement for black but allows setting the default quote style to single quotes (which we prefer over double quotes)
* neither ruff by itself or black by itself sort imports so isort is still being used.
* build.gradle files have steps added that automatically run isort and ruff formatting (in black style) as requirements to run pytests through gradle.
* Settings for these tools are in the pyproject.toml files.

nanotone (approved):
This diff made my eyes glaze over but 👍

PR #3049, opened 2024-05-29
Max Ratings API: Updates from eCFR for digestive systems.

dfitchett:
## What was the problem?

EE was provided an updated eCFR for digestive systems, and the maximum ratings API needs updating.

- #3047

## How does this fix it?[^1]

Adds another lookup table that adds/updates diagnostic codes with their associated maximum ratings. In future updates, only the file will need to be added to make eCFR updates. 

## How to test this PR
Run the checks (unit tests, isort, ruff, etc.,):
```./gradlew check```

### CODE REVIEW

File: domain-ee/ee-max-cfi-app/src/python_src/util/lookup_table.py [hunk 1]
```diff
@@ -1,13 +1,49 @@
...
+    for file in data_files:
+        if file.startswith('Max Ratings DC Lookup Table') and file.endswith('.csv'):
+            ratings = import_ratings(file)
+            msg = f'Loading {len(ratings)} ratings from "{file}"'
+            updates = ratings & diagnostic_code_to_max_rating.keys()
+            adds = updates ^ ratings.keys()
+            if len(adds) > 0:
+                msg += f' | Adds {len(adds)}'
+            if len(updates) > 0:
+                msg += f' | Updates {len(updates)}'
```
nanotone: It's Pythonic to just use the simplest boolean evaluation for truthiness tests. i.e. `if adds:` and `if updates:`

domain-ee/ee-max-cfi-app/src/python_src/util/data/Max Ratings DC Lookup Table v2.csv
```diff
@@ -0,0 +1,60 @@
...
+7200,"Soft tissue injury of the mouth, other than tongue or lips:",90,The Digestive System,,,38 CFR 4.114
+7201,"Lips, injuries of.",90,The Digestive System,,,38 CFR 4.114
+7202,"Tongue, loss of whole or part.",100,The Digestive System,,,38 CFR 4.114
+7203,"Esophagus, stricture of.",80,The Digestive System,,,38 CFR 4.114
+7204,Esophageal motility disorder:,80,The Digestive System,,,38 CFR 4.114
+7205,"Esophagus, diverticulum of, acquired.",80,The Digestive System,,,38 CFR 4.114
+7206,"Gastroesophageal reflux disease:	",80,The Digestive System,,,38 CFR 4.114
+7207,Barrett’s esophagus:,80,The Digestive System,,,38 CFR 4.114
+7301,"Peritoneum, adhesions of, due to surgery, trauma, disease, or infection:",80,The Digestive System,,,38 CFR 4.114
+7303,"Chronic complications of upper gastrointestinal surgery:	",80,The Digestive System,,,38 CFR 4.114
```
nanotone: This almost certainly doesn't really matter, but I noticed that the rated issue names in v2 often end with a colon, and occasionally a colon followed by some number of spaces. In v2, they end with a period or nothing at all. 🤷

nanotone (approved):
This looks good to go; I have no blocking concerns.

One question, maybe a slight worry: wouldn't it be easier, long-term, to collapse updates into the one CSV that gets loaded? Not that Python can't load up gobs and gobs of CSVs at startup-time, but it might get unwieldy for a human to figure out what file is responsible for a particular DC's maximum.

dfitchett:
> This looks good to go; I have no blocking concerns.
> 
> One question, maybe a slight worry: wouldn't it be easier, long-term, to collapse updates into the one CSV that gets loaded? Not that Python can't load up gobs and gobs of CSVs at startup-time, but it might get unwieldy for a human to figure out what file is responsible for a particular DC's maximum.

Yeah, that's a good point. My original thought was that it would be easier to add updates to the app if all you had to do was drop in the file, but that does make it somewhat harder to find where any diagnostic code's max rating comes from. Maybe I'll pull that functionality out of the app and into a script that combines the files into one.

domain-ee/ee-max-cfi-app/src/python_src/util/data/Max Ratings DC Lookup Table v2.csv
dfitchett: Yeah, i didnt do any formatting of that file. I can go through and remove unnecessary periods, colons, spaces, etc.,

domain-ee/ee-max-cfi-app/scripts/dc_lookup_table_updater.py
```diff
@@ -0,0 +1,75 @@
...
+        csv_writer.writerow(['Diagnostic Code', 'Rated Issue Name', 'Max Rating', 'Body System', 'Category', 'Subcategory', 'CFR Reference'])
+        for diagnostic_code, data_tuple in data.items():
+            csv_writer.writerow([diagnostic_code, *data_tuple])
+
+
+def increment_table_version():
+    with open('../src/python_src/util/data/table_version.py', 'r') as file:
+        version = int(file.readline().strip().split(' = ').pop())
+    with open('../src/python_src/util/data/table_version.py', 'w') as file:
+        file.write(f'TABLE_VERSION = {version + 1}\n')
```
nanotone, line 65: Perhaps the table_version.py path could be made into a module-level constant like `DATA_FILE`.

domain-ee/ee-max-cfi-app/scripts/dc_lookup_table_updater.py
```diff
@@ -0,0 +1,75 @@
...
+    with open(DATA_FILE, 'w') as file:
+        csv_writer = csv.writer(file)
+        csv_writer.writerow(['Diagnostic Code', 'Rated Issue Name', 'Max Rating', 'Body System', 'Category', 'Subcategory', 'CFR Reference'])
+        for diagnostic_code, data_tuple in data.items():
+            csv_writer.writerow([diagnostic_code, *data_tuple])
+
+
+def increment_table_version():
+    with open('../src/python_src/util/data/table_version.py', 'r') as file:
+        version = int(file.readline().strip().split(' = ').pop())
```
nanotone, line 63: The brittleness of this line makes me a bit nervous, because at a glance `table_version.py` is just any Python file, but it has to be formatted a certain way or else this function breaks.

Suggestions:
- You could convert it into a minimal text file with just the version number, which makes this parsing simpler, but instead of importing it later you'd have to read-and-parse.
- Or you could bank on the `readline()` behavior and add a human-readable comment at line 2 saying it's auto-generated by `dc_lookup_table_updater.py`.

nanotone (approved):
LGTM. I gave a suggestion about the version file but this is fine as-is, too.

I really like this new setup where Git tracks changes to the CSV. (If we want to do anything with the changes, GitHub UI brings a lot to the table, and custom tooling is possible too 🤓)

PR #3080, opened 2024-06-12
Chengjie8/update versions

chengjie8:
update rack version


msnwatson (approved):
please make sure we don't see any breaking changes by deploying this to LHDI environments since it looks like a major version change

PR #3098, opened 2024-06-18
revert back some library version numbers to test ep merge integration…

chengjie8:
Test

### DIFF
--- a/svc-bgs-api/src/Gemfile
+++ b/svc-bgs-api/src/Gemfile
@@ -7,8 +7,7 @@ gem 'cgi', '~> 0.3.6' # Address CVE-2021-41819
 gem 'bunny', '>= 2.13.0'
 gem 'activesupport', '~> 6.0'
 gem 'bgs_ext', git: 'https://github.com/department-of-veterans-affairs/bgs-ext.git', require: 'bgs'
-gem 'rexml', '>= 3.2.7'
-gem 'rack', '>= 3.0.6.1'
+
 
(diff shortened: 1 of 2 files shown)


svc-bgs-api/src/Gemfile
```diff
@@ -7,8 +7,7 @@ gem 'cgi', '~> 0.3.6' # Address CVE-2021-41819
 gem 'bunny', '>= 2.13.0'
 gem 'activesupport', '~> 6.0'
 gem 'bgs_ext', git: 'https://github.com/department-of-veterans-affairs/bgs-ext.git', require: 'bgs'
-gem 'rexml', '>= 3.2.7'
```
lisac, line 10: to double-check: is removing `rexml` here desired? looking at the PR #3080 that i believe we're attempting to partially revert, i don't see `rexml` as one of the changes.

lisac (approved):
seeing that the EP Merge CI tests pass when using this branch, i support merging it in. Great work, @chengjie8 !

svc-bgs-api/src/Gemfile
chengjie8: I was just testing to see which added library causes the ep-merge integration test to fail. It turned out that the rack library was the issue. I've added this back. Thanks for pointing it out @lisac.

chengjie8: rexml is also breaking the integration test. Now i need to remove it.

Issue #3114, opened 2024-06-21
Investigate Flakey EP Merge End2End tests


dfitchett:
## Summary 
With most recent PR of EP Merge, there is suddenly flakey end 2 end tests for EP Merge. 

## Acceptance Criteria:
Find and Fix

dfitchett:
Working with @brostk to pin-point failure source of flakey tests due to recent updates to BIP.

PR #3118, opened 2024-06-25
Investigate Flakey EP Merge End2End tests

brostk:
## What was the problem?
ep-merge application e2e tests have been failing since a recent update to metric logging within svc-bip-api.  This PR is an attempt to clean up the spring implementation and add environmental support to omit metric logging in local/test regions where the services are not configured to perform metric logging.

- #3114 

## How does this fix it?[^1]
We're now building multiple MetricLoggers or initializing the DataDogConfiguration only once, and we're now only setting up metric logging in non-local/test environments.

## How to test this PR
- Step 1 CI tests for ep-merge e2e integration should pass.

### DIFF
--- a/svc-bip-api/src/main/java/gov/va/vro/bip/config/LocalEnvironmentCondition.java
+++ b/svc-bip-api/src/main/java/gov/va/vro/bip/config/LocalEnvironmentCondition.java
@@ -1,0 +1,24 @@
+package gov.va.vro.bip.config;
+
+import org.apache.commons.lang3.StringUtils;
+import org.springframework.context.annotation.Condition;
+import org.springframework.context.annotation.ConditionContext;
+import org.springframework.core.type.AnnotatedTypeMetadata;
+
+import java.util.stream.Stream;
+
+public class LocalEnvironmentCondition implements Condition {
+
+  private final String[] LOCAL_ENVS = new String[] {"local", "test"};
+
+  @Override
+  public boolean matches(ConditionContext context, AnnotatedTypeMetadata metadata) {
+    String env = context.getEnvironment().getProperty("bip.env");
+
+    if (StringUtils.isBlank(env)) {
+      return false;
+    }
+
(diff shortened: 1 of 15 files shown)


svc-bip-api/src/main/java/gov/va/vro/bip/config/LocalEnvironmentCondition.java
```diff
@@ -0,0 +1,24 @@
...
+import org.apache.commons.lang3.StringUtils;
+import org.springframework.context.annotation.Condition;
+import org.springframework.context.annotation.ConditionContext;
+import org.springframework.core.type.AnnotatedTypeMetadata;
+
+import java.util.stream.Stream;
+
+public class LocalEnvironmentCondition implements Condition {
+
+  private final String[] LOCAL_ENVS = new String[] {"local", "test"};
```
lisac, line 12:
```suggestion
  private final String[] LOCAL_ENVS = new String[] {"local", "dev", "test"};
```
I'm not confident about this suggestion. Suggested in case this is relevant: 
https://github.com/department-of-veterans-affairs/abd-vro/blob/1078dba07e8795aa8d79d354eb40449808d15983/svc-bip-api/src/main/resources/application-dev.yaml#L6

lisac (approved):
this looks like a major improvement to the `svc-bip-api` codebase; and if it also solves the flaky end2end CI tests? amazing!

svc-bip-api/src/main/java/gov/va/vro/bip/config/LocalEnvironmentCondition.java
lisac: disregard! i thought `dev` in this case mapped to something like "local dev"; however, it's actually a match for the deployment environment named `dev` (which is in the pre-prod cluster, along with deployment environments `qa` and `sandbox`).  thus, `dev` should NOT be considered a local environment.

Issue #3175, opened 2024-07-10
EP Merge: Set TTL on requests sent to downstream VRO queues


dfitchett:
# Summary
Currently messages that are sent from EP Merge to any VRO service that is down will remain in the queue until that service comes online and consumes the messages from the queue. EP Merge will always reject the responses in those cases if they have reached the max latency set for the clients in the `hoppy_service.py` which is 30s. 

## Possible Effects:
None of the `GET` endpoints would have any adverse effects, but there can be issues if the requests are modifying downstream data. For all of the cases below, actions could be completed in any order even though the merge job might have been completed unsuccessfully: 
* cancelling - shouldn't have any issue, temp station of jurisdiction might have been reverted later
* adding or updating contentions - contentions could be duplicated on existing claims (both pending and ep400), special issues removed, temp station of jurisdiction, but in ANY order
* setting the temp_station_of_jurisdiction - could overwrite original and remove special issue in any order 
* adding claim note - claim note would be added for every request made that sat in the queue until BGS was back up (default is 3 requests)

## Temporary solution
Request that VRO apply queue level TTL to 0 seconds for all BGS/BIP queues used by EP Merge.

## Best / Long term solution
Update `lib-hoppy` to send requests to queue with per message TTL option set to client's max-latency value. 

# Acceptance Criteria
- [x] Update lib-hoppy #3180 
- [x] Update EP Merge to use updated lib-hoppy #3186
- [x] Redeploy to `sandbox`

![Image](https://github.com/department-of-veterans-affairs/abd-vro/assets/135860892/f3250ebd-7b4c-46a3-93d0-62a09bd25dff)

PR #3180, opened 2024-07-11
lib-hoppy: Update to include per message TTL parameter for AsyncHoppyClient.

dfitchett:
## What was the problem?

During a recent BGS outage, 3 requests for two failed jobs (6 total) were sent it with no response. Currently, there is no queue wide message TTL policy, and no per message TTL was being set by the publisher (AsyncHoppyClient) when sending the requests, and although the service was down, the messages sent to add notes to a claim were still in queue. When the service came back up, it resulted in duplicate notes being added to the queue.

- #3175 

## How does this fix it?[^1]

Adds a configurable per-message TTL to the message upon [publishing](https://www.rabbitmq.com/docs/ttl#per-message-ttl-in-publishers). The default is set to zero. From the documentation:

>  Setting the TTL to 0 causes messages to be expired upon reaching a queue unless they can be delivered to a consumer immediately. Thus this provides an alternative to the immediate publishing flag, which the RabbitMQ server does not support. Unlike that flag, no basic.returns are issued, and if a dead letter exchange is set then messages will be dead-lettered.

There is currently no dead lettering happening, and is unsupported by VRO and 

## How to test this PR
- navigate to hoppy folder
  - `cd abd-vro/shared/lib-hoppy`
- install new hoppy lib (in virtual env if you have it set up)
  - `pip install .`
- run `pytest`  


nanotone (approved):
👍

PR #3211, opened 2024-07-17
update grep for finding mock containers in CI check for healthy containers

lisac:
## What was the problem?
The `container-healthchecks` job in the suite of CI checks regularly fails due to a mock container not registering a healthy status. 
We have code in-place mean to remove mock containers from consideration for this test (because they don't have health checks); however, the container for `mock-bie-kafka` is named in a way that isn't caught by this code.
As a result, the job displays a Failure result, when we can see that it's only the mock container that was unhealthy. 

Sample run showing this problem: https://github.com/department-of-veterans-affairs/abd-vro/actions/runs/9975597466/job/27565759431?pr=3209
 

## Solution proposed in this PR
1. Loosen from a match of `-mock-` to `mock-`. why: at least one mock container is named with `_mock-`, which doesn't satisfy the original pattern. 
2. Adjust logging level for failed Datadog API metrics submissions from `error` to `warning`. why: 1) I would consider these `warning` level rather than `error`; 2) for compatibility with the existing script used in the CI check for whether a container's logs appear healthy. In the build environment, the DD metrics submission will fail; and if these log messages appear at the `error` threshold, that will be enough for the script logic to interpret the container as unhealthy (which, when restricted to the DD metrics failures, is not accurate).

The better solution would be to track down why the container is being named with "_mock-" rather than "-mock-", and seeing if we can standardize to "-mock-"; however, I wasn't able to track this down.

### DIFF
--- a/.github/workflows/container-healthchecks.yml
+++ b/.github/workflows/container-healthchecks.yml
@@ -82,7 +82,7 @@ jobs:
           # Ignore mocks since they don't have health checks
           # TODO: add health check for svc-bgs-api; ignore it for now
           if docker container ls --format '{{.Names}} \t {{.Status}}' | grep -v "(healthy)" \
-            | grep -v "svc-bgs-api\|-mock-"; then
+            | grep -v "svc-bgs-api\|mock-"; then
             echo 'There are unexpected unhealthy containers!'
             exit 2
           fi
--- a/svc-bgs-api/src/lib/metric_logger.rb
+++ b/svc-bgs-api/src/lib/metric_logger.rb
@@ -32,7 +32,7 @@ class MetricLogger
       api_instance.validate
       $logger.info('Succeeded Datadog authentication check')
     rescue Exception => e
-      $logger.error("Failed Datadog authentication check: #{e.message}")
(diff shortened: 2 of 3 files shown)


lisac:
update: not ready for review - i found something else to address

.github/workflows/container-healthchecks.yml
```diff
@@ -82,7 +82,7 @@ jobs:
           # Ignore mocks since they don't have health checks
           # TODO: add health check for svc-bgs-api; ignore it for now
           if docker container ls --format '{{.Names}} \t {{.Status}}' | grep -v "(healthy)" \
-            | grep -v "svc-bgs-api\|-mock-"; then
+            | grep -v "svc-bgs-api\|mock-"; then
```
nelsestu, line 85: great catch in identifying why this filtered list of containers is failing to exclude all mock containers.

nelsestu (approved):
I can also see that the softening of logging levels would allow for the application to continue operating despite the degradation of metrics.  I like this because it provides greater opportunity to debug the situation. As a warning we'll see multiple warnings logged when something is misconfigured, as opposed to a single error which be easily missed amongst the other noise.

Issue #3224, opened 2024-07-19
SecRel scheduled scan does not scan images


brostk:
The SecRel action is currently designed to only publish images once per commit hash.  Because of this, and because SecRel scans are automatically triggered with each PR merge, any scheduled scans will only ever run against code which has already been scanned and signed.  Because of the how the action is coded, when no images are published there are then no images passed along in the workflow to scan.

The action should be modified to perform a full SecRel scan while still only publishing images once per commit hash.

**Acceptance Criteria**

1. Scheduled runs continue to not publish duplicate images
2. Scheduled runs perform full scans, including Aqua and Snyk checks
3. Manual workflows continue to work for single services as well as 'all' services

PR #3226, opened 2024-07-19
3224 build image tag array when image already exists for schedule runs

brostk:
## What was the problem?
Scheduled SecRel runs did are not performing scans because images have already been built during the previous PR merge.

- #3224 

## How does this fix it?[^1]
In the action script, when we find an image which has already been updated, for scheduled runs we will still add the image name and tag to the output list of images used further down the SecRel pipeline.

## How to test this PR
- Step 1 wait for automated secrel scan to run and observe the results.  Currently, the Aqua and Snyk checks are skipped because there are no image tags to scan - this should no longer be the case with this PR.

### DIFF
--- a/.github/actions/publish-images/action.yml
+++ b/.github/actions/publish-images/action.yml
@@ -64,9 +64,20 @@ runs:
             IMG_NAME="$(getVarValue "${PREFIX}" _IMG)"
             echo "::group::Push image $IMG_NAME $IMG_TAG"
             GHCR_PATH="ghcr.io/${{ github.repository }}/${IMG_NAME}"
+
             if [ "$(imageTagExists "$IMG_NAME" "$IMG_TAG")" == "200" ]; then
               echo "Image already exists: $IMG_NAME:$IMG_TAG -- not overwriting"
               echo "* ($IMG_NAME:$IMG_TAG -- already exists, not overwriting)" >> "$GITHUB_STEP_SUMMARY"
+
+              if [ "${{ github.event_name }}" == "schedule" ]; then
+                echo "* $IMG_NAME:$IMG_TAG" >> "$GITHUB_STEP_SUMMARY"
+                echo -n "\"${GHCR_PATH}:$IMG_TAG\"" >> "$GITHUB_OUTPUT"
+
+                # For all but the last image include a comma
+                if [ ${PREFIX} != ${VAR_PREFIXES_ARR[-1]} ]; then
+                  echo -n "," >> "$GITHUB_OUTPUT"
+                fi
+              fi
             else
               GRADLE_IMG_NAME=$(getVarValue "${PREFIX}" _GRADLE_IMG)
 


lisac (approved):
This looks good to me. Cross-referenced the [github](https://docs.github.com/en/actions/learn-github-actions/contexts#example-usage-of-the-github-context) [docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule) re: `github.event_name` and `schedule`.

----
Out of scope for this PR, but wanted to ask: in theory, is there potential to do a check of what's in GHCR before running the "Build the images" step?  and if the images that would be generated by this build already exist in GHCR, then to skip the steps of "Build the images" and "Test code" (which would have had to have succeeded in the previous run that pushed the image to GHCR); but still populate the list of images into `$GITHUB_OUTPUT`, to be available for follow-on actions; or would this end up dropping a key step? 
potential benefit of reduced run time. on the other hand: considering this action is currently scheduled to run before the workday, we might not feel the effects of the reduced run time.
just asking in theory, not looking to put this on the backlog.

.github/actions/publish-images/action.yml
```diff
@@ -64,9 +64,20 @@ runs:
             IMG_NAME="$(getVarValue "${PREFIX}" _IMG)"
             echo "::group::Push image $IMG_NAME $IMG_TAG"
             GHCR_PATH="ghcr.io/${{ github.repository }}/${IMG_NAME}"
+
             if [ "$(imageTagExists "$IMG_NAME" "$IMG_TAG")" == "200" ]; then
               echo "Image already exists: $IMG_NAME:$IMG_TAG -- not overwriting"
               echo "* ($IMG_NAME:$IMG_TAG -- already exists, not overwriting)" >> "$GITHUB_STEP_SUMMARY"
```
nelsestu, line 70: Hi @brostk , I was just checking in on the status of this "scheduled SecRel" fix, and noticed that your PR needed another review.  So taking a look at it, it looks good, but the one thing I noticed was that we are still writing this "already exists, not overwriting" message to  stdout, even though that won't be true for the scheduled event trigger. If this sounds correct, I wanted to recommend that we skip these echo statements to avoid misleading messages in the build log. I would imagine that adding an else condition to the if on line 72 would be an easy way to remove those "not overwritting" messages.

brostk: @nelsestu Thanks for the second set of eyes - I thought it best to keep the echo since `docker tag ...; docker push ...` isn't being run and so we aren't actually overwriting the image in this case, we're just running the scan on it.  Does that sound right to you?  I'm happy to move the echo around so it's more representative if you think leaving it as-is might be misleading.

nelsestu: Ah! that makes sense and would be correct but the non-obvious implication of running a scan is that if that scan succeeds then it will result in the republication of images.  It is true that we won't change the tags and it is true that this action won't be pushing any images, however once SecRel passes it will sign the images and republish those.  This actually touches on one point of feedback we gave to the SecRel team last week, which was that it would be great if there was some way to query ghcr.io inquiring whether an image has been SecRel signed. At the present moment, and given how we tag images today, we don't know whether an image has been SecRel signed or not.  Going back to the GH Action logs is one way of knowing.  Pulling the image and then checking the binary for signatures would be another.

PR #3233, opened 2024-07-24
3224 move "not overwriting" echo for accuracy

brostk:
## What was the problem?
The echo for "image already exists, not overwriting" still prints even for a scheduled run when we will be overwriting the signed image.

- #3224

## How does this fix it?[^1]
The echo was moved to only print when the image already exists but we are not performing a scheduled run.

## How to test this PR
- Step 1 observe a scheduled SecRel run and verify the "$IMG_NAME:$IMG_TAG -- already exists, not overwriting" message is not printed.


nelsestu (approved):
Looks great.  Thanks for working through that with me

PR #3238, opened 2024-07-25
Add Dead Letter Queue and update existing queue configurations to use it

PaulKBaumann:
## What was the problem?
Messages that fail to process are lost.

- https://app.zenhub.com/workspaces/vro-team-6557e67173391c000e1409f3/issues/gh/department-of-veterans-affairs/abd-vro/2758

## How does this fix it?[^1]
When bip-api starts up and creates/configures queues, it will now also create a dead-letter queue. Messages that fail to process will be placed on this queue

## How to test this PR
- Build and launch the bip-api service in Docker
- Go to the RabbitMQ console and verify that the dead-letter queue was created
- Place some junk data on any of the queues. bip-api will fail to process it and you can see the message end up in the dead-letter queue.
- after implementing it, we discovered that coordinating changes for EPMerge and BIP are needed
- When we deploy this it will need to be planned to ensure we do not take down the EPMerge and BIP


shared/lib-camel-connector/src/main/java/gov/va/vro/camel/MessageQueueConfiguration.java
```diff
@@ -28,6 +32,12 @@ public class MessageQueueConfiguration {
   private final String BIE_EVENTS_CONTENTION_COMPLETED = "bie-events-contention-completed";
   private final String BIE_EVENTS_CONTENTION_DELETED = "bie-events-contention-deleted";
   private final String SAVE_TO_DB_PREFIX = "saveToDB-";
+  private final String QUEUE_MESSAGES_DLQ = "bie-events-dql";
```
CorySohrakoffUSDS: typo? `dql` -> `dlq`

shared/lib-camel-connector/src/main/java/gov/va/vro/camel/MessageQueueConfiguration.java
```diff
@@ -28,6 +32,12 @@ public class MessageQueueConfiguration {
...
+  private final String DLX_EXCHANGE_MESSAGES = "bie-events.dlx";
+  Map<String, Object> DQL_ARGS = Map.ofEntries(
```
CorySohrakoffUSDS: Same as comment above.

shared/lib-camel-connector/src/main/java/gov/va/vro/camel/MessageQueueConfiguration.java
```diff
@@ -28,6 +32,12 @@ public class MessageQueueConfiguration {
   private final String BIE_EVENTS_CONTENTION_COMPLETED = "bie-events-contention-completed";
   private final String BIE_EVENTS_CONTENTION_DELETED = "bie-events-contention-deleted";
   private final String SAVE_TO_DB_PREFIX = "saveToDB-";
+  private final String QUEUE_MESSAGES_DLQ = "bie-events-dql";
+  private final String DLX_EXCHANGE_MESSAGES = "bie-events.dlx";
```
chengjie8: Is there any reason we use the `.` instead of `-` for the exchange name here?

PaulKBaumann: I was using the same standard I saw in some RabbitMQ example code. I don't think it has any functional purpose

shared/lib-camel-connector/src/main/java/gov/va/vro/camel/MessageQueueConfiguration.java
PaulKBaumann: Fixed

shared/lib-camel-connector/src/main/java/gov/va/vro/camel/MessageQueueConfiguration.java
PaulKBaumann: Fixed

lisac:
I don't why the [CI checks](https://github.com/department-of-veterans-affairs/abd-vro/blob/develop/.github/workflows/continuous-integration.yml) haven't completed. Triggered a run manually, running in https://github.com/department-of-veterans-affairs/abd-vro/actions/runs/10115124790 .

lisac (commented):
hi @PaulKBaumann , would you please review the lint errors flagged in https://github.com/department-of-veterans-affairs/abd-vro/actions/runs/10115124790/job/27975213040 ?   I'd dismiss the `container-healthchecks` failure - looking at the log output, that looks consistent with an issue due to a buggy `grep` in the script, since corrected in the default branch.

lisac:
the lint check is now passing - ran it manually here: https://github.com/department-of-veterans-affairs/abd-vro/actions/runs/10115460344 


@PaulKBaumann : as a next step, do you mind updating your branch with what's on the default branch? I'm curious about the conflict that GH is flagging around `MessageQueueConfiguration.java`. Thanks!


---

tangent: I don't know why the CI checks appear to be stuck. My hunch: our tooling is getting confused by the `#` in the branch name, or we need to refine the filters in https://github.com/department-of-veterans-affairs/abd-vro/blob/ec7090930cf5d15b42f194717d64dc4c9c2f8d5b/.github/workflows/continuous-integration.yml#L7. The latter is on my mind after reading [this](https://github.com/actions/runner/issues/1183#issuecomment-887609732).  I think the fix is out of scope for this PR.

nelsestu:
Preparing review...

nelsestu:
The main thing I want to encourage in this PR is that we modify the message queue configuration loading to support missing configuration settings.  One problem that we frequently encounter with our services is that we fail to make them resilient to misconfiguration or missing configuration. Have a look at [RabbitMQ Config](https://github.com/department-of-veterans-affairs/abd-vro/pull/3238/files#diff-ae98042e8b36f722d19319d609d8f6bdaa30b3f79b4486158f02e06212b09203R163)- startup configuration will be difficult to diagnose without more graceful handling. For example: Add error handling or logging for the creation of the dead letter queue and its bindings.  Consider adding validation or setting default values where config wouldn't actually be needed.

nelsestu (changes requested):
Ensure that our services that depend on the MessageQueue config objects don't cause service startup issues by applying reasonable defaults. That is really the gist of my recommendation in this [comment](https://github.com/department-of-veterans-affairs/abd-vro/pull/3238#issuecomment-2253307869). I don't consider this recommendation a requirement for approval, but I am going to mark this "Request changes" to ensure the other feedback is actioned.  Thanks!

PaulKBaumann:
> @PaulKBaumann : as a next step, do you mind updating your branch with what's on the default branch? I'm curious about the conflict that GH is flagging around `MessageQueueConfiguration.java`. Thanks!

It's merged now. It was the Camel library removal changes that were conflicting.

svc-bip-api/src/main/resources/application.yaml
```diff
@@ -40,6 +40,8 @@ createClaimContentionsQueue: createClaimContentionsQueue
 updateClaimContentionsQueue: updateClaimContentionsQueue
 putTempStationOfJurisdictionQueue: putTempStationOfJurisdictionQueue
 exchangeName: bipApiExchange
+deadLetterQueueName: bipDeadLetterQueue
```
lisac:
```suggestion
deadLetterQueue: bipDeadLetterQueue
```
for consistency with the variables for queue names in the lines above

PaulKBaumann:
I added some default values for the RabbitMQ properties and logged the relevant values when creating the dead letter queue and binding.

nelsestu (approved):
Looks great. thanks Paul!

domain-ee/ee-ep-merge-app/src/python_src/service/hoppy_service.py
```diff
@@ -3,12 +3,14 @@
...
 from hoppy.async_hoppy_client import AsyncHoppyClient, RetryableAsyncHoppyClient
 from hoppy.hoppy_properties import ExchangeProperties, QueueProperties
+from typing import Dict
 
 
 class HoppyService:
     clients: dict[ClientName, AsyncHoppyClient] = {}
 
     def __init__(self) -> None:
+        self.create_client(ClientName.DEAD_LETTER)
```
dfitchett, line 13: This application does not need a client to the DLQ. This can be removed along with the logic in create client for this client.

dfitchett: This application does not need a client to the DLQ. This can be removed along with the logic in `create_client` and the queue and response-queue names in `config.py`.

domain-ee/ee-ep-merge-app/integration/conftest.py
```diff
@@ -18,6 +18,11 @@ def pytest_collection_modifyitems(items):
     items.sort(key=item_order.get)
 
 
+@pytest.fixture(autouse=True, scope='session')
```
dfitchett: This fixture isn't needed. This app won't have access to the DLQ.

domain-ee/ee-ep-merge-app/integration/mq_endpoint.py
```diff
@@ -4,26 +4,30 @@
...
 
 
 class MqEndpoint:
-    def __init__(self, name, exchange, req_queue, response_queue):
+    def __init__(self, name, exchange, req_queue, response_queue, arguments = {}):
         self.name = name
         self.index = 0
         self.auto_response_files = []
+        type = 'direct'
+        if (name == ClientName.DEAD_LETTER):
```
dfitchett: Won't need this logic since this app wont need to access the DLQ.

domain-ee/ee-ep-merge-app/src/python_src/config.py
```diff
@@ -19,10 +19,12 @@ class ClientName(str, Enum):
     UPDATE_CLAIM_CONTENTIONS = 'updateClaimContentionsClient'
     CANCEL_CLAIM = 'cancelClaimClient'
     BGS_ADD_CLAIM_NOTE = 'addClaimNoteClient'
+    DEAD_LETTER = 'bipDeadLetterQueue'
```
dfitchett: Can we prefix the client name with BIP? `BIP_DEAD_LETTER`

shared/lib-rabbitmq-connector/src/main/java/gov/va/vro/camel/MessageQueueConfiguration.java
```diff
@@ -30,6 +33,12 @@ public class MessageQueueConfiguration {
   private final String BIE_EVENTS_CONTENTION_COMPLETED = "bie-events-contention-completed";
   private final String BIE_EVENTS_CONTENTION_DELETED = "bie-events-contention-deleted";
   private final String SAVE_TO_DB_PREFIX = "saveToDB-";
+  private final String QUEUE_MESSAGES_DLQ = "bipDeadLetterQueue";
+  private final String DLX_EXCHANGE_MESSAGES = "bip-events.dlx";
+  Map<String, Object> DLQ_ARGS =
```
dfitchett: `private final` and you could use `Map.of(...)` since its a short list of entries.

svc-bip-api/src/main/java/gov/va/vro/bip/config/CancelClaimConfig.java
dfitchett, line 1: Just an observation, but I'm thinking that since you're already modifying all of these config classes, you could change them to favor [constructor injection over field injection](https://www.linkedin.com/pulse/you-should-stop-using-spring-autowired-felix-coutinho). Don't look at git blame too much as I'm guilty of some of them. 👻

svc-bip-api/src/main/java/gov/va/vro/bip/config/RabbitMqConfig.java
```diff
@@ -73,4 +73,24 @@ public void configureRabbitListeners(RabbitListenerEndpointRegistrar registrar)
   DirectExchange bipApiExchange() {
     return new DirectExchange(exchangeName, true, true);
   }
+
+  @Bean
+  FanoutExchange deadLetterExchange() {
+    log.info("Creating dead letter exchange with name={}", props.getDeadLetterExchangeName());
```
dfitchett, line 81: The log statements in these beans duplicate log statements already output by the `RabbitAdmin` on application startup. The look like this:
```
2024-08-07T15:33:22.540-07:00  INFO 67182 --- [           main] o.s.amqp.rabbit.core.RabbitAdmin         : Auto-declaring a non-durable or auto-delete Exchange (bipApi.dlx) durable:true, auto-delete:true. It will be deleted by the broker if it shuts down, and can be redeclared by closing and reopening the connection.
```

dfitchett (changes requested):

domain-ee/ee-ep-merge-app/src/python_src/service/hoppy_service.py
PaulKBaumann: It's needed because it creates the dead letter queue for the queues that use it. If the dead letter queue doesn't exist when another queue tries to use it, the configuration is just dropped and the pipeline fails later down the line when bip-api tries to connect.

domain-ee/ee-ep-merge-app/integration/conftest.py
PaulKBaumann: Same as above

domain-ee/ee-ep-merge-app/integration/mq_endpoint.py
PaulKBaumann: Same as above

svc-bip-api/src/main/java/gov/va/vro/bip/config/RabbitMqConfig.java
PaulKBaumann: My team wanted these added as a failsafe in case the configuration is botched and RabbitMQ isn't logging those details

domain-ee/ee-ep-merge-app/src/python_src/service/hoppy_service.py
dfitchett: The problem with creating it like a client is that it's also creating a publisher and consumer for this DLQ. So, for example, if a message is sent to the getClaimDetails queue, and the bip service cannot consume that message, it will time out due to the per-message TTL set when publishing. When that happens, it will be re-routed to the DLQ, but then the client in EP Merge will try consuming it three times.

domain-ee/ee-ep-merge-app/integration/mq_endpoint.py
```diff
@@ -4,26 +4,30 @@
...
     pass
 
 
 class MqEndpoint:
-    def __init__(self, name, exchange, req_queue, response_queue):
+    def __init__(self, name, exchange, req_queue, response_queue, arguments = {}):
         self.name = name
         self.index = 0
         self.auto_response_files = []
+        type = 'direct'
```
dfitchett: This can be removed now just like `hoppy_service`.

domain-ee/ee-ep-merge-app/integration/mq_endpoint.py
```diff
@@ -4,26 +4,30 @@
...
-        queue_props = QueueProperties(name=req_queue, auto_delete=True, passive_declare=False)
+        exchange_props = ExchangeProperties(name=exchange, auto_delete=True, passive_declare=False, type=type)
+        queue_props = QueueProperties(name=req_queue, auto_delete=True, passive_declare=False, arguments=arguments)
         self.consumer = async_consumer.AsyncConsumer(
             exchange_properties=exchange_props, queue_properties=queue_props, routing_key=req_queue, reply_callback=self._on_message
         )
-
-        reply_props = QueueProperties(name=response_queue, passive_declare=False, auto_delete=True)
-        self.publisher = async_publisher.AsyncPublisher(exchange_properties=exchange_props, queue_properties=reply_props, routing_key=response_queue)
+        if name != ClientName.BIP_DEAD_LETTER:
```
dfitchett: Unnecessary since we'll never make a client for `BIP_DEAD_LETTER`

domain-ee/ee-ep-merge-app/src/python_src/config.py
```diff
@@ -42,6 +45,7 @@ class ClientName(str, Enum):
     ClientName.UPDATE_CLAIM_CONTENTIONS: os.environ.get('UPDATE_CLAIM_CONTENTIONS_REQUEST') or 'updateClaimContentionsQueue',
     ClientName.CANCEL_CLAIM: os.environ.get('CANCEL_CLAIM_REQUEST') or 'cancelClaimQueue',
     ClientName.BGS_ADD_CLAIM_NOTE: os.environ.get('ADD_CLAIM_NOTE_REQUEST') or 'add-note',
+    ClientName.BIP_DEAD_LETTER: os.environ.get('BIP_DEAD_LETTER') or 'bipDeadLetterQueue',
```
dfitchett: Not needed! Same with the `REPLY_QUEUES` entry.

shared/lib-rabbitmq-connector/src/main/java/gov/va/vro/camel/MessageQueueConfiguration.java
```diff
@@ -119,6 +124,11 @@ FanoutExchange fanoutExchangeBieEventsContentionDeleted() {
         Collections.singletonMap("queues", queueSaveToDbBieDeleted().getName()));
   }
 
+  @Bean
+  FanoutExchange deadLetterExchange() {
+    return new FanoutExchange(DLX_EXCHANGE_MESSAGES, true, true);
```
dfitchett: The third argument here establishes the exchange as `auto-delete`. If we don't want the exchange to disappear when the last consumer drops the connection, then this should be set to `false`.

shared/lib-rabbitmq-connector/src/main/java/gov/va/vro/camel/MessageQueueConfiguration.java
```diff
@@ -30,6 +31,10 @@ public class MessageQueueConfiguration {
   private final String BIE_EVENTS_CONTENTION_COMPLETED = "bie-events-contention-completed";
   private final String BIE_EVENTS_CONTENTION_DELETED = "bie-events-contention-deleted";
   private final String SAVE_TO_DB_PREFIX = "saveToDB-";
+  private final String QUEUE_MESSAGES_DLQ = "bipDeadLetterQueue";
```
dfitchett: Should BIE contention events be routed to a different dead letter exchange / queue?

svc-bip-api/src/main/java/gov/va/vro/bip/config/CreateClaimContentionsConfig.java
```diff
@@ -4,22 +4,30 @@
...
-import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 
 @Configuration
 public class CreateClaimContentionsConfig {
 
-  @Value("${createClaimContentionsQueue}")
   String createClaimContentionsQueue;
```
dfitchett: `final`

dfitchett (changes requested):

shared/lib-rabbitmq-connector/src/main/java/gov/va/vro/camel/MessageQueueConfiguration.java
PaulKBaumann: It should be fine if they share it.

dfitchett: Technically, it would be! I'm just considering future use cases of dealing with the dead lettered messages in a more compartmentalized way such as different handling of messages originating from different services.

svc-bip-api/src/main/java/gov/va/vro/bip/config/RabbitMqConfig.java
```diff
@@ -73,4 +73,24 @@ public void configureRabbitListeners(RabbitListenerEndpointRegistrar registrar)
...
+    return new FanoutExchange(props.getDeadLetterExchangeName(), true, true);
```
dfitchett: Oh this will need to be changed to have `auto-delete=false` as well!

shared/lib-rabbitmq-connector/src/main/java/gov/va/vro/camel/MessageQueueConfiguration.java
PaulKBaumann: I changed it to 'vroDeadLetterQueue' so it's less confusing to see bie events in it. I think keeping it to one queue and one exchange is best for now because it keeps the config a lot simpler - we'd have to add a config option for each queue's DLX and DLQ. We'd also have more overhead because we're manually creating alarms for each one. I think we should put the one queue into production to start and see how it goes, and if we find that there's a queue that needs special attention we can go from there.

dfitchett (approved):

Issue #3261, opened 2024-08-01
monitoring corrections


lisac:
## Problem
By this point, through issues in previous sprints, all VRO apps should be emitting latency metrics to Datadog.
However, we are not seeing this data for BIP, xample-workflows, and potentially others. We ARE seeing it for BGS.

This is blocking #2817 (in-progress for this sprint).

## What needs to be done
Set up graphs in Datadog that show these metrics, per VRO app.
If the metrics aren't present, apply code changes as needed.

## Related
previous monitoring issues: #2979, #2980, #2981

lisac:
latency metrics for `svc-bip-api` should now be visible in Datadog under metric tag `vro_bip.request_duration`.

snapshot from dashboard for [VRO data services: BIP](https://lighthousedi.ddog-gov.com/dashboard/imt-ubp-z29/vro-data):

<img src='https://github.com/user-attachments/assets/a38e01ce-4478-409f-bfac-fff3f07a775b' width='500'/>

lisac:
current status: need to complete refactoring the datadog code (#3267); budget time to address issues surfaced by unit and integration tests; and also budget time for failures surfaced by deploying to pre-prod. A stretch goal would be Wednesday 8/7.

lisac:
status: anticipated last round of changes have passed code review. deployment expected 8/12.

PR #3265, opened 2024-08-01
add datadog certificate for java Dockerfile

lisac:
## What was the problem?
Calls using the Datadog API client for submitting metrics from `svc-bip-api` are failing due to a certificate issue. Observed in the app logs:
> ```exception submitting vro_bip.request_duration: javax.net.ssl.SSLHandshakeException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target```

This is a partial fix for #3261.

## How does this fix it?
- adds instructions in the dockerfile used by java apps, so that the datadog certificate is imported into the cacerts truststore. 
- adds an auth check to when datadog api client is created (as part of app startup). in lower environments, we expect this to fail due to credentials not being in-place. however, it should not fail due to the certificate issue. This auth check is a convenient way to get an early read on whether there will be a certificate issue.

## how I tested this
I deployed 47a842f to dev and verified:
1. that the cert exists on a deployed pod. Screenshot 2024-08-02 at 9 09 05 AM
2. that the app logs show the api key validation failed with a 403, which indicates that the api client was able to reach datadog (rather than running into an issue with the certification path). the 403 is expected on lower environments due to datadog credentials only being maintained on prod environment.


lisac:
in a roundabout way, potential indication that the certificate issue has been addressed:

deploying [38cf595](https://github.com/department-of-veterans-affairs/abd-vro/pull/3265/commits/38cf595da999e0b9c219ba66629c379b65d77d92) to the dev environment shows this error thrown when retrieve an AuthenticationValidationResponse through the Datadog API client. That it shows a 403 is expected, as we don't have valid credentials in the dev environment; and that we reached this error suggests to me that the api client was able to connect to datadog (that it avoided a SSLHandshakeException due to certificates).

> [gov/va/vro/bip/config/DatadogClientConfig.class]: Failed to instantiate [com.datadog.api.client.ApiClient]: Factory method 'apiClient' threw exception with message: {"status":"error","code":403,"errors":["Forbidden"],"statuspage":"https://status.ddog-gov.com","twitter":"http://twitter.com/datadogops","email":"support@datadoghq.com"}

gradle-plugins/src/main/resources/docker/Dockerfile-java
```diff
@@ -19,6 +19,13 @@ ARG JAR_FILE
 ENV JAR_FILENAME=${JAR_FILE}
 COPY ${JAR_FILE} fat.jar
 
+# add public datadog certificate to truststore
+RUN openssl s_client -showcerts -connect www.ddog-gov.com:443 2>&1 |\
+    sed -ne '/-BEGIN CERTIFICATE-/,/-END CERTIFICATE-/p' > temp.pem
+RUN openssl x509 -in temp.pem -out real.pem
+RUN keytool -importcert -alias ddog -file real.pem -cacerts -storepass changeit --noprompt
```
msnwatson, line 26: Maybe for follow up work but just washed to understand if this cert is something that should be managed by k8s cert manager that is described in LHDI docs

msnwatson (approved):

PR #3267, opened 2024-08-02
Refactor datadog metrics and apply to other apps

lisac:
## What was the problem?
We need visibility into latency metrics on the xample-workflows and BIE.

## summary of changes
1. refactor datadog api setup from `svc-bip-api` into shared lib `shared/lib-metrics`
   - change implementation of `LocalEnvironmentCondition`: assess the environment level through environment variable (`System.getenv("ENV")`)  rather than looking up a value passed through spring config (eg `bip.env`). 
     - why: for convenience in having this shared code be used across multiple apps; system environment variable will already be in-place via helm [_helpers.tpl](https://github.com/department-of-veterans-affairs/abd-vro/blob/82c80dd365b5dda759a945a566bb00988745424a/helm/_shared/named_templates/_helpers.tpl#L6) `vro.commonEnvVars` . while the java apps generally have a `{app}.env` value, I couldn't think of a generalized way to have the java code look up `{app}.env`
     - as part of this, change behavior of blank value to be assessed as a local environment
   - increase unit test coverage of the new shared lib to satisfy minimum jacoco minimum coverage 
2. update `xample-workflows` and `svc-bie-kafka` to emit metrics using the shared lib.

This is a partial fix for https://github.com/department-of-veterans-affairs/abd-vro/issues/3261.


domain-xample/xample-workflows/src/main/java/gov/va/vro/routes/xample/BieXampleRoutes.java
```diff
@@ -42,8 +60,13 @@ private void handleException(BieMessagePayload payload, Exception e) {
...
       String jsonBody = objectMapper.writeValueAsString(payload);
       log.error("FailedMessageEventBody: " + jsonBody);
-    } catch (Exception jsonException) {
+
+      metricLogger.submitCount(
+          IMetricLoggerService.METRIC.RESPONSE_ERROR, metricTagsSaveContentionEvent);
+    } catch (JsonProcessingException jsonException) {
       log.error("Error converting failed message to JSON", jsonException);
+    } catch (Exception exception) {
+      log.error(exception.toString());
```
brostk: What do we do elsewhere for loggin exceptions?  The log above (line 67) is more what I'm used to: 'log.error("some description", e)' to print the stack trace for 'e'.

shared/lib-metrics/src/main/java/gov/va/vro/metricslogging/MetricLoggerService.java
```diff
@@ -19,14 +17,15 @@
...
-  @Value("${bip.env}")
-  public String env;
-
   private static final String APP_PREFIX = "vro_bip";
   private static final String SERVICE_TAG = "service:vro-svc-bip-api";
 
   private final MetricsApi metricsApi;
 
+  public MetricLoggerService() {
+    this.metricsApi = new MetricsApi();
```
brostk: New MetricsApi here initializes it without the ApiClient parameter, is that okay?

nelsestu:
Preparing review...

nelsestu:
hold for formatted thoughts...

domain-xample/xample-workflows/src/main/java/gov/va/vro/routes/xample/BieXampleRoutes.java
```diff
@@ -14,6 +17,11 @@ public class BieXampleRoutes {
   private final DbHelper dbHelper;
   private final ObjectMapper objectMapper;
 
+  final IMetricLoggerService metricLogger = new MetricLoggerService();
```
nelsestu: Don't do it if you aren't interested, but I would encourage using spring's dependency injection magic for `MetricLoggerService` instead of direct instantiation. Everything surrounding this BieXampleRoutes class is already wired for it. For example dbHelper and objectMapper are both getting injected, so that is a good clue that this class is already doing that.  I believe all that is needed in this case is to remove the "= new MetricLoggerService()" and then just ensure that the MetricLoggerService is decorated with @Service.  This could be something to play around with if you are looking to improve familiarity with spring and dependency injections.  The next trick would be to figure out how toggle the implementation used based on the LocalEnvironmentCondition.

nelsestu (approved):
⚡ Areas of interest:

* Generally, I think this is great and really like the concept of a shared lib for metric production.   
* Increased unit testing noted, bonus points awarded!
* LocalEnvironmentCondition is a great add, and should be something we adopt in all our metric producing code. I do also notice that `LocalEnvironmentCondition` returns `true` when the environment variable is blank. This is probably good, but makes me think we might want to log the metrics mode that things start in at a WARN level.  When debugging metrics production, we might be very happy to see the startup mode reporting in the log.
* Potential Concern:  I don't think there is anything to be done about this issue due to our required use of the datadog sdk (dogstatsd is not available to us in LHDI).   Since these calls to the sdk are synchronous, it's important that we do some analysis on the performance implications of the metrics, to ensure that the metrics logging does not significantly impact the performance of message processing.  This could be as simple as before and after response times, recognizing that by simply toggling NoopMetricLoggerService could be useful for quick before and after testing.

Security concerns: None
Relevant tests:  Yes

Review Scope: 3 out of 5 - Involves multiple changes across various components including refactoring, adding new metrics logging, and updating configurations. The changes are spread across multiple services and involve Datadog as an external system.

domain-xample/xample-workflows/src/main/java/gov/va/vro/routes/xample/BieXampleRoutes.java
brostk: Seconding using dependency injection for a MetricLoggerService bean specifically but I'd argue it's a required change - a bean already exists and by creating a new instance here we're recreating the backend connections to DataDog, but with the default MetricApi/ApiClient configuration instead of whatever is defined in DatadogClientConfig.java.

shared/lib-metrics/src/main/java/gov/va/vro/metricslogging/MetricLoggerService.java
lisac: this constructor isn't needed in the first place. oops. removed in  818887e.

domain-xample/xample-workflows/src/main/java/gov/va/vro/routes/xample/BieXampleRoutes.java
lisac: 1. Spring has been a weak spot for me; I get nervous with dependency injection!
2. We should be using dependency injection in this case, agreed.
3. Addressed in 0ca7048.

lisac:
follow ups to Erik's first round of feedback (https://github.com/department-of-veterans-affairs/abd-vro/pull/3267#pullrequestreview-2228956164) : 
 
3: 
> LocalEnvironmentCondition is a great add, and should be something we adopt in all our metric producing code.

Agreed! Kyle added this (and NonLocalEnvironmentCondition).  In [f2af12d](https://github.com/department-of-veterans-affairs/abd-vro/pull/3267/commits/f2af12df3376062d43fa07523d65729c5ad501c6) added log at `INFO` level for when a condition is matched. sample output:
```
2024-08-09T15:23:53.702Z  INFO 1 --- [main] gov.va.vro.bip.BipApiApplication         : Starting BipApiApplication using Java 17.0.10 with PID 1 (/app/fat.jar started by tron in /app)
2024-08-09T15:23:53.779Z  INFO 1 --- [main] gov.va.vro.bip.BipApiApplication         : The following 1 profile is active: "default"
2024-08-09T15:23:57.255Z  INFO 1 --- [main] g.v.v.m.LocalEnvironmentCondition        : match for LocalEnvironmentCondition
```

4.
> Potential Concern: [...] Since these [metric logging] calls to the sdk are synchronous, it's important that we do some analysis on the performance implications of the metrics,

Turns out, there IS [Asynchronous Support](https://github.com/DataDog/datadog-api-client-java?tab=readme-ov-file#asynchronous-support). Implemented in [0b7b5df](https://github.com/department-of-veterans-affairs/abd-vro/pull/3267/commits/0b7b5dfaf185873767ee1920ed1804aae122cb44).

domain-xample/xample-workflows/src/main/java/gov/va/vro/routes/xample/BieXampleRoutes.java
lisac: addressed in [c9d6be3](https://github.com/department-of-veterans-affairs/abd-vro/pull/3267/commits/c9d6be3e9d5851dd928e78f2a82cc8ce0174a911).

PR #3268, opened 2024-08-02
rollback the version number

chengjie8:
The Bitnami RabbitMQ 3.13 version has enabled feature deprecation and deprecated management metrics. I tried to insert this line with both true and false values, but the RabbitMQ container failed to start.
`deprecated_features.permit.management_metrics_collection = true`

This requires further investigation. For now, we should roll back to the previous working version.

### DIFF
--- a/.github/actions/install-docker-compose/action.yml
+++ b/.github/actions/install-docker-compose/action.yml
@@ -1,0 +1,11 @@
+name: 'Install Docker Compose'
+description: 'Installs Docker Compose on the runner'
+runs:
+  using: composite
+  steps:
+    - name: Install docker-compose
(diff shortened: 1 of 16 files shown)


lisac (approved):

nelsestu (approved):
Downgrading RabbitMQ to stay on 3.12 is a great short term solution.  Another approach would be to remove the rabbitmq_management metrics that are now being exposed via helm definition. Both approaches are mentioned in this [related slack thread](https://dsva.slack.com/archives/C04QLHM9LR0/p1722555759186659), but for now, we can approve this approach and circle back to the issue with further investigation.

PR #3271, opened 2024-08-05
include datadog environment variables for svc-bip-api deployment

lisac:
## What was the problem?
We've configured `svc-bip-api` to emit metrics using the datadog API client, however we aren't setting the needed environment variables to authenticate against the Datadog API.

- #3261

## How does this fix it?
Adds the datadog environment variables to the helm deployment.yaml for `svc-bip-api`.

Copies from EP Merge:
https://github.com/department-of-veterans-affairs/abd-vro/blob/5157613d06a26d5e87c482dbf89186eb6840464f/helm/domain-ee-ep-merge-app/templates/deployment.yaml#L32


nelsestu (approved):
@lisac we'll also want to mirror this change in argocd repo, but this looks like about all we need to get this running from this side. unfortunately, we need to keep the deployment.yaml updated in both repositories until the migration is complete.

PR #3288, opened 2024-08-09
fix syntax error in helm deployment.yaml

lisac:
## What was the problem?
The helm chart deployments for `svc-bie-kafka` and `xample-workflows` contain a syntax error, causing deployments to fail. The bug was introduced in https://github.com/department-of-veterans-affairs/abd-vro/pull/3267. 

## How does this fix it? 

Fixes the syntax error.

Tested by running `helm lint --values values.yaml --values ../_shared/values.yaml` for both cases.


nelsestu (approved):
Great catch Lisa!  This looks like the kind of one character fix that has wide scale implications. Hopefully you found it without too much frustration.

PR #3300, opened 2024-08-13
update datadog metric prefixes to be app-specific

lisac:
## What was the problem?
Datadog metrics emitted by `xample-workflows` and `svc-bie-kafka` are being attributed to `svc-bip-api`.

`svc-bip-api` has been emitting metrics under the `vro_bip` metric prefix (eg `vro_bip.REQUEST_STARTED`).
In applying the metrics code to other apps (#3267), the metric prefix was hardcoded to `vro_bip` (🙀 ).

This PR updates the metrics interface so that the desired metric prefix is specified, with `svc-bie-kafka` using metric prefix `vro_bie_kafka` and `xample-workflows` using metric prefix `vro_xample_workflows`.


chengjie8 (approved):
LGTM

PR #3312, opened 2024-08-15
Set DLQ Max length

PaulKBaumann:
## What was the problem?
We don't know how many messages will get placed on the dead letter queue. Setting a limit will prevent RabbitMQ from running out of resources in the event that there are too many messages to store. 

- https://app.zenhub.com/workspaces/vro-team-6557e67173391c000e1409f3/issues/gh/department-of-veterans-affairs/abd-vro/3238
- Discussed this solution in a Slack huddle

## How does this fix it?[^1]
Sets the max length argument on the dead letter queue

### DIFF
--- a/shared/lib-rabbitmq-connector/src/main/java/gov/va/vro/camel/MessageQueueConfiguration.java
+++ b/shared/lib-rabbitmq-connector/src/main/java/gov/va/vro/camel/MessageQueueConfiguration.java
@@ -33,6 +33,7 @@ public class MessageQueueConfiguration {
   private final String SAVE_TO_DB_PREFIX = "saveToDB-";
   private final String QUEUE_MESSAGES_DLQ = "vroDeadLetterQueue";
   private final String DLX_EXCHANGE_MESSAGES = "vro.dlx";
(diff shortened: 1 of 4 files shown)


dfitchett:
I think this PR can be closed. It is on the RabbitMQ website that [queue length limit](https://www.rabbitmq.com/docs/maxlength) is highly recommended to be set via policy rather than through queue declaration via clients:

> To set the maximum length (of either type), you can define it using a [policy](https://www.rabbitmq.com/docs/parameters#policies) (this option is highly recommended) or by clients using the queue's optional arguments

dfitchett (changes requested):
I think this PR can be closed. It is on the RabbitMQ website that [queue length limit](https://www.rabbitmq.com/docs/maxlength) is highly recommended to be set via policy rather than through queue declaration via clients:

> To set the maximum length (of either type), you can define it using a [policy](https://www.rabbitmq.com/docs/parameters#policies) (this option is highly recommended) or by clients using the queue's optional arguments

dfitchett:
Doing through policy via RabbitMQ console, not via queue declaration arguments.

PR #3342, opened 2024-08-19
Enable prometheus plugin

ponny-m:
Updated rabbitmq deployment.yaml file to enable prometheus plugin to help us see data on datadog. 

Ticket: https://github.com/department-of-veterans-affairs/abd-vro/issues/3144


nelsestu (approved):
This simple config addition looks good, and is part of the ongoing effort to enable the datadog provided rabbitmq monitoring. This change looks good, but it also reminded me to ask: do you know how the pod is authorizing with datadog?  Is it possible we need to be passing the SDK credentials in, like we do with bgs, bip and bie?

ponny-m:
@nelsestu In reponse to your comment above, yes I created a config map (**rabbitmq-config**) in the va-abd-rrd-dev namespace and mounted it as a volume. But it wouldn't hurt to passed the SDK creds in just in case.

lisac:
shouldn't hold up this PR merge, but in case we want to be able to account for the prometheus plugin when deploying from argocd, here's the corresponding [deployment.yaml](https://github.com/department-of-veterans-affairs/va-abd-rrd-argocd-applications-vault/blob/a4c4fd478b6819947e322a931631e39701546f05/deploy/vro-rabbitmq/templates/deployment.yaml).

ponny-m:
@lisac Good call. I will make this change in that repo as well. Thanks

Issue #3327, opened 2024-08-15
RabbitMQ: Apply max queue length policy to all queues


dfitchett:
## Summary 
Currently none of the queues on any of the exchanges have a limit to the max number of messages they can hold at any given time. More information on how to apply policies can be found on on the [RabbitMQ docs](https://www.rabbitmq.com/docs/maxlength).

# Risk
If any service such as `svc-bie-kafka` goes down, then the messages will build up in the queue or build up in the dead letter queue and if the service is not brought back up to consume those messages, worse case, RabbitMQ will crash due to running out of space.

# Acceptance criteria
Every queue, including dead letter queues are configured to have a maximum number of messages in a queue. We can utilize the default overflow behavior. 

### Developer Notes
The max message limit must be set by policy in each deployment of RabbitMQ. If RabbitMQ is redeployed, the policy must be set again.

lisac:
@dfitchett  question about the timing on this: is this dependent on a dead letter queue being configured/deployed in an environment? (or is this desired regardless of whether a DLQ is present?)

dfitchett:
> @dfitchett question about the timing on this: is this dependent on a dead letter queue being configured/deployed in an environment? (or is this desired regardless of whether a DLQ is present?)

This is required regardless of the DLQ configurations 👍

dfitchett:
See [PR](https://github.com/department-of-veterans-affairs/abd-vro/pull/3346) to set this property via policy on upload!

PR #3346, opened 2024-08-20
Add max-length to global queue policy. Renamed Policy.

dfitchett:
## What was the problem?

There is no max queue size for any of the queues in RabbitMQ, thus risking an overflow event if any of the services are down for an extended period of time.

- #3327

## How does this fix it?[^1]

Sets the max-length via policy on rabbitmq launch. Any resource (queue/exchange) may only have a single policy definition, so the max-length and dead letter exchange properties are set by the same global queue policy. See info on [combining policies](https://www.rabbitmq.com/docs/parameters#combining-policy-definitions). 

## How to test this PR
- In docker: 
  - Run svc-bip-api and rabbitmq services
  - Verify the queues created have the policy setting the max-length property and the dead letter queue via the rest api or rabbitmq admin console

### DIFF
--- a/rabbitmq/src/main/resources/definitions.json
+++ b/rabbitmq/src/main/resources/definitions.json
@@ -23,11 +23,12 @@
   "policies": [
     {
       "vhost": "/",
-      "name": "DLX",
+      "name": "VRO-GLOBAL-QUEUES",
       "pattern": ".*",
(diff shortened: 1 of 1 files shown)


lisac (approved):
LGTM!

PaulKBaumann (approved):

PR #3349, opened 2024-08-21
Automate enablement of prometheus plugin for Rabbitmq pod

ponny-m:
Updated rabbitmq deployment.yaml file to enable prometheus plugin to help us see data on datadog. The issue was when the pod was created/restarted this plugin wasn't enabled and had to be done manually. This solution has been tested in dev pod and the plugin is enabled during restart/creation of the rabbitmq pod. 

Ticket: https://github.com/department-of-veterans-affairs/abd-vro/issues/3144

###How to Test?
1. Delete the pod (**ONLY in Dev**) OR restart the rabbitmq deployment
2. kubectl exec -it Rabbitmq-Pod-Name -- rabbitmq-plugins list


lisac (approved):
looks consistent with documentation on https://www.rabbitmq.com/docs/prometheus#enable-rabbitmq_prometheus

PR #3375, opened 2024-08-22
Remove vro prefix for rabbitmq image

ponny-m:
Removed vro prefixing for Rabbitmq image in hopes for LHDI datadog agent autodiscovery feature to auto discover the Rabbitmq container.


helm/platform/charts/rabbitmq/templates/deployment.yaml
```diff
@@ -17,7 +17,7 @@ spec:
       {{- include "vro.imagePullSecrets" . | nindent 6 }}
       containers:
         - name: rabbitmq{{ include "vro.containerSuffix" . }}
-          image: ghcr.io/department-of-veterans-affairs/abd-vro-internal/vro-rabbitmq:{{ .Values.imageTag }}
+          image: {{ include "vro.imageRegistryPath" . }}vro-rabbitmq:{{ .Values.imageTag }}
```
msnwatson: this still uses vro prefix

scripts/image-names.sh
```diff
@@ -17,7 +17,11 @@ bashVarPrefix() {
 
 # These names must match the images specified in Helm configs
 prodImageName() {
-  echo "vro-$1"
+  if [ "$1" == "rabbitmq" ]; then
```
msnwatson, line 23: can you comment as to why we have this special case

ponny-m: Absolutely. 

This condition ensures that our RabbitMQ image is named "rabbitmq" instead of "vro-rabbitmq", while all other images will retain the "vro-" prefix. This adjustment will allow LHDI Datadog agent's autodiscovery feature to correctly match our rabbitmq container image name thereby resulting in the capturing of metrics.

helm/platform/charts/rabbitmq/templates/deployment.yaml
ponny-m: Fixed

scripts/image-names.sh
msnwatson: oh i meant in the code 😅  just bc when someone comes back to this they'll think, why? @Ponnia-M

ponny-m: Ahh true.

ponny-m: Done. 

For next steps, which github action runs this script? and where can I confirm this change has occurred? @msnwatson

msnwatson: so, you'll have to run this script manually to generate the right output but then you'll also have to change the deployment pipelines for prod to use this new name and then also change the deployments in the argo repo to use this new image name pattern

msnwatson: be very cautious as this is a breaking change to our deployment of rabbitmq and there is a lot of stuff going on with that image right now. communicate this to the team probably in slack

msnwatson (approved):

scripts/image-names.sh
ponny-m: Thanks for the head up. I will do this tomorrow when things slow down.

PR #3378, opened 2024-08-23
Bgs healthcheck

ponny-m:
Here is a PR to remove vro prefix from rabbitmq and update image_var.src file


scripts/image_vars.src
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
-# This file is autogenerated -- update and re-run ./scripts/image-names.sh instead.
```
lisac: should this comment change be reverted to restore the `./scripts/` reference?

ponny-m: Fixed

lisac (approved):

scripts/image_vars.src
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 # This file is autogenerated -- update and re-run ./scripts/image-names.sh instead.
-# Usage: source scripts/image_vars.src
+# Usage: source image_vars.src
```
lisac: one more instance of restoring the comment?

PR #3385, opened 2024-08-27
Update datadog secret in BGS

ponny-m:
Updated BGS helm chart to retrieve secret details for datadog user in all environments.

### DIFF
--- a/helm/svc-bgs-api/values.yaml
+++ b/helm/svc-bgs-api/values.yaml
@@ -54,3 +54,8 @@ bip:
   # TODO: Consider moving these URL settings to application-*.yml files
   bipClaimUrl:    claims-uat.stage.bip.va.gov/api/v1
   bipEvidenceUrl: vefs-claimevidence-uat.stage.bip.va.gov/api/v1/rest
+
+secrets:
+  rabbitmq:
+    name: vro-dd-rabbitmq-pass
+    key: rabbitmq_pass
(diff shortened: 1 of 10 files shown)


chengjie8 (approved):
LGTM

helm/svc-bgs-api/values.yaml
```diff
@@ -54,3 +54,9 @@ bip:
   # TODO: Consider moving these URL settings to application-*.yml files
   bipClaimUrl:    claims-uat.stage.bip.va.gov/api/v1
   bipEvidenceUrl: vefs-claimevidence-uat.stage.bip.va.gov/api/v1/rest
+
+secrets:
+  rabbitmq:
+    name: vro-dd-rabbitmq-pass
+    key: rabbitmq_pass
```
lisac, line 61: it looks like `secrets.rabbitmq.name` and `secrets.rabbitmq.key` are consistent, regardless of environment, is that correct? if so, then I'd recommend removing the overrides in the env-specific files (eg `values-for-sandbox.yaml`).

scripts/image_vars.src
```diff
@@ -49,7 +49,7 @@ imageTagExists(){
 export postgres_GRADLE_IMG="va/abd_vro-postgres"
 export postgres_IMG="vro-postgres"
 
-export rabbitmq_GRADLE_IMG="va/abd_vro-rabbitmq"
+export rabbitmq_GRADLE_IMG="va/abd_rabbitmq"
```
lisac: to double-check: is this change in scope of this PR?

ponny-m: Nope. Took it out

helm/svc-bgs-api/values.yaml
ponny-m: Yes but when we had it in the deployment.yaml it wasn't being pulled/mounted in other environments.

Issue #2817, opened 2024-04-03
Custom Monitoring for Rabbit MQ and K8s termination grace period


msnwatson:
**User Story**

As a VRO engineer, I want to take action on the findings from #2816 and close any availability or monitoring gaps, so that the platform's ability to offer service uptime for partners is improved.

**Acceptance Criteria**

1. This ticket only requires request latency metrics instrumentation for RabbitMQ
2. Request latency metrics for all services in VRO are analyzed and used to set K8s termination grace periods for all of our services. 
3. Solutions for any problems identified in #2816 are implemented and documented as necessary
4. How would we test this? Can I see this metrics in datadog after I deploy to lower environments? You should be able to.

msnwatson:
We'll need to go a slightly separate direction than the ticket had originally planned. It does not look like RabbitMQ supports standard metrics for the time messages sit in a queue.

Rather, I think we need to do two things which is set a TTL on messages in our queues and get metrics and alert on spikes in messages expiring within our queues. This will allow us to still set a termination grace period in a principled way while maintaining visibility into any negative effects of this policy change.

lisac:
status: still working on resolving the blocker. details added to #3261. in summary: BIP metrics should now be visible; and I see Wednesday Aug 7 as a stretch goal for getting the metrics on the other apps working.

lisac:
status: last round of changes on the blocker expected to be deployed 8/12.

lisac:
status: I believe the blocker has been addressed.
Request duration is being logged under these metrics: `vro_xample_workflows.request_duration`, `vro_bie_kafka.request_duration`, `vro_bip.request_duration`

msnwatson:
Per this thread: https://dsva.slack.com/archives/C04QLHM9LR0/p1723848302659089 I took a slightly different approach than mentioned in the ticket based on the metrics that I was seeing. Still blocked on getting RabbitMQ metrics.

PR #3413, opened 2024-08-30
Add 25 minute TTL for all RabbitMQ messages

msnwatson:
## What was the problem?

I could not find a default TTL for messages in RabbitMQ which could result in messages persisting longer than default termination grace period for k8s. 

- #2817 

## How does this fix it?[^1]

see title

## How to test this PR
- Confirmed by deployment of changes to `dev` env in LHDI


chengjie8 (approved):
LGTM

rabbitmq/src/main/resources/definitions.json
```diff
@@ -33,6 +33,14 @@
     },
     {
       "vhost": "/",
+      "name": "VRO-GLOBAL-MESSAGE-TTL",
+      "pattern": ".*",
+      "apply-to": "queues",
+      "definition": {
+        "message-ttl": 1500000
```
dfitchett: Every resource (queues/exchanges) can only have one policy applied at a time, and if the priorities match, then the policy applied to any queue matching the pattern (as in regex) is basically random and not guaranteed to be consistent. `VRO-GLOBAL-MESSAGE-TTL` and `vro_global_queues` both match the pattern ".*". 

This means that you really need to add the message-ttl to every policy if that's the desired behavior, and you can remove this `VRO-GLOBAL-MESSAGE-TTL` policy.

dfitchett (changes requested):

rabbitmq/src/main/resources/definitions.json
msnwatson: @dfitchett  this is not great as there is no way to enforce that for any new policies include this condition which we will come to depend on in our monitoring for uptime.

Is there a reason we don't set the more specific properties at the code level. I was thinking that we could keep this file more for global properties of rabbitmq

rabbitmq/src/main/resources/definitions.json
```diff
@@ -37,7 +38,8 @@
       "pattern": ".*healthcheck.*",
```
lisac, line 38: (this is not a change introduced by this PR and shouldn't hold up the merge, but might be relevant for testing the PR)

question: how many queue(s) do we anticipate this matching? I think the answer is one, but I wanted to double-check.

Browsing rabbitmq on sandbox, i see we have a queue named `healthcheck`; and in the default branch i see  in [constants.rb](vro/blob/23c883a536a36545acead4267f38645723e51d66/svc-bgs-api/src/config/constants.rb) a queue named `svc_bgs_api.healthcheck` (and i think it's just not deployed). regarding the pattern match, is the idea that it'll match the latter?  if so, should the regex be `*healthcheck`?

lisac (approved):

rabbitmq/src/main/resources/definitions.json
dfitchett: > @dfitchett this is not great as there is no way to enforce that for any new policies include this condition which we will come to depend on in our monitoring for uptime.

We will have to enforce this through pull requests, and ensure that each of the desired properties are added to the appropriate policies.

> Is there a reason we don't set the more specific properties at the code level. I was thinking that we could keep this file more for global properties of rabbitmq

The main reason for setting some of these properties via policy, is **deployability**. For example, if `message-ttl` is set on the queue at declaration time for a specific queue, then every service that declares that queue for use must also set that property on declaration. If the two or more services using that queue do not declare it exactly, then only the first one to declare it will be able to connect to it, and the others will fail with a 406 (see [here](https://www.rabbitmq.com/docs/queues#property-equivalence)).

Lets say that we change all services to match the new configuration, then all those services will have to be redeployed, which we just tried to go through (painfully) with the dead letter configurations. Additionally, queues/exchanges will need to be manually deleted if any queues/exchanges were declared with `autodelete=false` or any cases where `autodelete=true` but the queue never had a consumer, or the exchange never had any bindings (meaning they will not autodelete themselves). 

The docs say that its recommended that all optional parameters are set via policy:

> Optional queue arguments can be set in a couple of ways:
> * To groups of queues using [policies](https://www.rabbitmq.com/docs/parameters#policies) (recommended)
> * On a per-queue basis when a queue is declared by a client
> * For the x-queue-type argument, [using a default queue type](https://www.rabbitmq.com/docs/vhosts#default-queue-type)
>
> The former option is more flexible, non-intrusive, does not require application modifications and redeployments. Therefore it is highly recommended for most users.

Additionally, there are a handful of properties that are better set at the message level on publish as opposed to at the policy level, `message-ttl` being on of them. By setting this at the client level, the client (publisher) is charge of how long it should wait for that message to be consumed. 

For example, the EP merge app will add a message-ttl via the `expires = 0` property to every request it makes to the BIP and BGS services. This means that if the BIP or BGS service is not available to consume the message, the message is dropped or dead lettered immediately, and EP merge assumes it failed, and will continue in processing the merge appropriately. If another application was willing to wait longer, they could set it to 60 seconds or whatever is appropriate for that application's logic.

rabbitmq/src/main/resources/definitions.json
dfitchett: So that `pattern` is a regex pattern, and `".*healthcheck.*"` will actually match both of those. I'll be documenting some of this in the wiki when I get a chance today, but essentially, this will match any services "healthcheck" queues, and yes, currently only bgs has one.

dfitchett (approved):
Changes look good to me. One non-blocking question, why 25 minutes?

msnwatson:
> Changes look good to me. One non-blocking question, why 25 minutes?

The K8S termination grace period is 30 minutes from when the pod receives the signal. I'm trying to set an upper bound on message processing times that still allows enough time for a quick investigator to inspect the DLQ in the event of pod shutdown after 5 minutes. I also want this to trigger a monitor in Datadog so that we can be alerted about the potential for any bad behavior in between pods.

msnwatson:
I'm not super confident that we'll be able to enforce the policy that message-ttl is added to all policies, as I think there is no way to add comments to a JSON file

PR #3434, opened 2024-09-09
update setuptools to use version 70 or higher

chengjie8:


lisac (approved):
consistent with the version in `lib-hoppy`
https://github.com/department-of-veterans-affairs/abd-vro/blob/858c30d54101f29fd085f44bc2e24d65a18f9d2c/shared/lib-hoppy/pyproject.toml#L2

PR #3518, opened 2024-09-26
Brostk/db init flyway migration fix

brostk:
Adding an empty 1.7 migration file fails the checksum against the previous 1.7 file.  Adding a `repair` call and setting `ignoreMissingMigrations` to `true` to clean up the 1.7 migration and ignore it going forward.


chengjie8 (approved):
LGTM

dfitchett (approved):

Issue #3536, opened 2024-09-30
svc-bgs-api: Failed Datadog requests delay responses being sent to requestor


dfitchett:
# Summary
Failed EP Merge Job alerts indicated an issue happening with BGS downstream services. The BGS service is coded in a way where any delay on publishing metrics will cause delay in returning the response to the requesting client. 

This was noticed by a failed EP Merge job that sent multiple requests to BGS. Here are the BGS logs below: 
![Image](https://github.com/user-attachments/assets/7068ab54-c13e-4303-b538-a5dd941ad624)

## Acceptance Criteria
`svc-bgs-api` is updated so that the calls to publish metrics does not actually affect the success of processing the request and returning the response to the client.

dfitchett:
Previous PR did not address this issue completely. New PR is up.

PR #3543, opened 2024-10-01
svc-bgs-api: Make datadog calls happen after response is sent to client.

dfitchett:
## What was the problem?

Datadog API requests are sent synchronously during the processing of a message from the queue to add claim notes. This can cause delays if there are issues with the connection to datadog. This also artificially inflates the duration metric for processing add note requests. 

- This is a follow-on to #3541 
- #3536 

## How does this fix it?[^1]

Makes the calls async by using ruby gem `async` with a do block adding the calls to a task list to be run concurrently. This will ensure that the datadog calls do not affect the responses sent to the client. 

Previous PR #3541 did not solve this, as the publishing was still being affected by the datadog calls. This PR moves those async calls to the ensure block where the response is published to the client. 

## How to test this PR
* Run rspec: `bundle exec rspec` from `svc-bgs-api/src` directory
* Run integration tests:
  1. Run platform with bgs
  2. Run `bundle exec ruby integrationTest/run_integration_test.rb` from `svc-bgs-api/src` directory

### DIFF
--- a/svc-bgs-api/src/lib/metric_logger.rb
+++ b/svc-bgs-api/src/lib/metric_logger.rb
@@ -36,8 +36,10 @@ class MetricLogger
     begin
       api_instance = DatadogAPIClient::V1::AuthenticationAPI.new
       api_instance.validate
+    rescue DatadogAPIClient::APIError => e
+      $logger.warn("event=failedDatadogAuthentication error=#{e.class} status=#{e.code} body=#{e.response_body}")
     rescue Exception => e
-      $logger.warn("event=failedDatadogAuthentication error=#{e.message}")
+      $logger.warn("event=failedDatadogAuthentication error=#{e.class}  error=#{e.message}")
     end
   end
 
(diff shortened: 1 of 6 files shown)


PaulKBaumann (approved):

svc-bgs-api/src/lib/metric_logger.rb
```diff
@@ -36,8 +36,10 @@ def initialize
     begin
       api_instance = DatadogAPIClient::V1::AuthenticationAPI.new
       api_instance.validate
+    rescue DatadogAPIClient::APIError => e
+      $logger.warn("event=failedDatadogAuthentication error=#{e.class} status=#{e.code} body=#{e.response_body}")
     rescue Exception => e
-      $logger.warn("event=failedDatadogAuthentication error=#{e.message}")
+      $logger.warn("event=failedDatadogAuthentication error=#{e.class}  error=#{e.message}")
```
lisac, line 42: question: is the intent to log a dictionary object of key-values? in which case, will only one of the "error" keys get stored?

lisac (approved):
one question about a logging message, but i think shouldn't block the merge

svc-bgs-api/src/lib/metric_logger.rb
dfitchett: It depends on the exception that is caught. Most issues with the datadog metrics will be caught in the rescue above. The second one is a catch-all, and we can just log the message. I didn't want to do that with the datadog client because it formats it funky with several newlines. See [here](https://github.com/DataDog/datadog-api-client-ruby/blob/60166bd7d9694515d4daf06c506b7b57f43cdfde/lib/datadog_api_client/api_client.rb#L569) to see how it formats the `e.message`.

PR #3548, opened 2024-10-02
svc-bgs-api: Fixed liveness and readiness checks.

dfitchett:
Previous commit broke liveness/readiness check. 


brostk (approved):

PaulKBaumann (approved):

ponny-m (changes requested):
This looks solid, but what about the **start probe**. Issue I encountered when setting this up was bgs app took a while to start and the liveness/readiness failed so to check this I created the start probe to delay and check that the BGS app started before liveness/readiness check.

dfitchett:
> This looks solid, but what about the **start probe**. Issue I encountered when setting this up was bgs app took a while to start and the liveness/readiness failed so to check this I created the start probe to delay and check that the BGS app started before liveness/readiness check.

@Ponnia-M Looking at the startup_script.rb there is not anything that deals with creating a connection to RabbitMQ, so it **should** remain unaffected.

ponny-m:
@dfitchett True it does. Since the focus is Rabbitmq then this is fine then. Thanks

ponny-m (approved):

Issue #3528, opened 2024-09-27
Add monitor tags to svc-bip-api


brostk:
**User Story**

As a VRO engineer, I want mandatory tags added to metric events, so that we meet VA monitoring requirements.

**Acceptance Criteria**

1. All required tags in the below tagging standards document have been added to BIP events.
2. Update datadog [dashboard](https://lighthousedi.ddog-gov.com/dashboard/imt-ubp-z29/vro-data-services-bip?fromUser=true&refresh_mode=sliding&from_ts=1725227499039&to_ts=1727819499039&live=true) to be able to filter by new tags.
3. Update monitors to filter by new tags.

**Notes about work**
- Monitor Tagging Standards: https://depo-platform-documentation.scrollhelp.site/developer-docs/monitor-tagging-standards
- An easy solution could be to add new properties to MetricLoggerService by modifying the list of incoming tags

PR #3550, opened 2024-10-02
3528 add monitoring tags

brostk:
## What was the problem?
Monitoring tags are required by the VA but not present on metrics.

- #3528 

## How does this fix it?[^1]
`MetricLoggerService` now includes all required tags with values populated by properties defined in `application.yaml`.

### DIFF
--- a/shared/lib-metrics/src/main/java/gov/va/vro/metricslogging/MetricLoggerService.java
+++ b/shared/lib-metrics/src/main/java/gov/va/vro/metricslogging/MetricLoggerService.java
@@ -5,6 +5,7 @@ import com.datadog.api.client.v1.model.*;
 import jakarta.validation.constraints.NotNull;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.Conditional;
 import org.springframework.stereotype.Service;
 
@@ -18,6 +19,21 @@ import java.util.*;
 public class MetricLoggerService implements IMetricLoggerService {
 
   private final MetricsApi metricsApi;
+
+  @Value("${vro.env}")
+  public final String env;
+
+  @Value("${vro.it-portfolio:benefits-delivery}")
+  public final String itPortfolio;
+
+  @Value("${vro.team:va-abd-rrd}")
+  public final String team;
+
+  @Value("${vro.app.name}")
+  public final String service;
+
+  @Value("${vro.app.dependency:")
+  public final String dependency;
 
   public static double getTimestamp() {
     return Long.valueOf(OffsetDateTime.now().toInstant().getEpochSecond()).doubleValue();
@@ -43,6 +59,15 @@ public class MetricLoggerService implements IMetricLoggerService {
     if (customTags != null) {
       tags.addAll(Arrays.asList(customTags));
     }
+
+    tags.add("env:" + env);
+    tags.add("itportfolio:" + itPortfolio);
+    tags.add("team:" + team);
+    tags.add("service:" + service);
+    if (dependency != null && !dependency.isEmpty()) {
+      tags.add("dependency:" + dependency);
+    }
+
     return tags;
   }
 
--- a/shared/lib-metrics/src/main/java/gov/va/vro/metricslogging/MetricLoggerService.java
+++ b/shared/lib-metrics/src/main/java/gov/va/vro/metricslogging/MetricLoggerService.java
@@ -5,6 +5,7 @@ import com.datadog.api.client.v1.model.*;
 import jakarta.validation.constraints.NotNull;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.StringUtils;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.Conditional;
 import org.springframework.stereotype.Service;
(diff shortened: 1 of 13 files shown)


shared/lib-metrics/src/main/java/gov/va/vro/metricslogging/MetricLoggerService.java
```diff
@@ -43,6 +59,15 @@ public ArrayList<String> getTagsForSubmission(String[] customTags) {
     if (customTags != null) {
       tags.addAll(Arrays.asList(customTags));
     }
+
+    tags.add("env:" + env);
+    tags.add("itportfolio:" + itPortfolio);
+    tags.add("team:" + team);
+    tags.add("service:" + service);
+    if (dependency != null && !dependency.isEmpty()) {
```
dfitchett, line 67: since we use apache commons this line:
`if (dependecy != null && !dependency.isEmpty()) {`
can be replaced with
`if (StringUtils.isNotBlank()) {`

shared/lib-metrics/src/main/java/gov/va/vro/metricslogging/MetricLoggerService.java
```diff
@@ -19,6 +20,21 @@ public class MetricLoggerService implements IMetricLoggerService {
...
+  public final String itPortfolio;
+
+  @Value("${vro.team:va-abd-rrd}")
+  public final String team;
+
+  @Value("${vro.app.name}")
+  public final String service;
+
+  @Value("${vro.app.dependency:")
+  public final String dependency;
```
dfitchett: This could potentially be a list of tags like `dependency:bip_claims_api, dependency:bip_evidence_api`. So really this should maybe be a list of dependencies.

shared/lib-metrics/src/main/java/gov/va/vro/metricslogging/MetricLoggerService.java
dfitchett: since we use apache commons this line:
`if (dependecy != null && !dependency.isEmpty()) {`
can be replaced with
`if (StringUtils.isNotBlank()) {`

shared/lib-metrics/src/main/java/gov/va/vro/metricslogging/MetricLoggerService.java
```diff
@@ -19,6 +20,21 @@ public class MetricLoggerService implements IMetricLoggerService {
...
+  @Value("${vro.env}")
+  public final String env;
+
+  @Value("${vro.it-portfolio:benefits-delivery}")
+  public final String itPortfolio;
+
+  @Value("${vro.team:va-abd-rrd}")
+  public final String team;
+
+  @Value("${vro.app.name}")
```
dfitchett: Non-blocking. Can we make this consistent with the name of the variable so that in my properties, i know this will correspond directly to the `service` tag?

svc-bip-api/src/main/resources/application.yaml
```diff
@@ -59,3 +59,9 @@ management:
 truststore: ${BIP_TRUSTSTORE}
 keystore: ${BIP_KEYSTORE}
 truststore_password: ${BIP_PASSWORD}
+
+vro:
+  env: ${ENV}
+  app:
+    name: vro-svc-bip-api
+    dependency: bip
```
dfitchett: Let's change the dependency here to `bip-claims-api` since that team owns several APIs

dfitchett (changes requested):
The changes here propagate into the [ContentionEventsListener](https://github.com/department-of-veterans-affairs/abd-vro/blob/b09f5585fdb964342f65890474e128b8a344125e/svc-bie-kafka/src/main/java/gov/va/vro/services/bie/service/kafka/ContentionEventsListener.java) class in the `svc-bie-kafka` and some of the tags will be duplicated. Can you extend your review to fix that as well?

dfitchett:
Ah one other thing that is just a thought... what about making the metric prefix also part of this MetricLogger class and imported as a property?

shared/lib-metrics/src/main/java/gov/va/vro/metricslogging/MetricLoggerService.java
dfitchett, line 1: Also, the log.info statements when  the metrics were successfully submitted should probably be debug not info

shared/lib-metrics/src/main/java/gov/va/vro/metricslogging/MetricLoggerService.java
brostk: I'd assumed for some reason we didn't use apache commons, will update!

shared/lib-metrics/src/main/java/gov/va/vro/metricslogging/MetricLoggerService.java
brostk: Interesting, I didn't interpret the documentation as allowing a collection.  I'll turn this into a collection.

shared/lib-metrics/src/main/java/gov/va/vro/metricslogging/MetricLoggerService.java
brostk: Good call out, I'll update this.

brostk:
> The changes here propagate into the [ContentionEventsListener](https://github.com/department-of-veterans-affairs/abd-vro/blob/b09f5585fdb964342f65890474e128b8a344125e/svc-bie-kafka/src/main/java/gov/va/vro/services/bie/service/kafka/ContentionEventsListener.java) class in the `svc-bie-kafka` and some of the tags will be duplicated. Can you extend your review to fix that as well?

This was your recent addition, right?  Any objections to removing the tags here in favor of the common code addition in MetricLoggerService?

> Ah one other thing that is just a thought... what about making the metric prefix also part of this MetricLogger class and imported as a property?

Looks plenty do-able, but it'll touch a few interfaces.  Be prepared for a larger PR!

> Also, the log.info statements when the metrics were successfully submitted should probably be debug not info

I can update these while I'm in the code.

dfitchett (approved):
LGTM! 🚀

PR #3969, opened 2025-01-15
removing terms to correct issue in classifications

tyler-spangler6:
## What was the problem?
This fixes an error identified in the rollout of the new expanded classification method.  With this method the classifier is classifying the terms osteoarthritis, tendinitis, and difficulty swallowing incorrectly.  Because there are multiple classifications for these based on additional information, the classification should be None for these terms. This PR removess these from the lookup table until a more robust solution can be implemented.  

[DataDog logs displaying incorrect classifications](https://lighthousedi.ddog-gov.com/logs?query=service%3Avro-cc-app%20%28%40is_lookup_table_match%3Atrue%20%40is_in_dropdown%3Afalse%20%40claim_type%3Anew%29%20cluster_name%3Aldx-prod-1%20%40endpoint%3A%2Fexpanded-contention-classification%20%28%40contention_text%3A%22unmapped%20contention%20text%20%5B%27osteoarthritis%27%5D%22%20OR%20%40contention_text%3A%22unmapped%20contention%20text%20%5B%27tendinitis%27%5D%22%20OR%20%40contention_text%3A%22unmapped%20contention%20text%20%5B%27difficulty%20swallowing%27%5D%22%29&agg_m=count&agg_m_source=base&agg_t=count&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&saved-view-id=928&storage=hot&stream_sort=desc&viz=stream&from_ts=1734364981902&to_ts=1736956981902&live=true)

Associated tickets or Slack threads:
- https://dsva.slack.com/archives/C04AZ8T7XN1/p1736959628494219

## How does this fix it?[^1]
Removes the terms from the list

## How to test this PR
- Step 1
- Step 2[^secrel]

[^1]: [Pull-Requests guidelines](https://github.com/department-of-veterans-affairs/abd-vro/wiki/Pull-Requests). If PR is significant, update [Current Software State](https://github.com/department-of-veterans-affairs/abd-vro/wiki/Current-Software-State) wiki page.
[^secrel]: To check if a PR will succeed in the SecRel workflow, [test PRs in the SecRel pipeline](https://github.com/department-of-veterans-affairs/abd-vro-internal/wiki/Secure-Release-process#to-test-prs-in-the-secrel-pipeline).

PR #3972, opened 2025-01-15
Fully corrects Parentheses Issue in Lookup table build

tyler-spangler6:
## What was the problem?
This is a larger fix for the bugfix in https://github.com/department-of-veterans-affairs/abd-vro/pull/3969.  Before this PR, parenthetical terms were included as individual terms in the lookup table and for some terms this provided incorrect classifications.  https://github.com/department-of-veterans-affairs/abd-vro/pull/3969 only removed these terms from the list, but a larger fix would be to include a way that we can correctly classify these terms.  

- https://dsva.slack.com/archives/C04AZ8T7XN1/p1736959628494219

## How does this fix it?[^1]

This fix will include the body part with the parenthetical term to ensure proper classification. Example: 
term: degenerative arthritis (osteoarthritis) in wrist, left" will create following lookup options:
1. degenerative arthritis osteoarthritis wrist
2. osteoarthritis wrist
3. degenerative arthritis wrist
Separating out the options in this manner will allow us to correctly classify this term.  

## How to test this PR
- Unit tests are written for the functionality to remove parenthetical terms. 


domain-cc/cc-app/tests/test_expanded_lookup.py
```diff
@@ -271,3 +271,34 @@ def test_removed_terms_not_in_lut():
...
+def test_parenthetical_removal_multiple():
+    """
+    There are currently no terms in the CSV that have multiple parenthetical values
+    """
+    test_str = "acl tear (anterior cruciate ligament tear) in (left knee), right knee"
+    expected = [
+        "anterior cruciate ligament tear ) in (left knee), right knee",
+        "left knee ), right knee",
+        "acl tear  in , right knee",
+    ]
```
CorySohrakoffUSDS: Understood that we don't have any terms currently that match this test case, but is this the behavior we want in this case?

tyler-spangler6: No I don't think that is the right behavior; thanks for asking this. I think it would be better to have each parenthetical term use the last part of the string. I updated the functionality and tried to think of a better test case.  I am not sure of a time that we would include two synonyms for a condition, but swapped out the test case for "knee replacement (knee arthroplasty) (knee joint replacement), bilateral" to try and be a little more realistic with something we could add.

CorySohrakoffUSDS (approved):
LGTM but recommend at least one other reviewer.

domain-cc/cc-app/src/python_src/util/expanded_lookup_table.py
```diff
@@ -85,6 +85,37 @@ def _removal_pipeline(self, text):
 
         return text.lower().strip()
 
+    def _remove_parenthetical_terms(self, text: str) -> List[str]:
```
williamphelps13, line 88: What do you think of something like the following? Was exploring ways to DRY up the code:
```
   def _remove_parenthetical_terms(self, text: str) -> List[str]:
        """
        Removes terms within contention text values that are included in parenthetical. This will append the
        rest of the string to the parenthetical, and create a list of terms from this.

        Ex: "degenerative arthritis (osteoarthritis) in wrist, right" will return
            ["osteoarthritis in wrist, right", "degenerative arthritis in wrist, right"]

        This will then move through the removal pipeline to build the lookup table.
        """
        parenthetical_terms = re.findall(r"\((.*?)\)", text)
        base_text = re.sub(r"\(.*?\)", "", text)
        variations = [
            f"{term} {base_text.split(term)[1]}" for term in parenthetical_terms
        ]
        variations.append(base_text)
        return [self._removal_pipeline(v) for v in variations]

    def _add_to_lut(self, key: str, row: dict, classification_code_mappings: dict):
        """
        Adds a processed key and corresponding classification data to the lookup table.

        Args:
            key (str): The string key to process and add.
            row (dict): A row from the CSV containing classification data.
            classification_code_mappings (dict): The lookup table to update.
        """
        processed_key = frozenset(self._removal_pipeline(key).split())
        classification_code_mappings[processed_key] = {
            "classification_code": int(row[self.classification_code]),
            "classification_name": row[self.classification_name],
        }

    def _build_lut(self):
        """
        Builds the lookup table using the CSV file

        This also pulls out terms in parentheses and adds the separated strings to the list and also keeping the OG term
        """
        classification_code_mappings = {}
        with open(FILE_READ_HELPER["filepath"]) as fh:
            csv_reader = csv.DictReader(fh)
            for row in csv_reader:
                if "(" in row[self.key_text]:
                    ls_terms = self._remove_parenthetical_terms(row[self.key_text])
                    for term in ls_terms:
                        self._add_to_lut(term, row, classification_code_mappings)

                # adds the original string
                self._add_to_lut(row[self.key_text], row, classification_code_mappings)

            # adds the joint lookup to the table
            classification_code_mappings.update(self._musculoskeletal_lookup())

        return classification_code_mappings
```

williamphelps13 (approved):
LGTM!

